PR #13, opened 2016-10-17
Add "volume" parameter to command line and handle reconnection to the agent

sboeuf:
- Add volume parameter to command line
- Fix comment formatting issues
- Check connection status every time start() is called

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -105,7 +105,7 @@ func newAgentConfig(config PodConfig) interface{} {
 	}
 }
 
-// Agent is the virtcontainers agent interface.
+// agent is the virtcontainers agent interface.
 // Agents are running in the guest VM and handling
 // communications between the host and guest.
 type agent interface {
@@ -122,9 +122,9 @@ type agent interface {
 	// exec will tell the agent to run a command in an already running container.
 	exec(podID string, contID string, cmd Cmd) error
 
-	// StartPod will tell the agent to start all containers related to the Pod.
+	// startPod will tell the agent to start all containers related to the Pod.
 	startPod(config PodConfig) error
 
-	// StopPod will tell the agent to stop all containers related to the Pod.
+	// stopPod will tell the agent to stop all containers related to the Pod.
 	stopPod(config PodConfig) error
 }
--- a/hyperstart.go
+++ b/hyperstart.go
@@ -230,7 +230,10 @@ func sendCmd(c net.Conn, cmdID string, payload interface{}) error {
 
 	glog.Infof("payload: %s\n", payloadStr)
 	intLen := len(payloadStr) + ctlHdrSize
-	payloadLen := intTo4BytesString(intLen)
+	payloadLen, err := uint64ToNBytesString(uint64(intLen), 4)
+	if err != nil {
+		return err
+	}
 	glog.Infof("payload len: %x\n", payloadLen)
 
 	frame := frame{
@@ -239,7 +242,7 @@ func sendCmd(c net.Conn, cmdID string, payload interface{}) error {
 		payload:    payloadStr,
 	}
 
-	err := send(c, frame)
+	err = send(c, frame)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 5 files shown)


virtc/main.go
```diff
@@ -100,6 +100,18 @@ var podConfigFlags = []cli.Flag{
...
+		Name:  "hyper-mount-tag-list",
+		Value: new(cli.StringSlice),
+		Usage: "the hyperstart mount tag list",
+	},
+
+	cli.StringSliceFlag{
+		Name:  "hyper-host-path-list",
+		Value: new(cli.StringSlice),
+		Usage: "the hyperstart host path list",
+	},
```
sboeuf: I have updated the commit. It now expects an entry like this: -volume="mounttag1:hostpath1 mounttag2:hostpath2 mounttag3:hostpath3"

virtc/main.go
```diff
@@ -100,6 +100,12 @@ var podConfigFlags = []cli.Flag{
 		Value: "",
 		Usage: "the hyperstart tty socket type",
 	},
+
+	cli.StringSliceFlag{
+		Name:  "volume",
+		Value: new(cli.StringSlice),
+		Usage: "the volume to be shared with VM",
+	},
```
sameo: Here you should declare that one as a `cli.GenericFlag` and then get it directly as a `Volume`. See how virtc/main.go does that for `agentType` for example.

agent.go
```diff
@@ -127,12 +127,15 @@ type agent interface {
 	// start will start the agent on the host.
 	start() error
 
+	// isStarted will check the agent is started.
+	isStarted() bool
```
sameo: I'd prefer that one to be more generic: `verifyState() agentStateString`. And we should update and potentially store the new agent state based as part of that function.

agent.go
```diff
@@ -127,12 +127,15 @@ type agent interface {
...
+
 	// exec will tell the agent to run a command in an already running container.
 	exec(podID string, contID string, cmd Cmd) error
 
-	// StartPod will tell the agent to start all containers related to the Pod.
+	// startPod will tell the agent to start all containers related to the Pod.
 	startPod(config PodConfig) error
 
-	// StopPod will tell the agent to stop all containers related to the Pod.
+	// stopPod will tell the agent to stop all containers related to the Pod.
```
sameo: Those 2 are separate changes, so I'd appreciate if you could put that into a separate patch as well.

pod.go
```diff
@@ -680,6 +681,22 @@ func (p *Pod) delete() error {
 	return nil
 }
 
+func (p *Pod) startAgent() error {
+	if p.state.AgentState != agentStarted {
```
sameo: A more idiomatic code style:
```
if p.state.AgentState == agentStarted {
       return nil
}

err := p.agent.start()
[...]
```

hyperstart.go
```diff
@@ -361,6 +361,16 @@ func (h *hyper) start() error {
 	return nil
 }
 
+// isStarted is the agent state checking implementation for hyperstart.
+func (h *hyper) isStarted() bool {
```
sameo: We probably want to return `agentStopped` if the agent state is stopped.

pod.go
```diff
@@ -682,7 +682,7 @@ func (p *Pod) delete() error {
 }
 
 func (p *Pod) startAgent() error {
-	if p.state.AgentState != agentStarted {
+	if p.state.AgentState != agentStarted || p.agent.isStarted() == false {
```
sameo: Could we just call `if p.agent.verifyState() != agentStarted {` here ?

sameo:
@sboeuf Please provide a more descriptive pull request title.

pod.go
```diff
@@ -135,10 +136,50 @@ func (state *PodState) validTransition(oldState stateString, newState stateStrin
 // defined by its mount tag and its host path.
 type Volume struct {
 	// MountTag is a label used as a hint to the guest.
-	MountTag string
+	mountTag string
 
 	// HostPath is the host filesystem path for this volume.
-	HostPath string
+	hostPath string
```
sameo: We need to keep those exported as it is part of our `PodConfig` structure.

pod.go
```diff
@@ -135,10 +136,50 @@ func (state *PodState) validTransition(oldState stateString, newState stateStrin
...
+}
+
+// Volumes is a Volume list.
+type Volumes []Volume
+
+// Set assigns volume values from string to a Volume.
+func (v *Volumes) Set(volStr string) error {
+	volArray := strings.Split(volStr, " ")
```
sameo: Nitpicking here: `volSlice` would be a more accurate name.

hyperstart.go
```diff
@@ -315,6 +347,38 @@ func buildHyperContainerProcess(cmd Cmd) (hyperJson.Process, error) {
 	return process, nil
 }
 
+func isStarted(c net.Conn, chType chType) bool {
+	ret := false
+	timeoutDuration := 5 * time.Second
```
sameo: 5s is too much imho. If we can't ping in less than a second, we're not connected.

hyperstart.go
```diff
@@ -400,10 +468,17 @@ func (h *hyper) startPod(config PodConfig) error {
 		containers = append(containers, container)
 	}
 
+	var mountTag string
+	if h.config.Volumes != nil {
+		mountTag = h.config.Volumes[0].mountTag
+	} else {
+		mountTag = ""
+	}
+
```
sameo: Not sure this chunk is related to this patch ?

hyperstart.go
```diff
@@ -315,6 +347,38 @@ func buildHyperContainerProcess(cmd Cmd) (hyperJson.Process, error) {
 	return process, nil
 }
 
+func isStarted(c net.Conn, chType chType) bool {
```
sameo: We probably only need to check for one channel only. I can't see a case where we'd have only one channel alive and not the other.

Issue #9, opened 2016-10-17
Complete hyperstart agent implementation


sameo:
Implement the agent interface for [hyperstart](https://github.com/hyperhq/hyperstart).

PR #19, opened 2016-10-24
Start/Stop a pod from virtc command line

sboeuf:
With those patches, we are able to start/stop a pod, using the virtc command line. It uses Qemu + Hyperstart combination.
It fixes issue #9 as we are able to use the hyperstart protocol.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -127,4 +127,7 @@ type agent interface {
 
 	// stopPod will tell the agent to stop all containers related to the Pod.
 	stopPod(config PodConfig) error
+
+	// stop will stop the agent on the host.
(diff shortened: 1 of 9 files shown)


sameo:
@sboeuf `Sockets` is more of a hyperstart specific configuration rather than a pod one, so I'd prefer not having them in the pod config. So I'd prefer if we'd do the following:

- Keep the `Sockets` definition in pod.go as it might be useful to other components
- Remove 'Sockets` from the pod configuration structure
- Extend `addDevice()` to support socket devices
- Call `addDevice()` from hyperstart to add the hyperstart sockets

Issue #3, opened 2016-10-17
Implement StatusPod entry point


sameo:
`StatusPod()` should take a pod ID and returns the Pod details (Including its containers statuses) and state.

PR #21, opened 2016-10-24
Add StatusPod entry point and introduce containers state

sboeuf:
This merge request first adds the new StatusPod entry point, and then it introduces states related to containers. Indeed, the user wants to know about pod and containers states when it will call StatusPod().
It fixes issue #3

### DIFF
--- a/api.go
+++ b/api.go
@@ -149,6 +149,7 @@ func RunPod(podConfig PodConfig) (*Pod, error) {
 }
 
 var listFormat = "%s\t%s\t%s\t%s\n"
+var statusFormat = "%s\t%s\n"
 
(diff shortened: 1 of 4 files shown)


sameo:
@sboeuf Why are the container states different than the pod ones ? For example when a pod is created it's in the READY state, and so should be the containers that are part of it. Am I missing something ?

sboeuf:
@sameo I thought that a container can be considered as READY when the pod is running. If the pod is not running, we can't really create a container, meaning it can't be ready.

sameo:
@sboeuf I did not mean that pods and containers should always be in the same state, but rather than the set of defined states for pods and containers should be identical, and that the state machine should be the same as well.

Issue #1, opened 2016-10-17
Support VM resource constraint


sameo:
We want to be able to constraint our VM to a number of CPUs and a certain amount of RAM. This should be described through the Pod configuration.

PR #23, opened 2016-10-25
Specify VM resources for each pod configuration

sboeuf:
Adding a new HardwareConfiguration structure in the existing PodConfig
allows to specify resources we need for every VM running a pod.
It fixes issue #1

### DIFF
--- a/pod.go
+++ b/pod.go
@@ -286,6 +286,20 @@ type Cmd struct {
 	Group string
 }
 
+// HardwareConfig describes VM resources configuration.
+type HardwareConfig struct {
+	// Related to processor resources
+	CPUs    string
+	Cores   string
+	Sockets string
+	Threads string
+
+	// Related to memory resources
+	MemSize  string
+	MemSlots string
+	MemMax   string
+}
+
 // ContainerConfig describes one container runtime configuration.
 type ContainerConfig struct {
 	ID string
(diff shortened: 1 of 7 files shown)


pod.go
```diff
@@ -286,6 +286,20 @@ type Cmd struct {
...
+	Cores   string
+	Sockets string
+	Threads string
+
+	// Related to memory resources
+	MemSize  string
+	MemSlots string
+	MemMax   string
+}
+
```
sameo: This is a `Resources` structure and it should not be so close to a qemu definition. It should be more high level:

```
type Resources struct {
        VCPUs int

        // Memory is the amount of available memory on mega bytes.
        Memory int
}
```

Then it is up to the selected hypervisor to convert that into some option/command.

qemu.go
```diff
@@ -66,6 +67,19 @@ const (
 
 const defaultQemuPath = "/usr/bin/qemu-system-x86_64"
 
+const (
+	defaultCPUs    = 2
+	defaultCores   = 1
+	defaultSockets = 2
+	defaultThreads = 1
+)
+
```
sameo: So our internal benchmarks showed that we really want 1 socket only. Also the default CPUs should map the host cpu count and not a hard coded number. Basically we want the qemu smp command line to look like:
```
"-smp %d,sockets=1,cores=%d,threads=1", vCPUs"
```

And vCPUs is either the passed CPUs resource or the host cpu count if the former is 0.

virtc/main.go
```diff
@@ -119,6 +119,48 @@ var podConfigFlags = []cli.Flag{
...
+		Name:  "vm-mem-slots",
+		Value: "",
+		Usage: "the number of memory slots available for this pod",
+	},
+
+	cli.StringFlag{
+		Name:  "vm-mem-max",
+		Value: "",
+		Usage: "the maximum amount of memory available for this pod",
+	},
```
sameo: Here we only want to pass a number of vCPUs and an amount of memory in MB.

PR #70, opened 2016-12-02
hyperstart: Don't special-case destroypod

dlespiau:
We're supposed to be able to receive an ack message after a destroypod.
The fact that we didn't was caused by our use of hyperstart as a daemon
(vs an init process) and some bugs introduced by that.

With those fixed, we have the destroypod ack message back.


dlespiau:
Note that for this to work our hyperstart needs the last two commits of (they will find their way into our package but one can also compile hyperstart from that branch for testing purposes):

https://github.com/dlespiau/hyperstart/tree/20161201-intel

We'll probably need to enhance hyperstart.go to support that, so opened issue #https://github.com/containers/virtcontainers/issues/71.

PR #72, opened 2016-12-03
Implement network isolation

sboeuf:
This PR starts the implementation of issue #2, that is the networking isolation for our pods.

### DIFF
--- a/api.go
+++ b/api.go
@@ -25,15 +25,24 @@ import (
 // CreatePod is the virtcontainers pod creation entry point.
 // CreatePod creates a pod and its containers. It does not start them.
 func CreatePod(podConfig PodConfig) (*Pod, error) {
-	// Create the pod.
-	p, err := createPod(podConfig)
+	// Create the network.
+	n := newNetwork(podConfig.NetworkModel)
+	networkNS, err := n.add(&(podConfig.NetworkConfig))
 	if err != nil {
+		return nil, err
+	}
+
+	// Create the pod.
+	p, err := createPod(podConfig, networkNS)
+	if err != nil {
+		n.remove(networkNS)
 		return nil, err
 	}
 
 	// Store it.
 	err = p.storePod()
 	if err != nil {
+		n.remove(p.networkNS)
 		return nil, err
 	}
 
@@ -56,6 +65,13 @@ func DeletePod(podID string) (*Pod, error) {
 
 	// Fetch the pod from storage and create it.
 	p, err := fetchPod(podID)
+	if err != nil {
+		return nil, err
+	}
+
+	// Remove the network.
+	n := newNetwork(p.config.NetworkModel)
+	err = n.remove(p.networkNS)
 	if err != nil {
 		return nil, err
 	}
@@ -87,6 +103,13 @@ func StartPod(podID string) (*Pod, error) {
 
 	// Fetch the pod from storage and create it.
 	p, err := fetchPod(podID)
+	if err != nil {
+		return nil, err
+	}
+
+	// Join the network.
+	n := newNetwork(p.config.NetworkModel)
+	err = n.join(p.networkNS)
 	if err != nil {
 		return nil, err
 	}
@@ -138,15 +161,24 @@ func StopPod(podID string) (*Pod, error) {
 // RunPod is the virtcontainers pod running entry point.
 // RunPod creates a pod and its containers and then it starts them.
 func RunPod(podConfig PodConfig) (*Pod, error) {
-	// Create the pod.
-	p, err := createPod(podConfig)
+	// Create the network.
+	n := newNetwork(podConfig.NetworkModel)
+	networkNS, err := n.add(&(podConfig.NetworkConfig))
 	if err != nil {
+		return nil, err
+	}
+
+	// Create the pod.
+	p, err := createPod(podConfig, networkNS)
+	if err != nil {
+		n.remove(networkNS)
 		return nil, err
 	}
 
 	// Store it.
 	err = p.storePod()
 	if err != nil {
+		n.remove(p.networkNS)
 		return nil, err
 	}
 
@@ -155,6 +187,12 @@ func RunPod(podConfig PodConfig) (*Pod, error) {
 		return nil, err
 	}
 	defer unlockPod(lockFile)
+
+	// Join the network.
+	err = n.join(p.networkNS)
+	if err != nil {
+		return nil, err
+	}
 
 	// Start it.
 	err = p.start()
--- a/cni.go
+++ b/cni.go
@@ -16,6 +16,10 @@
 
 package virtcontainers
 
+import (
+	"github.com/containernetworking/cni/pkg/ns"
+)
+
 // cni is a network implementation for the CNI plugin.
 type cni struct {
 }
@@ -62,6 +66,13 @@ func (n *cni) add(config *NetworkConfig) (NetworkNamespace, error) {
 		return NetworkNamespace{}, err
 	}
 
+	for _, endpoint := range networkNS.Endpoints {
+		err = bridgeNetworkPair(endpoint.NetPair)
+		if err != nil {
+			return NetworkNamespace{}, err
+		}
+	}
+
 	return networkNS, nil
 }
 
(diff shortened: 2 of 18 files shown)


network.go
```diff
@@ -0,0 +1,365 @@
...
+
+// NetworkConfig is the network configuration related to a network.
+type NetworkConfig struct {
+	PluginName string
+
+	NetworkNS     string
+	NetworkNSPath string
+
+	NetworkPairs []NetworkInterfacePair
+}
```
sameo: So our `NetworkConfig` should only contain:

* A networking model string ("CNI", "CNM" or "none") e.g. `NetworkModel`
* A networking namespace path. If left empty, this means we need to create one for the container. We could call it `NetNSPath`.

We may need to add a number of interfaces to create, but for now let's assume we want to create only one.

The network pairs is something internal to virtcontainers and should not be exposed. Virtcontainers will build a list of such pairs by scanning the networking namespace or by getting e.g. the CNI plugin information back.

### `NetNSPath`
The first things we need to do related to networking when creating a Pod is to potentially create a networking namespace and have the VM pid join that namespace.
If `NetNSPath` is empty, we will create a persitent networking namespace, i.e. not one linked to the current PID, and then join it. Otherwise we will simply join it.

### `NetworkModel`
This one will let us know what to look for in order to set the netns up:
- "CNI": We will have to look for the default CNI plugin to use, see what [ocicni](https://github.com/rajatchopra/ocicni/blob/master/ocicni.go#L76) does for example. Then:
    1. We call that plugin and the localhost one on the networking namespace we have for the pod
    2. We get the networking information back from CNI (networking setup and interface name)
    3. We create a bridge and a TAP interface, and we bridge the CNI interface and the TAP one through the created bridge.
    4. We add the tap interface information to both our kernel command line and our QEMU parameters
- "CNM": We will need to look for a libnetworking hook. TBD, let's postpone that one.
- "none": We will not set anything related to networking, at all.

network.go
```diff
@@ -0,0 +1,365 @@
...
+	if err != nil {
+		return fmt.Errorf("destroy bridge %v", err)
+	}
+
+	*tapInterface = NetworkInterface{}
+
+	return nil
+}
+
+func createBridge() (*netlink.Bridge, string, error) {
```
sameo: Could we use [libsnnet](https://github.com/01org/ciao/tree/master/networking/libsnnet) here ?

network.go
```diff
@@ -0,0 +1,365 @@
...
+
+	err = netlink.LinkSetMaster(tapl, brl)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func addVirtInterfaceToBridge(virtInterface NetworkInterface, brl *netlink.Bridge) error {
```
sameo: There should not be any difference between adding a TAP and a regular interface to a bridge.

network.go
```diff
@@ -0,0 +1,365 @@
...
+
+	brl, ok := link.(*netlink.Bridge)
+	if !ok {
+		return nil, fmt.Errorf("incorrect interface type %s %v", linkName, link.Type())
+	}
+
+	return brl, nil
+}
+
+func addTAPInterfaceToBridge(tapInterface NetworkInterface, brl *netlink.Bridge) error {
```
sameo: Is there anything specific about the interface being a TAP one ? If not, can we rename that one to addInterfaceToBridge() ?
Would even be better to create a bridge structure, and use methods on it:
```
func (b *Bridge) addInterface(if netInterface) error
```

sameo: Or enhance libsnnet to do that...

pod.go
```diff
@@ -239,6 +239,9 @@ type PodConfig struct {
 	AgentType   AgentType
 	AgentConfig interface{}
 
+	NetworkType   NetworkType
```
sameo: I don't think we need `NetworkType` here.

network.go
```diff
@@ -0,0 +1,365 @@
...
+	// and it creates and bridges TAP interfaces for the CNI network.
+	create(pod *Pod) error
+
+	// set switches the current process to the specified network namespace.
+	set(pod *Pod) error
+
+	// delete unbridges and deletes TAP interfaces. It also removes virtual network
+	// interfaces and deletes the network namespace.
+	delete(pod *Pod) error
+}
```
sameo: I'd personnaly prefer:

```
type network struct {
        add()
        remove()
        join()
}
```

pod.go
```diff
@@ -383,6 +390,16 @@ func createPod(podConfig PodConfig) (*Pod, error) {
 		state:      State{},
 	}
 
+	// Create network only if the namespace is empty in our configuration.
+	// Being empty means this is the first time we create the network for
+	// this pod.
+	if podConfig.NetworkConfig.NetworkNSPath == "" {
```
sameo: We will always have to call `network.create()` even if `NetNSPath` is not empty, because `create()` will always have to set the neworking namespace up.

sboeuf: I get it, but we don't want to call network creation every time we are in createPod() because createPod() is called from fetchPod(), meaning it is called from every API.
Maybe we could call it from CreatePod() in api.go instead ?

pod.go
sboeuf: Why ? Don't we want to find back the right network interface handler after a fetchPod() ?

network.go
sboeuf: Sounds good to me, I will change that.

network.go
sboeuf: Yes now we can. I have just submitted one PR to ciao to export API from bridge.go.

network.go
sboeuf: libsnnet already handle this and we can now use it because I exported more functions from vnic.go.

network.go
sboeuf: I have just exported more things from libsnnet (vnic.go) so as to do that.

network.go
```diff
@@ -0,0 +1,365 @@
...
+	default:
+		return &noopNetwork{}
+	}
+}
+
+// genRandomInterface generates locally a unique interface name
+func genRandomInterface(prefix string) (string, error) {
+	var random *rand.Rand
+	for i := 0; i < 100; i++ {
+		iface := fmt.Sprintf("%s_%x", prefix, random.Uint32())
```
sboeuf: We don't need this function anymore, we will rely completely on libsnnet package instead.

plugin/network/cni/cni.go
```diff
@@ -0,0 +1,167 @@
+//
```
sameo: The code looks fine, but I would prefer to move it under `network/cni` and not `plugin/network/cni`

qemu.go
```diff
@@ -533,6 +547,9 @@ func (q *qemu) addDevice(devInfo interface{}, devType deviceType) error {
 	case serialPortDev:
 		socket := devInfo.(Socket)
 		q.qemuConfig.Devices = q.appendSocket(q.qemuConfig.Devices, socket)
+	case netDev:
+		netPair := devInfo.(NetworkInterfacePair)
+		q.qemuConfig.Devices = q.appendNetwork(q.qemuConfig.Devices, netPair)
 	default:
 		break
 	}
```
sameo: We also need to add the networking parameters to the kernel command line.

api.go
```diff
@@ -31,6 +31,12 @@ func CreatePod(podConfig PodConfig) (*Pod, error) {
 		return nil, err
 	}
 
+	// Create the network.
+	err = p.network.add(p)
```
sameo: Adding the network should happen *before* we create the pod.

sameo: Some more details: We really want the networking namespace to be created and ready before we create the pod. That has several implications:

1. `network.add()` should not take a pod as a parameter, but a network config and an ID.
2. `createPod()` should take the list of created networking pairs as an additional parameter.

api.go
```diff
@@ -31,6 +31,12 @@ func CreatePod(podConfig PodConfig) (*Pod, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
 	// Store it.
 	err = p.storePod()
 	if err != nil {
```
sboeuf: done

network.go
```diff
@@ -0,0 +1,392 @@
...
+	TAPIface  NetworkInterface
+}
+
+// NetworkInterfacePairs defines a list of NetworkInterfacePair.
+type NetworkInterfacePairs []NetworkInterfacePair
+
+// NetworkConfig is the network configuration related to a network.
+type NetworkConfig struct {
+	NetNSPath  string
+	NumOfPairs int
```
sboeuf: done

cni.go
```diff
@@ -0,0 +1,138 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// add creates a new network namespace and its virtual network interfaces,
+// and it creates and bridges TAP interfaces for the CNI network.
+func (n *cni) add(config *NetworkConfig) ([]NetworkInterfacePair, error) {
```
sameo: I'd like this one to return more details, i.e. a complete networking namespace description.

```Go
type Endpoint struct {
        netPair    NetworkInterfacePair
        properties types.Result
}

type NetworkNamespace struct {
       netNsPath string
       endpoints  []Endpoint
}
```

I know this is bit CNI centric, but `types.Result` is actually network generic, and we can build it from a networking namespace scan with CNM.

So `add()` would return a `NetworkNamespace` and we would add that to the created pod.
The storing the pod would be split into storing the config and storing the whole `NetworkNamespace` into 2 different files.

sameo:
@sboeuf LGTM, please go ahead and fix the unit tests and we'll merge that beast.

PR #73, opened 2016-12-06
Make the hyperstart package follow upstream changes and fix ctl panic

dlespiau:
The first commit is to fix the hyperstart package as upstream has renamed a payload name, breaking the compilation.

More importantly, the second commit fixes a panic that can occur when killing the shim.

Note that travis fails because it got gets the current hyperstart package, which doesn't have the fix.


sameo:
@dlespiau Fix the hyperstart upstream support with PR #74. Please rebase this one on top of it and Travis should be happy.

dlespiau:
Superseded by #76

PR #75, opened 2016-12-06
Fix broken build

sboeuf:
Build was broken after some changes to hyperstart.go.


sboeuf:
Build was not broken. I had not synchronized locally the recent changes to runv/hyperstart/api/json.

PR #80, opened 2017-01-02
Update virtcontainers project to work with latest hyperstart changes

sboeuf:
Several changes to hyperstart API and pod lifecycle forced us to modify our library to continue to work properly.
Pause container issue #83.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,25 @@
+all: binaries
+	go build ./...
+	cd hack/virtc && go build && cd ../..
+
+pause:
+	make -C $@
+
+binaries: pause
+
+clean:
+	make -C pause clean
+	rm hack/virtc/virtc
+
+install:
+	install -D -m 755 pause/pause /usr/bin/pause
+
+uninstall:
+	rm -f /usr/bin/pause
+
+.PHONY: \
+	binaries \
+	clean \
+	install \
+	pause \
+	uninstall
--- a/hack/virtc/README.md
+++ b/hack/virtc/README.md
@@ -26,12 +26,12 @@ $ sudo apt install linux-container
 #### Get your image
 
 It has to be a recent Clear Linux image to make sure it contains hyperstart binary.
-You can dowload the following tested [image](https://download.clearlinux.org/releases/11130/clear/clear-11130-containers.img.xz), or any version more recent.
+You can dowload the following tested [image](https://download.clearlinux.org/releases/12210/clear/clear-12210-containers.img.xz), or any version more recent.
 
 ```
-$ wget https://download.clearlinux.org/releases/11130/clear/clear-11130-containers.img.xz
-$ unxz clear-11130-containers.img.xz
-$ sudo cp clear-11130-containers.img /usr/share/clear-containers/clear-containers.img
+$ wget https://download.clearlinux.org/releases/12210/clear/clear-12210-containers.img.xz
+$ unxz clear-12210-containers.img.xz
+$ sudo cp clear-12210-containers.img /usr/share/clear-containers/clear-containers.img
 ```
 
 #### Get virtc
@@ -41,10 +41,16 @@ _Download virtc_
 $ go get github.com/containers/virtcontainers
 ```
 
-_Build virtc_
+_Build and install pause binary_
 ```
 $ cd $GOPATH/src/github.com/containers/virtcontainers
-$ cd virtc && go build
+$ make
+$ make install
+```
+
+_Go to virtc_
+```
+$ cd hack/virtc
 ```
 
 ### Run virtc
@@ -67,11 +73,11 @@ $ echo -e 'HOME=/root\nPATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/s
 
 #### Run a new pod (Create + Start)
 ```
-./virtc pod run --bundle="/tmp/bundles/busybox/rootfs" --agent="hyperstart" --init-cmd="/bin/sh"
+./virtc pod run --agent="hyperstart" --network="CNI"
 ```
 #### Create a new pod
 ```
-./virtc pod create --bundle="/tmp/bundles/busybox/rootfs" --agent="hyperstart" --init-cmd="/bin/sh"
+./virtc pod create --agent="hyperstart" --network="CNI"
 ```
 This should generate that kind of output
 ```
@@ -98,7 +104,6 @@ POD ID                                  STATE   HYPERVISOR      AGENT
 306ecdcf-0a6f-4a06-a03e-86a7b868ffc8    running qemu            hyperstart
 
 CONTAINER ID    STATE
-1               ready
 ```
 
 #### Delete an existing pod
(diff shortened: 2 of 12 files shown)


container.go
```diff
@@ -32,12 +32,6 @@ type ContainerConfig struct {
 	// RootFs is the container workload image on the host.
 	RootFs string
 
-	// Interactive specifies if the container runs in the foreground.
-	Interactive bool
-
-	// Console is a console path provided by the caller.
-	Console string
-
```
sameo: Could you please explain why we can't have a tty per container ?
`Interactive` might not make sense, but tty for stderr/out and signaling is relevant on a per containers basis.

pod.go
```diff
@@ -242,11 +242,11 @@ type PodConfig struct {
...
-	// Rootfs is the pod root file system in the host.
-	// This can be left empty if we only have a set of containers
-	// workload images and expect the agent to aggregate them into
-	// a pod from the guest.
-	RootFs string
+	// Interactive specifies if the pod runs in the foreground.
+	Interactive bool
+
+	// Console is a console path provided by the caller.
+	Console string
```
sameo: It seems to me we're mixing 2 different changes from the same commit, which makes things a little difficult to review, and will not look nice on our git history.
Could you please separate things out a little better ?

container.go
sboeuf: You're right, we can do per container tty console, but I noticed some errors in the code, preventing us from doing such a thing. Everything was pod oriented, meaning in case of several containers, it would not have worked as expected. And in the same time, I have seen that interactive was specified at container level but it should be specified at pod level. I was thinking about fixing the tty console for each container in a future commit.
But I can also remove what I have removed and instead, I can fix it in another commit later (I mean the console thing), but we keep the Interactive change.

pod.go
sboeuf: Ah yes sorry, I have made lots of different changes and I didn't split the commits very well.

hyperstart.go
```diff
@@ -139,16 +143,53 @@ func (h *hyper) buildHyperContainerProcess(cmd Cmd) (*hyperJson.Process, error)
...
 
-	return bindMount(container.RootFs, rootfsDest)
+	linkPauseBinPath := filepath.Join(rootfsDest, rootfsDir, h.buildPauseBinName(pauseBinName, h.pod.id, container.ID))
+
+	_, err = os.Stat(linkPauseBinPath)
+	if err == nil {
+		return nil
+	}
+
+	return bindMount(h.config.PauseBinPath, linkPauseBinPath)
```
sameo: I honestly don't understand why we need to do this second bind mount. Please have a look at what I have in mind with the pause container here:

https://github.com/containers/virtcontainers/issues/83

hyperstart.go
```diff
@@ -334,14 +373,57 @@ func (h *hyper) startContainer(pod Pod, contConfig ContainerConfig) error {
 		return err
 	}
 
+	err = h.exec(pod, contConfig, contConfig.Cmd)
```
sameo: Let me try to understand here: When starting an hyperstart container, we're still sending the `NEWCONTAINER` command to hyperstart but now we're also exec'ing a `pause-podID-containerID` ?
I don't think we need this at all, but please prove me wrong.

hyperstart.go
```diff
@@ -334,14 +373,57 @@ func (h *hyper) startContainer(pod Pod, contConfig ContainerConfig) error {
...
+
+	if !restartPod {
+		return nil
+	}
+
+	noContPodConfig := *(pod.config)
+	noContPodConfig.Containers = []ContainerConfig{}
 
-	_, err := h.hyperstart.SendCtlMessage(hyperstart.KillContainer, payload)
+	err = h.startPod(noContPodConfig)
```
sameo: We don't want to have to restart the pod, we simply want it to stay alive until we explicitly destroy/stop it. That's why we want to have one additional container, the pause container, that never terminates.
See #83 for a little more details.

container.go
sameo: @sboeuf I understand, thanks. I'd prefer if you create a separate PR for the tty and interactive fixes, together with an issue to track the problem.

pod.go
sameo: No worries.

hyperstart.go
sboeuf: I don't know why but I thought it was not possible to execute a workload on a container if this one was not running with a pause binary. I have just made a test and I realized that it was working fine as long as I have at least one container (the pause container) running. That's why I was starting every container with the pause binary first.

hyperstart.go
sboeuf: Yes sure, I will do that. I had designed it that way because I thought every container needed to be run with a pause binary, and there was no point in adding an extra pause container.

sboeuf:
Changes done !

hyperstart.go
```diff
@@ -43,10 +47,11 @@ const (
...
 type HyperConfig struct {
-	SockCtlName string
-	SockTtyName string
-	Volumes     []Volume
-	Sockets     []Socket
+	SockCtlName  string
+	SockTtyName  string
+	Volumes      []Volume
+	Sockets      []Socket
+	PauseBinPath string
```
sameo: The whole pause (container, binary, path, etc...) should be fully transparent to the caller.
In other words, we (The virtcontainers build system) should build it and install it, and the caller should not handle any of this. It probably means that we want to have a simple Makefile for building and installing the pause binary and to build the library itself.

container.go
```diff
@@ -88,6 +87,17 @@ func fetchContainer(pod *Pod, containerID string) (*Container, error) {
 	return createContainer(pod, config)
 }
 
+// fetchContainerState fetches a container state from a pod and container ID and return
```
sboeuf: Not needed anymore, it was used by my previous version of stopContainer().

Makefile
```diff
@@ -0,0 +1,23 @@
+all: binaries
+	go build ./...
+
+pause:
+	make -C $@
+
+binaries: pause
```
sameo: Can we also build virtc here as well ?

hack/virtc/main.go
```diff
@@ -130,15 +118,13 @@ var podConfigFlags = []cli.Flag{
...
-	console := context.String("console")
-	bundle := context.String("bundle")
 	sshdUser := context.String("sshd-user")
 	sshdServer := context.String("sshd-server")
 	sshdPort := context.String("sshd-port")
 	sshdKey := context.String("sshd-auth-file")
 	hyperCtlSockName := context.String("hyper-ctl-sock-name")
 	hyperTtySockName := context.String("hyper-tty-sock-name")
-	initCmd := context.String("init-cmd")
+	hyperPauseBinPath := context.String("hyper-pause-bin-path")
```
sameo: Let's just call that one `--pause-path`

hack/virtc/README.md
```diff
@@ -41,9 +41,15 @@ _Download virtc_
...
-_Build virtc_
+_Build and install pause binary_
 ```
 $ cd $GOPATH/src/github.com/containers/virtcontainers
+$ make
+$ make install
+```
+
+_Build virtc_
+```
```
sameo: If we add virtc to our default built binaries, we would just call make and make install.

PR #81, opened 2017-01-02
Add benchmarks

sboeuf:
Those benchmarks measure the time it takes to create, start, and stop a pod.
Eventually, we should have start specific benchmarks, and stop ones.

### DIFF
--- a/api_test.go
+++ b/api_test.go
@@ -17,6 +17,7 @@
 package virtcontainers
 
 import (
+	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


sameo:
LGTM.
@sboeuf We will have to extend those benchmarks to measure:

- Pod start times only. See `b.StopTimer()`, `b.StartTimer()` and `b.ResetTImer()`
- Pod stop times only.
- Parallel starts. See `b.RunParallel()`

PR #86, opened 2017-01-11
Update hyperstart in virtcontainers to follow recent changes from hyperstart project

sboeuf:
Recent changes in hyperstart project introduced new commands, and particularly PROCESSASYNCEVENT which is returned on the CTL channel to return the exit code of a process running on a container.
This PR handles those changes by fixing "hyperstart" and "mock" packages in virtcontainers, and also by fixing virtcontainers itself.

### DIFF
--- a/hyperstart/hyperstart.go
+++ b/hyperstart/hyperstart.go
@@ -121,6 +121,8 @@ type Hyperstart struct {
 	// ctl access is arbitrated by ctlMutex. We can only allow a single
 	// "transaction" (write command + read answer) at a time
 	ctlMutex sync.Mutex
+
+	ctlMulticast *multicast
 }
 
 // NewHyperstart returns a new hyperstart structure.
@@ -149,6 +151,8 @@ func (h *Hyperstart) OpenSockets() error {
 	}
 	h.ioState.open()
 
+	h.ctlMulticast = startCtlMonitor(h.ctl)
+
 	return nil
 }
 
@@ -171,6 +175,8 @@ func (h *Hyperstart) CloseSockets() error {
 
 		h.ioState.close()
 	}
+
+	h.ctlMulticast = nil
 
 	return nil
 }
@@ -371,42 +377,32 @@ func codeFromCmd(cmd string) (uint32, error) {
 	return codeList[cmd], nil
 }
 
-func (h *Hyperstart) expectReadingCmd(conn net.Conn, code uint32) (*hyper.DecodedMessage, error) {
-	var msg *hyper.DecodedMessage
-	var err error
-
-	for {
-		msg, err = ReadCtlMessage(conn)
-		if err != nil {
-			return nil, nil
+func (h *Hyperstart) checkReturnedCode(recvCode, expectedCode uint32) error {
+	if recvCode != expectedCode {
+		if recvCode == hyper.INIT_ERROR {
+			return fmt.Errorf("ERROR received from Hyperstart")
 		}
 
-		if msg.Code == code {
-			break
-		}
-
-		if msg.Code == hyper.INIT_NEXT || msg.Code == hyper.INIT_READY {
-			continue
-		}
-
-		if msg.Code != code {
-			if msg.Code == hyper.INIT_ERROR {
-				return nil, fmt.Errorf("ERROR received from Hyperstart")
-			}
-
-			return nil, fmt.Errorf("CMD ID received %d not matching expected %d", msg.Code, code)
-		}
+		return fmt.Errorf("CMD ID received %d not matching expected %d", recvCode, expectedCode)
 	}
 
-	return msg, nil
+	return nil
 }
 
 // WaitForReady waits for a READY message on CTL channel.
 func (h *Hyperstart) WaitForReady() error {
-	h.ctlMutex.Lock()
-	defer h.ctlMutex.Unlock()
+	if h.ctlMulticast == nil {
+		return fmt.Errorf("No multicast available for CTL channel")
+	}
 
-	_, err := h.expectReadingCmd(h.ctl, hyper.INIT_READY)
+	channel, err := h.ctlMulticast.listen("", "", replyType)
+	if err != nil {
+		return err
+	}
+
+	msg := <-channel
+
+	err = h.checkReturnedCode(msg.Code, hyper.INIT_READY)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


dlespiau:
Chatting a bit with Sébastien, a few notes:

  -  We now have two methods blocking on read(ctl) so we can't do both, sending a message and waiting for an event
  - The end of process event can be sent between a command and the response (success status) of that command.

I think the low level hyperstart API has now to spawn a goroutine with a read loop on the ctl channel. Depending on what we read, either a response from a previous command or an event, we dispatch the paquet to either a channel to unlock SendCtlMessage() or to signal the event to the user of the library.

This could look like:

1. A new API call returning a channel on which the virtcontainers user will receive the process related events (and other events in the future?)
2. modify SendCtlMessage as:
    - write json packet to the ctl fd
    - wait onto an internal channel
    - receive command response through that channel
    - return response
3. A new internal goroutine handling all the reads on the ctl fd, it will
    - write reponses into the internal channel that SendCtlMessage waits on
    - write events to the channel defined in 1.

hyperstart/mock/hyperstart.go
```diff
@@ -257,6 +262,43 @@ func (h *Hyperstart) handleCtl() {
 
 		h.SendMessage(hyper.INIT_ACK, nil)
 
+		sendPAE := false
```
dlespiau: Instead of this, I would add a method simlar to SendExitStatus() that sends an event through the CTL channel so the client code can decide when it wants the process to end instead of an automatic end as soon as it has been created. That will allow testing multiple processes running at the same time.

sameo:
@dlespiau @sboeuf Being able to receive asynchronous events mixed with synchronous ones on the CTL port makes thing a little complex.

As @dlespiau adviced we probably now want to introduce the following API and interfaces:

- A CTL multicast interface. We need to be able to multicast events to whoever registered for listening for a events from a container, and forward CTL replies to the right listenr From this interface we need to be able to write events and replies, and register listeners:

```Go
type CtlDataType string

const (
         Event CtlDataType = "ctlEvent"
         Reply CtlDataType = "ctlReply"
)

type Multicaster interface {
        Write (p []byte) (n int, err error)
        Listen (ID string, type CtlDataType) (chan []byte, error)
}
```
A multicaster implementation would get its input from `Write()` and multicast the data down to its list of listener.

- A `StartCtlMonitor()` call that would do something like that:

```Go
cltMulticast := NewHyperMulticast()

go func() {
     for {
            data := read(ctlFD)
            ctlMulticast.Write(data)
    }
} ()

return ctlMulticast
```

- `SendCtlMessage()` would now take an additional parameter: `ctlMulticast *Multicaster` and would look like:

```Go
dataChannel := ctlMulticast.Listen(podID, Data)
writeCtlMessage(msg)
select {
        case data <- dataChannel:
                 return data
}
```

I'm not particularily happy about the SendCtlMessage() prototype, but I can't think about anything better atm.

sboeuf:
@sameo @dlespiau please could you both take a look at those changes to handle sync/async messages on the CTL channel.

hyperstart/multicast.go
```diff
@@ -0,0 +1,158 @@
...
+const (
+	eventType ctlDataType = "ctlEvent"
+	replyType ctlDataType = "ctlReply"
+)
+
+type multicast struct {
+	reply []chan *hyper.DecodedMessage
+	event map[string]chan *hyper.DecodedMessage
+	ctl   net.Conn
+	mutex sync.Mutex
```
dlespiau: A nice pattern is to use struct embedding here (just sync.Mutex) so the locking becomes multicast.Lock()

sboeuf: Thanks @dlespiau, I have just made the change ;)

hyperstart/multicast.go
```diff
@@ -0,0 +1,158 @@
...
+
+type ctlDataType string
+
+const (
+	eventType ctlDataType = "ctlEvent"
+	replyType ctlDataType = "ctlReply"
+)
+
+type multicast struct {
+	reply []chan *hyper.DecodedMessage
```
sameo: We have to restrict this to one single channel because the hyperstart implementation simply does not allow for more than one command to be processed at the same time. So we can only have one listener.
We could have a permanent channel, and an atomic counter to track its usage. If it's in use, anyone calling `listen` on the reply type would get an error back.

hyperstart/multicast.go
```diff
@@ -0,0 +1,158 @@
...
+	m.Lock()
+	for {
+		chanListLen = len(m.reply)
+		if chanListLen > 0 {
+			break
+		}
+		m.Unlock()
+		time.Sleep(time.Millisecond)
+		m.Lock()
+	}
```
sameo: The logic should be a little different here, as I understand you're trying to fix a potential race between an early reply and a late `listen`. `sendReply` should queue the reply if the reply channel is not used and return.
Then `listen` should always launch a go routine calling `sendReply` with the potentially queued reply.

hyperstart/multicast.go
```diff
@@ -0,0 +1,158 @@
...
+
+	replyChannel := m.reply[0]
+
+	tmpChanList := make([]chan *hyper.DecodedMessage, chanListLen-1)
+	copy(tmpChanList, m.reply[1:])
+	m.reply = tmpChanList
```
sameo: We do not need a list of channels anymore, but just for the record, I believe removing the first element from a slice could just be:

```Go
reply = reply[1:]
```

hyperstart/multicast.go
```diff
@@ -0,0 +1,158 @@
...
+		newChan := make(chan *hyper.DecodedMessage)
+
+		m.Lock()
+		m.reply = append(m.reply, newChan)
+		m.Unlock()
+
+		return newChan, nil
+	case eventType:
+		uniqueID := m.buildEventID(containerID, processID)
+		m.event[uniqueID] = make(chan *hyper.DecodedMessage)
```
sameo: You want to check if `m.event[uniqueID]` is empty. If it's not, you should return an error or else you'll have a memory leak.

hyperstart/hyperstart.go
```diff
@@ -425,26 +449,35 @@ func (h *Hyperstart) SendCtlMessage(cmd string, data []byte) (*hyper.DecodedMess
...
 	}
-	err = h.WriteCtlMessage(h.ctl, msg)
+	err = h.WriteCtlMessage(h.ctl, msgSend)
 	if err != nil {
 		return nil, err
 	}
 
-	// Wait for answer
-	resp, err := h.expectReadingCmd(h.ctl, hyper.INIT_ACK)
+	msgRecv := <-channel
```
sameo: As discussed on IRC, you can release the lock before waiting on the channel.

hyperstart/multicast.go
sboeuf: Nice simplification, I will do that!

hyperstart/multicast.go
sboeuf: Sure let me add this check.

hyperstart/multicast.go
sboeuf: Okay I think I see what you're explaining here. What do you think about sendReply() being called from write() and queueing the reply if needed (what you suggested). And another function called processQueueReply() called from listen().
So everything as you explained but split into 2 different functions for more clarity ?

sboeuf:
@sameo I have addressed your 3 comments, please take another look :)

sameo:
LGTM

Issue #85, opened 2017-01-09
Can not multiplex a pod containers process streams (stdin, stdout, stderr) with hyperstart


sameo:
When creating a new container process through hyperstart, we currenly pass it random `stdio` and `stderr` sequence numbers. Those sequence numbers are used to multiplex containers streams through the single hyperstart IO serial interface, but we do not use them currently. We simply lose all streams from all containers.

In order to e.g. demux all stdout streams from all running containers we should be able to keep a mapping of those sequence numbers to container IDs, and then forward each stream to a given file descriptor passed to us.

We basically have 2 solutions for this:

1) Handle the multiplexing/demultiplexing of stin/stdout and the forwarding part from `virtcontainers`

Or

2) Rely on an external component and have `virtcontainers` rely on it for handling all containers process streams.

Solution 1 seems to be outside of `virtcontainers` scope, solution 2 looks better.

## TODO ##

1) [ ] Handle the container `Console` field as a boolean and set hyperstart `Process` structure accordingly
2) [ ] Create a `Proxy` interface for connecting, disconnecting, attaching and possibly sending raw commands to it. It will be up to a `Proxy` implementation to interpret the raw commands and understand if they need to be encapsulated or not. In other words, the `virtcontainers` hyperstart agent implementation will send the same data to either a `Proxy` implementation or hyperstart itself.
3) [ ] Implement a `cc-proxy` support.

sameo:
cc @dlespiau

dlespiau:
I think we'd like https://github.com/01org/cc-oci-runtime/issues/342 for that one. The client code would only have to know about 2 fds (ctl & io) and the hyperstart protocol. That way, having a proxy or not will be effectively transparent (except the initialisation). As @sameo would say, a real proxy instead of encapsulation for the ctl commands.

sameo:
@dlespiau Yes. I was about to open a cc-oci-runtime issue for that, glad to see there's one already ;)
We can start working on this issue from `virtcontainers` and keep the whole encapsulation in our cc-proxy layer and change it when #342 is done.

sameo:
@dlespiau Also, on the sequence number allocations, we can decide to let the client generate it and let it know if it conflicts with an already existing allocation. This would simplify the protocol itself.

dlespiau:
we could indeed do that, one would have to be terribly unlucky to collide on the 64 bits seq number with only a few processes per VM. Just need to make sure we receive an error from hyperstart.

[...]

PR #87, opened 2017-01-13
Implement proxy

sboeuf:
Virtcontainers reached a point where it needs to rely on the hyperstart proxy if we want to go further. This PR solves issue #85 by implementing two proxy interfaces, a raw hyperstart intended to communicate directly  with hyperstart, and a real proxy responsible to do the abstraction layer with the VM.

### DIFF
--- a/hyperstart.go
+++ b/hyperstart.go
@@ -114,7 +114,7 @@ type hyperstartProxyCmd struct {
 	message interface{}
 }
 
-func (h *hyper) buildHyperContainerProcess(cmd Cmd, stdio uint64, stderr uint64) (*hyperJson.Process, error) {
+func (h *hyper) buildHyperContainerProcess(cmd Cmd, stdio uint64, stderr uint64, terminal bool) (*hyperJson.Process, error) {
 	var envVars []hyperJson.EnvironmentVar
 
 	for _, e := range cmd.Envs {
@@ -127,13 +127,14 @@ func (h *hyper) buildHyperContainerProcess(cmd Cmd, stdio uint64, stderr uint64)
 	}
 
 	process := &hyperJson.Process{
-		User:    cmd.User,
-		Group:   cmd.Group,
-		Stdio:   stdio,
-		Stderr:  stderr,
-		Args:    cmd.Args,
-		Envs:    envVars,
-		Workdir: cmd.WorkDir,
+		User:     cmd.User,
+		Group:    cmd.Group,
+		Terminal: terminal,
+		Stdio:    stdio,
+		Stderr:   stderr,
+		Args:     cmd.Args,
+		Envs:     envVars,
+		Workdir:  cmd.WorkDir,
 	}
 
 	return process, nil
@@ -244,7 +245,7 @@ func (h *hyper) exec(pod Pod, container Container, cmd Cmd) error {
 		return err
 	}
 
-	process, err := h.buildHyperContainerProcess(cmd, ioStream.StdoutID, ioStream.StderrID)
+	process, err := h.buildHyperContainerProcess(cmd, ioStream.StdoutID, ioStream.StderrID, container.config.Interactive)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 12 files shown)


sboeuf:
@sameo @dlespiau please take a look at this PR, this is the first step in getting the proxy for virtcontainers.

laijs:
adding proxy is also a job in runv
see the 5th item in https://github.com/hyperhq/runv/issues/415

hyperstart.go
```diff
@@ -136,11 +121,16 @@ func (h *hyper) buildHyperContainerProcess(cmd Cmd) (*hyperJson.Process, error)
 		envVars = append(envVars, envVar)
 	}
 
+	ioBase, _, err := h.proxy.allocateIO(2)
```
sameo: So `ioBase` and the `fd` needs to somehow be externally accessible for something like COR to be able to send that back to e.g. the shim.

sameo: We also need to store those as container resources.

proxy.go
```diff
@@ -0,0 +1,93 @@
...
+
+	// detach detaches current proxy connection from a previously registered VM.
+	detach() error
+
+	// allocateIO asks the proxy to allocate IO stream sequence numbers and to
+	// route the IO from hyperstart to the right socket.
+	allocateIO(nStreams int) (uint64, *os.File, error)
+
+	// sendCmd sends a command to the agent inside the VM through the proxy.
+	sendCmd(cmd string, message interface{}) error
```
sameo: Here we're assuming that none of the commands will ever send us an answer other than an error. Although this is true for `cc-proxy`, I'd like this one to be a little more generic and allow for returning an answer.

sameo: Also, `sendCmd()` should only take a message, not a `cmd` argument. Then it's up to our cc-proxy implementation to inspect that frame and extract the right cmd string from it.

hyperstart_proxy.go
```diff
@@ -0,0 +1,44 @@
+//
+// Copyright (c) 2016 Intel Corporation
```
sameo: Copyright (c) 2017, please :)

hyperstart_proxy.go
```diff
@@ -0,0 +1,44 @@
+//
```
sameo: Could we call that one cc_proxy.go, please ?

proxy.go
```diff
@@ -0,0 +1,103 @@
...
+	Stdin    *os.File
+	Stdout   *os.File
+	Stderr   *os.File
+	StdinID  uint64
+	StdoutID uint64
+	StderrID uint64
+}
+
+// proxy is the virtcontainers proxy interface.
+type proxy interface {
```
sameo: This interface is fine for now, but please note that it may change depending on further cc-proxy changes.

hyperstart_proxy.go
```diff
@@ -16,29 +16,141 @@
...
-	return []IOStream{}, nil
+	var err error
+	var ioStreams []IOStream
+
+	p.client, err = p.connectProxy()
+	if err != nil {
+		return []IOStream{}, err
+	}
+
+	hyperConfig, ok := newAgentConfig(*(pod.config)).(HyperConfig)
```
sameo: Note to ourselves: This should not be hyper specific.

hyperstart.go
```diff
@@ -193,17 +178,18 @@ func (h *hyper) bindUnmountAllRootfs() {
 }
 
 // init is the agent initialization implementation for hyperstart.
-func (h *hyper) init(pod Pod, config interface{}) error {
+func (h *hyper) init(pod *Pod, config interface{}) error {
```
sameo: Why do we need a pod pointer ? For this:

`pod.config.AgentConfig = h.config` ?

sboeuf: Yes because I realized that SockCtlName and SockTtyName are empty when I need those values from register() function from proxy interface. Indeed, if those are empty, the validate() function in init() will fill them, but then, because the pod is not a pointer, we have no way to keep them...

PR #89, opened 2017-01-19
Low level hyperstart read/write

dlespiau:
This PR exposes low level functions to read/write hyperstart commands, this will be needed for the proxy to emulate a ctl fd for each client.


sameo:
LGTM

PR #92, opened 2017-01-20
CNM network implementation

sboeuf:
This PR modifies the way the network interface was handled for the CNI implementation so that it can work for the CNM implementation. Then, it adds the OCI hooks handling, because we need to be able to handle those hooks for network setup. At last, it implements the CNM network interface.

### DIFF
--- a/api.go
+++ b/api.go
@@ -25,24 +25,15 @@ import (
 // CreatePod is the virtcontainers pod creation entry point.
 // CreatePod creates a pod and its containers. It does not start them.
 func CreatePod(podConfig PodConfig) (*Pod, error) {
-	// Create the network.
-	n := newNetwork(podConfig.NetworkModel)
-	networkNS, err := n.add(Pod{}, &(podConfig.NetworkConfig))
-	if err != nil {
-		return nil, err
-	}
-
 	// Create the pod.
-	p, err := createPod(podConfig, networkNS)
+	p, err := createPod(podConfig)
 	if err != nil {
-		n.remove(Pod{}, networkNS)
 		return nil, err
 	}
 
 	// Store it.
 	err = p.storePod()
 	if err != nil {
-		n.remove(Pod{}, p.networkNS)
 		return nil, err
 	}
 
@@ -65,13 +56,6 @@ func DeletePod(podID string) (*Pod, error) {
 
 	// Fetch the pod from storage and create it.
 	p, err := fetchPod(podID)
-	if err != nil {
-		return nil, err
-	}
-
-	// Remove the network.
-	n := newNetwork(p.config.NetworkModel)
-	err = n.remove(Pod{}, p.networkNS)
 	if err != nil {
 		return nil, err
 	}
@@ -107,9 +91,21 @@ func StartPod(podID string) (*Pod, error) {
 		return nil, err
 	}
 
-	// Join the network.
+	// Create the network.
 	n := newNetwork(p.config.NetworkModel)
-	err = n.join(p.networkNS)
+	networkNS, err := n.add(*p, &(p.config.NetworkConfig))
+	if err != nil {
+		return nil, err
+	}
+
+	// Store the network
+	err = p.storage.storePodNetwork(p.id, networkNS)
+	if err != nil {
+		return nil, err
+	}
+
+	// Join the network.
+	err = n.join(networkNS)
 	if err != nil {
 		return nil, err
 	}
@@ -150,6 +146,19 @@ func StopPod(podID string) (*Pod, error) {
 		return nil, err
 	}
 
+	// Fetch the network config
+	networkNS, err := p.storage.fetchPodNetwork(podID)
+	if err != nil {
+		return nil, err
+	}
+
+	// Remove the network
+	n := newNetwork(p.config.NetworkModel)
+	err = n.remove(*p, networkNS)
+	if err != nil {
+		return nil, err
+	}
+
 	err = p.endSession()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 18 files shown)


oci_hook.go
```diff
@@ -0,0 +1,126 @@
+//
```
sameo: Could we please call that file simply hook.go ?

oci_hook.go
```diff
@@ -0,0 +1,126 @@
...
+	PostStopHooks  []OciHook
+}
+
+func buildOciHookState(processID int) specs.State {
+	return specs.State{
+		Pid: processID,
+	}
+}
+
+func runOciHook(hook OciHook) error {
```
sameo: This could be a `Hook` method.

oci_hook.go
```diff
@@ -0,0 +1,126 @@
...
+	firstArg := hook.Args[0]
+	otherArgs := hook.Args[1:]
+	args := []string{"-a", firstArg, hook.Path}
+	args = append(args, otherArgs...)
+
```
sameo: You need to add the hook env variables here as well.

oci_hook.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return err
+	}
+
+	_, err = stdin.Write(stateJSON)
+	if err != nil {
+		return err
+	}
+
+	err = cmd.Wait()
```
sameo: This needs to be handled a little differently as we have to support timeouts as well. Something like:

```Go
if hook.Timeout == 0 {
      err = cmd.Wait
} else {
      done := make(chan error)
      go func() { done <- cmd.Wait() }()
      select {
      case err := <-done:
              if err != nil {
                        return err
              }
      case <-time.After(hook.Timeout):
              return fmt.Errorf("Hook timeout")
    }
}
```

oci_hook.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (h *OciHooks) preStartHooks() error {
+	if len(h.PreStartHooks) < 1 {
```
sameo: nitpick: `if len() == 0{` could be more explicit...

oci_hook.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (h *OciHooks) postStartHooks() error {
+	if len(h.PostStartHooks) < 1 {
+		return nil
+	}
+
+	for _, hook := range h.PostStartHooks {
+		err := runOciHook(hook)
+		if err != nil {
```
sameo: With posts start hooks, we only need to log the error and continue running the remaining hooks.

oci_hook.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (h *OciHooks) postStopHooks() error {
+	if len(h.PostStopHooks) < 1 {
+		return nil
+	}
+
+	for _, hook := range h.PostStopHooks {
+		err := runOciHook(hook)
+		if err != nil {
+			return err
```
sameo: Ditto.

oci_hook.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+// OciHook represents an OCI hook, including its required parameters.
+type OciHook struct {
+	Path string
+	Args []string
+}
+
+// OciHooks gathers all existing OCI hooks list.
+type OciHooks struct {
```
sameo: I'd prefer to call that structure `Hooks`

oci_hook.go
```diff
@@ -0,0 +1,126 @@
...
+import (
+	"encoding/json"
+	"os"
+	"os/exec"
+
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+// OciHook represents an OCI hook, including its required parameters.
+type OciHook struct {
```
sameo: I'd prefer to call that structure `Hook`.

sameo:
@mcastelino It would be great if you could have a look.

cnm.go
```diff
@@ -16,29 +16,147 @@
...
+	}
+
+	uniqueID := uuid.Generate().String()
+
+	for i, addr := range addrs {
+		var endpoint Endpoint
+
+		if addr.Network() == "lo" || reflect.DeepEqual(net.ParseIP(addr.String()), net.IPv4(127, 0, 0, 1)) {
+			continue
+		} else {
```
sboeuf: But I don't want to go through the code below (endpoint.Properties, err = n.createResult(addr)) in case I find the "lo" interface.

hook.go
```diff
@@ -0,0 +1,145 @@
...
+
+func (h *Hook) runHook() error {
+	state := buildHookState(os.Getpid())
+	stateJSON, err := json.Marshal(state)
+	if err != nil {
+		return err
+	}
+
+	var stdout, stderr bytes.Buffer
+	cmd := &exec.Cmd{
```
mcastelino: @sboeuf Is the PID in this state file correct. You need to pass the PID of a process that is in the network namespace you created. I think you are in the parent ns here. I may be mistaken though. Is this code working?

cnm.go
```diff
@@ -16,29 +16,140 @@
...
+	}
+
+	networkNS := NetworkNamespace{
+		NetNsPath: config.NetNSPath,
+		Endpoints: endpoints,
+	}
+
+	err = doNetNS(networkNS.NetNsPath, func(_ ns.NetNS) error {
+		for _, endpoint := range networkNS.Endpoints {
+			err := bridgeNetworkPair(endpoint.NetPair)
```
sboeuf: @mcastelino I have no problem in bringing something more generic, that's what we want for 3.0. But I don't know what to expect with SRIOV yet. Therefore, I would follow your suggestion that we can postpone this implementation improvement to the moment we introduce SRIOV.

hook.go
sboeuf: @mcastelino yes I tried the code and it works. Check api.go line 102:
err = n.join(p.config.NetworkConfig.NetNSPath)
It switches our process to the network namespace previously created. And then if I manually switch to this netns from my terminal, I can see that everything is created as expected.

api.go
```diff
@@ -27,22 +27,22 @@ import (
 func CreatePod(podConfig PodConfig) (*Pod, error) {
 	// Create the network.
 	n := newNetwork(podConfig.NetworkModel)
-	networkNS, err := n.add(&(podConfig.NetworkConfig))
+	networkNS, err := n.add(Pod{}, &(podConfig.NetworkConfig))
```
sboeuf: @mcastelino yes I agree the add() function creates some data and we want to know them from the caller, that's why I pass the NetworkConfig pointer. The pod is used here only as a value on purpose (it wouldn't hurt to have a pointer though). Regarding QEMU parameters, take a look at add() implementation, it calls an hypervisor (qemu in our case) API called AddDevice() to add the appropriate network configuration to qemu parameters.

cnm.go
sameo: @sboeuf @mcastelino Yes, let's postpone this generalization to the SRIOV support.

sameo:
@sboeuf LGTM but it's missing some unit tests for e.g. cnm.go.

sboeuf:
@sameo yes let me add some unit tests before we merge this ;)

PR #91, opened 2017-01-19
cc-proxy: Follow the latest API change

dlespiau:
Hopefully the first and only client API change before we tag 2.1.

References: https://github.com/01org/cc-oci-runtime/pull/595
Signed-off-by: Damien Lespiau <damien.lespiau@intel.com>

--

Of course, this PR is going to fail and the related PR in the runtime isn't merged yet.

PR #94, opened 2017-01-23
Fix build due to recent dependencies changes

sboeuf:
Our cni package relies on libcni library and they recently updated one function we were using, thus we provide the fix for that.
Then, proxy API from cc-oci-runtime changed recently due to a global re-design, meaning it had to change the Client API too. Because we use this Client API to communicate with cc-proxy, we have to fix it to allow our build to pass.

### DIFF
--- a/cc_proxy.go
+++ b/cc_proxy.go
@@ -77,7 +77,7 @@ func (p *ccProxy) register(pod Pod) ([]IOStream, error) {
 		return []IOStream{}, fmt.Errorf("Wrong agent config type, should be HyperConfig type")
 	}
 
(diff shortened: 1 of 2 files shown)


sameo:
LGTM

dlespiau:
Just wanted to say I had, of course, sent the PR to adapt to the new proxy changes (https://github.com/containers/virtcontainers/pull/91) in case @sboeuf has been cursing me over the API change :)

sboeuf:
@dlespiau Yeah I had noticed you had uploaded this but we also needed a fix for cni package relying on libcni to get the build completely fixed. That's why I pushed this PR with both the cni fix and your fix regarding the cc-proxy Client ;)

PR #97, opened 2017-01-27
Fix netns usage and improve network unit testing coverage

sboeuf:
We first needed to fix the way we were using the netns. Indeed, it is not safe to switch to a netns, but instead we should execute a block of code inside the specified namespace.
This allowed us to avoid netns confusion and mixing with CNI plugin, and we have been able to add a new unit test relying on CNI network.

### DIFF
--- a/network.go
+++ b/network.go
@@ -392,7 +392,7 @@ type network interface {
 	init(config *NetworkConfig) error
 
 	// join switches the current process to the specified network namespace.
-	join(networkNSPath string) error
+	join(networkNSPath string, cb func() error) error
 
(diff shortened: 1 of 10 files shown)


network.go
```diff
@@ -392,7 +392,7 @@ type network interface {
 	init(config *NetworkConfig) error
 
 	// join switches the current process to the specified network namespace.
-	join(networkNSPath string) error
+	join(networkNSPath string, cb func() error) error
```
sameo: I am fine with this patch, but then we should rename that method to `run` or `do`.

PR #105, opened 2017-02-01
Add a proxy configuration interface

sameo:
We want to be able to specify some proxy specific settings, like e.g. the proxy runtime socket path for our Clear Containers proxy implementation. Eventually, this will also include the proxy shim socket path.


sameo:
@sboeuf @dlespiau PTAL

dlespiau:
lgtm

PR #106, opened 2017-02-01
pkg: Move all packages under pkg/ directory

sboeuf:
It seems a standard to gather all packages under the pkg directory, so let's
do that for virtcontainers with cni, hyperstart and mock packages.
All files using those packages are updated accordingly.


sameo:
LGTM

PR #107, opened 2017-02-02
Move VM creation and network setup to CreatePod()

sboeuf:
In order to allow future performance savings with OCI runtime, this PR moves VM creation/deletion to CreatePod()/DeletePod(). It does the same for the network which is now created/removed from CreatePod()/DeletePod() too.


sameo:
LGTM

PR #110, opened 2017-02-03
hyperstart: Fix pod stop/restart behavior

sboeuf:
This patch allows to re-enable the start/stop pod. Indeed, after we
introduced the change to create and delete a VM from the CreatePod()
and DeletePod(), we were not able to start the pod after it was stopped.

### DIFF
--- a/cc_proxy.go
+++ b/cc_proxy.go
@@ -98,15 +98,6 @@ func (p *ccProxy) register(pod Pod) ([]IOStream, error) {
 		return []IOStream{}, err
 	}
 
-	for i := 0; i < len(pod.containers); i++ {
-		ioStream, err := p.allocateIOStream()
-		if err != nil {
-			return []IOStream{}, err
-		}
-
-		ioStreams = append(ioStreams, ioStream)
-	}
-
 	return ioStreams, nil
 }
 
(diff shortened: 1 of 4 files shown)


cc_proxy.go
```diff
@@ -98,15 +98,6 @@ func (p *ccProxy) register(pod Pod) ([]IOStream, error) {
...
 
-	for i := 0; i < len(pod.containers); i++ {
-		ioStream, err := p.allocateIOStream()
-		if err != nil {
-			return []IOStream{}, err
-		}
-
-		ioStreams = append(ioStreams, ioStream)
-	}
-
```
sameo: Not sure I understand: Why are we no longer returning the list of IO streams (eventually the list of tokens) here ?

sboeuf: Well the reason is that we are no longer calling register from StartPod, but from StartAgent instead. Several calls to connect will provide the tokens needed. The root reason is that we cannot disconnect completely the proxy from the VM and then reconnect (in case we are doing a stop pod then start pod) because the proxy will expect a READY message from hyperstart but because hyperstart is already running at that time and because we already consumed the READY from the previous StartPod, the proxy will hang. That's why the register has to be called only once, and in that case, we have to handle the ioStreams with the connect (one for each container) instead of the register.

sameo: > Well the reason is that we are no longer calling register from StartPod, but from StartAgent instead. Several calls to connect will provide the tokens needed

Right, but I don't see those tokens/IOstream returned anywhere. They will be used as shim parameters, so they need to be passed down to the caller. Am I missing something again ?

sameo: > Well the reason is that we are no longer calling register from StartPod, but from StartAgent instead. Several calls to connect will provide the tokens needed.

So we will call `proxy.register` and discard all tokens it created. It think this is a problem because:
1. Those are allocated resources by the proxy that we won't use.
2. The proxy will map a token to a container ID and will expect a shim connection with that token for the given token ID. Running subsequent `connect` for basically the same container ID seems wrong.

I think we should put that PR on hold for now, we're trying to address a use case that's non existent.

sameo:
I put this PR on `do-not-merge`. Please do not delete it as we might have to restore it later on.

sboeuf:
Let's close this one and maybe try to fix it later. The code in this PR is too old and needs to be completely updated.

PR #112, opened 2017-02-04
Adapt virtcontainers to fit different runtime expectations

sboeuf:
Some adjustments related to what can expect different runtimes, and particularly OCI runtime, needs to be made to get everything working as expected.

### DIFF
--- a/pod.go
+++ b/pod.go
@@ -215,6 +215,16 @@ type Cmd struct {
 
 	User  string
 	Group string
+
+	// Hidden field filled by agent implementation and used later by
+	// virtcontainers consumer who wants to start a shim to connect
+	// the proxy.
+	token string
+}
+
+// GetToken returns a shim token associated to the process defined by Cmd structure.
+func (c *Cmd) GetToken() string {
+	return c.token
 }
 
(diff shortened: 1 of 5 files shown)


pod.go
```diff
@@ -235,6 +235,13 @@ type PodStatus struct {
 	ContainersStatus []ContainerStatus
 }
 
+// ShimInfo gathers information needed by a shim, returned from a proxy.
+type ShimInfo struct {
+	Token string
+	IP    string
+	Port  string
+}
```
sameo: One Pod could have many shims associated to it. `IP` and `Port` make sense, but I don't think `Token` does.

sboeuf: Yes you're right. But in case I use a token list, how can I know (from runtime implementation) how to differentiate the tokens ?

sameo: You don't need to store the shim tokens at the PodConfig level. It's going to be up to the proxy to store those and also their mappings with e.g. the hyperstart seqnums.

sameo:
LGTM

sameo:
@sboeuf Could you please rebase and push ?

PR #114, opened 2017-02-07
pod: Keep containers instead of containers configs in Pod structure

sboeuf:
Until now, the Pod structure was keeping containers configurations but
it is not the right thing to do because we want to be able to access
containers through a handler kept by the Pod structure. That's why
this patch changes the containers configuration list to containers
handler list.


sameo:
LGTM
@sboeuf please rebase and push.

PR #115, opened 2017-02-07
oci: Container ID should be considered as pod ID

sboeuf:
Because OCI specification does not have the concept of pod, we have
to match the container ID on the pod ID, and assign an arbitrary
value to the virtcontainers internal container ID. In that case,
we chose "0" but it could be anything else.

### DIFF
--- a/pkg/oci/utils.go
+++ b/pkg/oci/utils.go
@@ -150,7 +150,6 @@ func PodConfig(runtime RuntimeConfig, bundlePath, cid, console string) (*vc.PodC
 	}
 
 	containerConfig := vc.ContainerConfig{
-		ID:          cid,
 		RootFs:      rootfs,
 		Interactive: ocispec.Process.Terminal,
 		Console:     console,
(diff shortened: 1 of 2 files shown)


sboeuf:
@sameo let me know what you think about this. I think the container ID given by OCI should be the pod ID because the container for OCI (with namespaces isolation) is similar to what we consider as a pod.

sameo:
@sboeuf I'm not opposed to that change, but if possible I would prefer not to make OCI an exception. Is the current OCI package implementation preventing us from implementing an OCI runtime ?

sameo:
Oh and if we really need this change, I'd rather not set `container.ID` at all and let container.go generate an ID for us.

sboeuf:
@sameo I think the current implementation will prevent us from starting the Pod for example. In case the start from OCI will provide container-ID, we have no clue to determine which pod-ID (assigned randomly) holds the right container.

sboeuf:
And I agree we can let virtcontainers assign the container ID instead of hardcoding this value.

sboeuf:
@sameo let me know what you think about this new change.

sameo:
LGTM

PR #116, opened 2017-02-08
Add KillContainer API

sboeuf:
Because OCI spec needs to be able to send a signal with the kill command and because it is something generic (not only OCI will need to send signals to containers), this PR adds the way to send a signal to a running container.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -18,6 +18,7 @@ package virtcontainers
 
 import (
 	"fmt"
+	"syscall"
 
 	"github.com/mitchellh/mapstructure"
 )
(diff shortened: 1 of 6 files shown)


sameo:
@sboeuf A few comments on this PR:

- I agree the `StopContainer()` implementation is incomplete, but it should be a combination of `kill TERM` and `removeContainer()`, not only `kill`:
   1) Send `kill TERM` (and not `KILL`) to the running container
   2) Remove the container

- `SignalProcess()` should be `KillContainer()` and should not take a process ID. There's only one running process per container (Because there's one Cmd associated to it) and the virtcontainers API should be kept at the Pod and Container level, not down to the PID.

sameo:
LGTM

PR #117, opened 2017-02-08
pod: Change paused state to stopped state

sboeuf:
Because we want to make the difference with a pod just created and a pod
already started and then stopped, we remove the paused state not used, and
we replace it with a stopped state.
This means that we can have now a pod and its containers in the following
states:

On Create():
- "" -> created

On Delete():
- created -> ""
- stopped -> ""

On Start():
- created -> running
- stopped -> running

On Stop():
- running -> stopped

### DIFF
--- a/container.go
+++ b/container.go
@@ -175,8 +175,8 @@ func (c *Container) delete() error {
 		return err
 	}
 
-	if state.State != stateReady {
-		return fmt.Errorf("Container not ready, impossible to delete")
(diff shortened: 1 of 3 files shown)


sboeuf:
@sameo this is the PR we were talking about to introduce stopped state in case we want to differentiate a start and a restart.

sameo:
LGTM

PR #118, opened 2017-02-08
Introduce Process structure to allow runtime to get useful info about the container's process

sboeuf:
This PR adds the Process structure so that any runtime implementation can update PID related to the started process on the container. Also it can retrieve the token necessary for the shim to connect the proxy.


sameo:
LGTM

PR #122, opened 2017-02-10
pkg: oci: Translate virtcontainers status to OCI state

sboeuf:
Through this patch, the OCI package provides a convenient way to convert
internal virtcontainers status to the expected OCI state defined in the
OCI specification.


sameo:
LGTM

PR #123, opened 2017-02-11
pod: Fix pod state not saved

sboeuf:
When a pod object was created, we were never assigning its state in case
this state was already existing.


sameo:
LGTM

PR #124, opened 2017-02-13
OCI package: Return container state instead of pod state

sboeuf:
Because the container state can be updated from the runtime in case we
check if the process ID is running or not, we want to retrieve the state
of the container instead the state of the pod, in order to be accurate.


sameo:
LGTM
@sboeuf Please rebase.

PR #130, opened 2017-02-24
hyperstart/mock: Don't directly print to stderr

dlespiau:
Instead of manually printing to stderr, use the built-in testing.Log(f)
functions to output debug messages when the test running is set up to be
verbose.


sameo:
LGTM

PR #131, opened 2017-02-27
pod: Don't hard-code details in returned Errors.

jodh-intel:
Volumes.Set() and Sockets.Set() now construct the Error message rather
than hard-coding separator and expected field counts.

### DIFF
--- a/pod.go
+++ b/pod.go
@@ -113,12 +113,15 @@ type Volumes []Volume
 // Set assigns volume values from string to a Volume.
 func (v *Volumes) Set(volStr string) error {
 	volSlice := strings.Split(volStr, " ")
+	const expectedVolLen = 2
+	const volDelimiter = ":"
 
 	for _, vol := range volSlice {
-		volArgs := strings.Split(vol, ":")
+		volArgs := strings.Split(vol, volDelimiter)
 
-		if len(volArgs) != 2 {
-			return fmt.Errorf("Wrong string format: %s, expecting only 2 parameters separated with ':'", vol)
+		if len(volArgs) != expectedVolLen {
+			return fmt.Errorf("Wrong string format: %s, expecting only %v parameters separated with %q",
+				vol, expectedVolLen, volDelimiter)
 		}
 
 		if volArgs[0] == "" || volArgs[1] == "" {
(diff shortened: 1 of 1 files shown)


pod.go
```diff
@@ -113,12 +113,15 @@ type Volumes []Volume
...
+	const volDelimiter = ":"
 
 	for _, vol := range volSlice {
-		volArgs := strings.Split(vol, ":")
+		volArgs := strings.Split(vol, volDelimiter)
 
-		if len(volArgs) != 2 {
-			return fmt.Errorf("Wrong string format: %s, expecting only 2 parameters separated with ':'", vol)
+		if len(volArgs) != expectedVolLen {
+			return fmt.Errorf("Wrong string format: %s, expecting only %v parameters separated with %q",
```
sameo: `expectedVolLen` is a int, the printf format should be `%d`.

pod.go
```diff
@@ -162,12 +165,14 @@ type Sockets []Socket
 // Set assigns socket values from string to a Socket.
 func (s *Sockets) Set(sockStr string) error {
 	sockSlice := strings.Split(sockStr, " ")
+	const expectedSockCount = 4
+	const sockDelimiter = ":"
 
 	for _, sock := range sockSlice {
-		sockArgs := strings.Split(sock, ":")
+		sockArgs := strings.Split(sock, sockDelimiter)
```
dlespiau: For the record, I think `strings.Split(sock, ":")` is more readable than `strings.Split(sock, sockDelimiter)` and then go look what `sockDelimiter` is.

pod.go
jodh-intel: yep - I'm having issues running the tests which hopefully would have caught this before pushing ;)

jodh-intel: I understand the sentiment, but you do only need to look up 3 lines to see the definition ;) But if we don't use a variable, that error will bitrot - duplication is evil.

pod.go
sameo: @dlespiau 
@jodh-intel commented:

> I understand the sentiment, but you do only need to look up 3 lines to see the definition ;) But if we don't use a variable, that error will bitrot - duplication is evil.

sameo:
LGTM

sameo:
@jodh-intel Please rebase.

jodh-intel:
done.

PR #132, opened 2017-02-27
Fix dir+file perms

jodh-intel:


pkg/cni/cni_test.go
```diff
@@ -27,6 +27,10 @@ import (
 	"github.com/containernetworking/cni/pkg/ns"
 )
 
+const (
+	dirMode = os.FileMode(0750)
```
sameo: `os.FileMode(0750)` is `-rwxr-x---`. I think we want `drwxr-x---`, and thus that should be:

```Go
      dirMode = os.FileMode(0750) | os.ModeDir

Comment (jodh-intel) [inline pkg/cni/cni_test.go]: I think `os.ModeDir` is only used for checking perms - it's not actually required by `Mkdir*` (probably because the `os.ModeDir` is implicit in the call).

Comment (sameo) [inline pkg/cni/cni_test.go]: Yes, that's right. Sorry for the noise.

File: pod_test.go [hunk 2]
```diff
@@ -675,7 +675,7 @@ func TestPodCheckContainerStateFailingNotExpectedState(t *testing.T) {
 	}
 
 	path := filepath.Join(runStoragePath, testPodID, contID)
-	err := os.MkdirAll(path, os.ModeDir)
+	err = os.MkdirAll(path, dirMode)
```
sameo: s/`=`/`:=`/

pod_test.go
```diff
@@ -734,7 +734,7 @@ func TestPodCheckContainersStateFailingEmptyPodID(t *testing.T) {
 func TestMain(m *testing.M) {
 	flag.Parse()
 
-	err := os.MkdirAll(testDir, os.ModeDir)
+	err = os.MkdirAll(testDir, dirMode)
```
sameo: ditto

pod_test.go
jodh-intel: Fixed. Not sure what happened there. I blame YCM ;)

sameo:
LGTM

PR #133, opened 2017-03-01
Non root tests

jodh-intel:


jodh-intel:
Travis seems to be borked atm.

virtcontainers_test.go
```diff
@@ -102,6 +103,7 @@ func TestMain(m *testing.M) {
 	podDirLock = filepath.Join(runStoragePath, testPodID)
 	podFileState = filepath.Join(runStoragePath, testPodID, stateFile)
 	podFileLock = filepath.Join(runStoragePath, testPodID, lockFileName)
+	testBinHookPath = filepath.Join(testDir, "bin", "hook")
```
sboeuf: Unfortunately, this won't work. Look at travis.yml, we are building and copying the hook binary at that moment, meaning that /tmp/bin/hook is a hardcoded value. The dynamic directory created by ioutil.TempDir() won't fit.

virtcontainers_test.go
```diff
@@ -31,6 +31,8 @@ const testImage = "image"
 const testHypervisor = "hypervisor"
 const testBundle = "bundle"
 
+const testDisabledAsNonRoot = "Test disabled as requires root privs"
```
sboeuf: "privileges" instead of "privs" would be better, wouldn't it ?

virtcontainers_test.go
jodh-intel: yep - I just discovered the interesting travis config when I raised the PR.

virtcontainers_test.go
jodh-intel: Agreed.

pkg/cni/cni_test.go
```diff
@@ -391,9 +391,21 @@ func TestMain(m *testing.M) {
 		os.Exit(1)
 	}
 
+	testNetNsPath, err = ioutil.TempDir("", "virtcontainers-test-ns-")
```
sboeuf: I think this won't work. The testNetNsPath will be overriden by testNetNsPath assignments in other test functions, won't it ?

jodh-intel:
Aargh - it looks like my adding a `t.Skip()` to `TestStartStopContainerHyperstartAgentSuccessful()` has pushed it over the cyclomatic complexity threshold. Any ideas (aside from bumping the threshold by 1 (that test has value `16` and the threshold is `15`))?

sboeuf:
@jodh-intel The only clean solution is to factorize some code inside a new function...

jodh-intel:
@sameo, @sboeuf - I think we can remove the `do-not-merge` label now - I don't have privs to do that (sniff :cry: )

sboeuf:
@sameo if you are okay with this PR, I can merge it.

sameo:
LGTM

PR #134, opened 2017-03-01
pod: Remove redundant call to newAgent().

jodh-intel:
PodConfig.valid() was calling newAgent() but ignoring the return value.
Since the return value isn't needed, remove the call.


sboeuf:
lgtm

PR #136, opened 2017-03-02
filesystem: Remove unecessary calls to os.Stat().

jodh-intel:
The call that follows will detect if the file does/does not
exist.

### DIFF
--- a/filesystem.go
+++ b/filesystem.go
@@ -146,11 +146,6 @@ func (fs *filesystem) createAllResources(pod Pod) (err error) {
 }
 
 func (fs *filesystem) storeFile(path string, data interface{}) error {
-	_, err := os.Stat(path)
-	if err == nil {
-		os.Remove(path)
-	}
-
 	f, err := os.Create(path)
 	if err != nil {
 		return err
@@ -167,11 +162,6 @@ func (fs *filesystem) storeFile(path string, data interface{}) error {
 }
 
 func (fs *filesystem) fetchFile(path string, data interface{}) error {
-	_, err := os.Stat(path)
-	if err != nil {
-		return err
-	}
-
 	fileData, err := ioutil.ReadFile(path)
 	if err != nil {
 		return err


filesystem.go
```diff
@@ -152,11 +152,6 @@ func (fs *filesystem) createAllResources(pod Pod) error {
 }
 
 func (fs *filesystem) storeFile(path string, data interface{}) error {
-	_, err := os.Stat(path)
```
jodh-intel: BTW, @sameo - I don't really understand that `os.Stat(podlockFile)` call above - again, it seems like it *could* be dropped, or could it...? It looks racy though.

sboeuf: @jodh-intel We need this function call to make sure we remove the existing file before we create it again. That's the way to override data stored in the filesystem.

filesystem.go
```diff
@@ -173,11 +168,6 @@ func (fs *filesystem) storeFile(path string, data interface{}) error {
 }
 
 func (fs *filesystem) fetchFile(path string, data interface{}) error {
-	_, err := os.Stat(path)
```
sboeuf: Why do you want to remove this one ? You have to return an error in case you cannot find the file which is supposed to be stored, don't you ?

jodh-intel: Yes, and `ioutil.Readfile()` will return an error if `path` doesn't exist, and crucially do it atomically rather than having to have 2 syscalls.

sboeuf: I didn't know that, then sounds good :)

filesystem.go
sboeuf: What do you mean, you try to store everything to a temporary file and if everything succeed, in the end of the function you move the content of this temp file to the one we are targeting ?

jodh-intel: yes, that would be safest approach, as the move is atomic so you can't end up with a partially written file in the final destination.

sboeuf: Okay then I suggest that you modify your PR to do that, right ?

jodh-intel: Hi @sboeuf - I've looked at this again and I don't think the `podlockFile` `Stat()` call can be removed. Also, since it's an empty file it's going to be created atomically anyhow.

Summary: I don't think there are any further changes required for this PR :-)

sameo:
LGTM

PR #139, opened 2017-03-02
Anticipate changes introduced by new proxy version

sboeuf:
The purpose of this PR is to bring the changes we will need to integrate very quickly the new proxy version. Once the proxy will be released, only very minor changes will be needed to get virtcontainers working with this new version.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -132,6 +132,9 @@ type agent interface {
 	// stopPod will tell the agent to stop all containers related to the Pod.
 	stopPod(pod Pod) error
 
+	// createContainer will tell the agent to create a container related to a Pod.
+	createContainer(contConfig ContainerConfig) error
+
 	// startContainer will tell the agent to start a container related to a Pod.
 	startContainer(pod Pod, contConfig ContainerConfig) error
 
--- a/agent.go
+++ b/agent.go
@@ -117,30 +117,30 @@ type agent interface {
 	// to handle all other Agent interface methods.
 	init(pod *Pod, config interface{}) error
 
-	// startAgent will start the agent.
-	startAgent() error
+	// start will start the agent.
+	start(pod *Pod) error
 
-	// stopAgent will stop the agent.
-	stopAgent() error
+	// stop will stop the agent.
+	stop(pod Pod) error
 
 	// exec will tell the agent to run a command in an already running container.
-	exec(pod Pod, container Container, cmd Cmd) (*Process, error)
+	exec(pod Pod, c Container, cmd Cmd) (*Process, error)
 
 	// startPod will tell the agent to start all containers related to the Pod.
-	startPod(config PodConfig) error
+	startPod(pod Pod) error
 
 	// stopPod will tell the agent to stop all containers related to the Pod.
 	stopPod(pod Pod) error
 
 	// createContainer will tell the agent to create a container related to a Pod.
-	createContainer(contConfig ContainerConfig) error
+	createContainer(pod Pod, c *Container) error
 
 	// startContainer will tell the agent to start a container related to a Pod.
-	startContainer(pod Pod, contConfig ContainerConfig) error
+	startContainer(pod Pod, c Container) error
 
 	// stopContainer will tell the agent to stop a container related to a Pod.
-	stopContainer(pod Pod, container Container) error
+	stopContainer(pod Pod, c Container) error
 
 	// killContainer will tell the agent to send a signal to a container related to a Pod.
-	killContainer(pod Pod, container Container, signal syscall.Signal) error
+	killContainer(pod Pod, c Container, signal syscall.Signal) error
 }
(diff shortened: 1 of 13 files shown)


cc_proxy.go
```diff
@@ -49,65 +49,63 @@ func (p *ccProxy) connectProxy(runtimeSocketPath string) (*api.Client, error) {
...
+		return []ProxyInfo{}, fmt.Errorf("Wrong agent config type, should be HyperConfig type")
 	}
 
 	_, err = p.client.Hello(pod.id, hyperConfig.SockCtlName, hyperConfig.SockTtyName, nil)
 	if err != nil {
-		return []IOStream{}, err
+		return []ProxyInfo{}, err
+	}
+
+	if newInfos == false {
```
jodh-intel: This name is in keeping with `ProxyInfo`, but I found it rather unclear. If I'm understanding it, a comment to state that this parameter controls whether a new set of proxy I/O streams is allocated, or whether the existing ones are used would have made it much clearer I think (or maybe renaming it `allocateNewStreams` or similar).

sboeuf: Well we are not allocating new streams with the new proxy, we are only getting tokens, and I didn't want to use a too specific name like "token". That's why I used info.

sboeuf: maybe newProxyInfo ?

jodh-intel: Yeah, or maybe `newConnectionInfo` (well, that's a bit long, but... ;-)

Maybe you could invert the logic and have:

```
reuseConnection bool
```

... as that's clearer still I think.

sboeuf: yes reuseConnection sounds good :)

jodh-intel:
lgtm

container.go
```diff
@@ -233,20 +233,27 @@ func createContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 		return nil, err
 	}
 
+	process, err := c.pod.storage.fetchContainerProcess(c.podID, c.id)
```
sameo: `c.fetchProcess()`

cc_proxy.go
```diff
@@ -49,65 +49,63 @@ func (p *ccProxy) connectProxy(runtimeSocketPath string) (*api.Client, error) {
...
+		return []ProxyInfo{}, fmt.Errorf("Wrong agent config type, should be HyperConfig type")
 	}
 
 	_, err = p.client.Hello(pod.id, hyperConfig.SockCtlName, hyperConfig.SockTtyName, nil)
 	if err != nil {
-		return []IOStream{}, err
+		return []ProxyInfo{}, err
+	}
+
+	if reuseConnection == true {
```
sboeuf: Agreed

agent.go
```diff
@@ -118,29 +118,29 @@ type agent interface {
 	init(pod *Pod, config interface{}) error
 
 	// startAgent will start the agent.
-	startAgent() error
+	startAgent(pod *Pod) error
```
sboeuf: Agreed, I'll do the same for stopAgent() ==> stop()

cc_proxy.go
```diff
@@ -49,65 +49,67 @@ func (p *ccProxy) connectProxy(runtimeSocketPath string) (*api.Client, error) {
...
 	_, err = p.client.Hello(pod.id, hyperConfig.SockCtlName, hyperConfig.SockTtyName, nil)
 	if err != nil {
-		return []IOStream{}, err
+		return []ProxyInfo{}, err
+	}
+
+	// reuseConnection is intended to be true in case we don't want
+	// the proxy to create new tokens for every container described
+	// in the pod, but instead only connect to the agent inside the
+	// VM.
```
sameo: That should only happen with `connect`, I think we can remove that one.

proxy.go
```diff
@@ -89,30 +88,38 @@ func newProxyConfig(config PodConfig) interface{} {
...
 	// register connects and registers the proxy to the given VM.
-	// It also returns streams related to containers workloads.
-	register(pod Pod) ([]IOStream, error)
+	// It also returns information related to containers workloads.
+	//
+	// reuseConnection is intended to be true in case we don't want
+	// the proxy to create new tokens for every container described
+	// in the pod, but instead only connect to the agent inside the
+	// VM.
+	register(pod Pod, reuseConnection bool) ([]ProxyInfo, error)
```
sameo: You can remove that one.

proxy.go
```diff
@@ -89,30 +88,38 @@ func newProxyConfig(config PodConfig) interface{} {
...
 
 	// connect gets the proxy a handle to a previously registered VM.
-	// It also returns streams related to containers workloads.
-	connect(pod Pod) (IOStream, error)
+	// It also returns information related to containers workloads.
+	//
+	// reuseConnection is intended to be true in case we don't want
+	// the proxy to create a new token, but instead only get a handle
+	// to be able to communicate with the agent inside the VM.
+	connect(pod Pod, reuseConnection bool) (ProxyInfo, error)
```
sameo: I personally find that passing a `createToken bool` would be more logical.

sameo:
LGTM

PR #140, opened 2017-03-02
agent: Cleanup the interface by removing unused parameters

sboeuf:
A lot of functions exposed by the interface were expecting a pod as
parameter but it could generate some conflicts with the pod pointer
internally stored by each agent implementation.
Moreover, because we are always expecting the agent init() function
to be called every time we need to recreate a pod handler, this means
we expect the agent to store this handler when passed through init().
That way, we don't have to give this parameter again for each function.


sboeuf:
Covered by #139 instead

PR #141, opened 2017-03-03
hypervisor: Add Debug option to HypervisorConfig.

jodh-intel:
Setting HypervisorConfig.Debug will change the default set of kernel
parameters to enable debug (boot) output. It could also be used to
enable hypervisor debug where available.


sameo:
LGTM

PR #143, opened 2017-03-06
container: Add helper function for boilerplate code.

jodh-intel:
Refactored a common preamble used by start(), stop(),
enter() and kill() into a helper function.

### DIFF
--- a/container.go
+++ b/container.go
@@ -262,17 +262,31 @@ func (c *Container) delete() error {
 	return nil
 }
 
-func (c *Container) start() error {
+// fetchState retrieves the container state.
+//
+// cmd specifies the operation (or verb) that the retieval is destined
+// for and is only used to make the returned error as descriptive as
+// possible.
+func (c *Container) fetchState(cmd string) (State, error) {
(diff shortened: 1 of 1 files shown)


container.go
```diff
@@ -262,18 +262,32 @@ func (c *Container) delete() error {
 	return nil
 }
 
-func (c *Container) start() error {
+// fetchRunningState retrieves the container state.
+//
+// cmd specifies the operation (or verb) that the retieval is destined
+// for and is only used to make the returned error as descriptive as
+// possible.
+func (c *Container) fetchRunningState(cmd string) (State, error) {
```
sboeuf: I think you could simplify this function name (e.g fetchState), or use fetchStateRunningPod(), because I think the name can be a bit confusing as we can think that we are checking the container's state being "running". What do you think ?

sboeuf:
@jodh-intel This change makes sense, I have only one comment related to the function name.

container.go
jodh-intel: Agreed - updated.

sameo:
LGTM

PR #144, opened 2017-03-07
oci: Add a EnvVar slice conversion API

sameo:
This converts an OCI Process.Env slice into a vc.EnvVar one.

This is needed by https://github.com/clearcontainers/runtime/pull/53

### DIFF
--- a/pkg/oci/utils.go
+++ b/pkg/oci/utils.go
@@ -218,3 +218,40 @@ func stateToOCIState(state vc.State) string {
 		return ""
 	}
 }
+
+// EnvVars converts an OCI process environment variables slice
+// into a virtcontainers EnvVar slice.
+func EnvVars(envs []string) ([]vc.EnvVar, error) {
+	var envVars []vc.EnvVar
+
+	envDelimiter := "="
+	expectedEnvLen := 2
+
+	for _, env := range envs {
+		envSlice := strings.SplitN(env, envDelimiter, expectedEnvLen)
+
+		if len(envSlice) < expectedEnvLen {
+			return []vc.EnvVar{}, fmt.Errorf("Wrong string format: %s, expecting only %v parameters separated with %q",
+				env, expectedEnvLen, envDelimiter)
+		}
+
+		if envSlice[0] == "" {
+			return []vc.EnvVar{}, fmt.Errorf("Environment variable cannot be empty")
+		}
+
+		envSlice[1] = strings.Trim(envSlice[1], "' ")
+
+		if envSlice[1] == "" {
+			return []vc.EnvVar{}, fmt.Errorf("Environment value cannot be empty")
+		}
+
+		envVar := vc.EnvVar{
+			Var:   envSlice[0],
+			Value: envSlice[1],
+		}
+
+		envVars = append(envVars, envVar)
+	}
+
+	return envVars, nil
+}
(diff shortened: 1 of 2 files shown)


pkg/oci/utils_test.go
```diff
@@ -293,6 +293,50 @@ func TestStateToOCIState(t *testing.T) {
...
+		t.Fatalf("Got %v\n expecting %v", vcEnvVars, expectecVcEnvVars)
+	}
+}
+
+func TestMalformedEnvVars(t *testing.T) {
+	envVars := []string{"foo"}
+
+	_, err := EnvVars(envVars)
+	if err == nil {
+		t.Fatalf("Could not create environment variable slice %v", err)
```
jodh-intel: I think this message (and the one below) need to say instead something like "EnvVars() succeeded unexpectedly: ...".
Also, you are displaying `err` which is only ever going to be `nil` in that block.

pkg/oci/utils.go
```diff
@@ -218,3 +218,38 @@ func stateToOCIState(state vc.State) string {
...
+			return []vc.EnvVar{}, fmt.Errorf("Wrong string format: %s, expecting only %v parameters separated with %q",
+				env, expectedEnvLen, envDelimiter)
+		}
+
+		if envSlice[0] == "" {
+			return []vc.EnvVar{}, fmt.Errorf("Environment variable cannot be empty")
+		}
+
+		if envSlice[1] == "" {
+			return []vc.EnvVar{}, fmt.Errorf("Environment value cannot be empty")
```
jodh-intel: Surely you should be able to define `name=""`?

pkg/oci/utils.go
```diff
@@ -218,3 +218,38 @@ func stateToOCIState(state vc.State) string {
...
+		}
+
+		envVar := vc.EnvVar{
```
jodh-intel: I wonder whether any further validation should be performed here (begins with underscore or alpha-numeric, etc)?

pkg/oci/utils.go
```diff
@@ -218,3 +218,38 @@ func stateToOCIState(state vc.State) string {
...
+// EnvVars converts an OCI process environment variables slice
+// into a virtcontainers EnvVar slice.
+func EnvVars(envs []string) ([]vc.EnvVar, error) {
+	var envVars []vc.EnvVar
+
+	envDelimiter := "="
+	expectedEnvLen := 2
+
+	for _, env := range envs {
+		envSlice := strings.Split(env, envDelimiter)
```
jodh-intel: This needs to be something like:

```
strings.Split(env, envDelimiter, expectedEnvLen)

if len(envSlice) < expectedEnvLen {
    // ...
}
```

Otherwise, values like `foo=bar=` will be rejected incorrectly.

pkg/oci/utils_test.go
```diff
@@ -293,6 +293,50 @@ func TestStateToOCIState(t *testing.T) {
...
+	}
+
+	envVars = []string{"TERM="}
```
jodh-intel: It would be good to also test the following explicitly:

- `name='='` (invalid)
- `value="="` (valid)
- `value=' '` (space/tab test)

Then, various other weird values like:

- `===`
- `=foo`
- `bar=`
- `=baz=`
- `"I am a string with spaces and escaped characters like \""`

pkg/oci/utils.go
sboeuf: Yes maybe, but you need to identify what you don't expect to be in a parameter or in its value.

sboeuf:
LGTM

jodh-intel:
lgtm

PR #145, opened 2017-03-07
container: Add public function to retrieve container process

sboeuf:
Programs using virtcontainers library will sometimes need to get
the process related to a container in case they have retrieved this
container. Process being a private field in the container structure,
we have to create a new public function to return it.


sameo:
LGTM

jodh-intel:
lgtm

PR #146, opened 2017-03-07
Deal with URLs related to proxy

sboeuf:
Every time a runtime wants to start a shim, it needs a URL to know where to connect to the proxy. Through this PR, virtcontainers provide the way to get this URL from the Pod handler.
Also, at the time a runtime loads the proxy configuration, it will provide virtcontainers a URL that has to be parsed properly to know how to connect to the proxy. This PR also handles this.

### DIFF
--- a/container.go
+++ b/container.go
@@ -251,7 +251,7 @@ func createContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 	// specific case.
 	pod.containers = append(pod.containers, c)
 
-	if err := c.pod.agent.createContainer(*pod, c); err != nil {
+	if err := c.pod.agent.createContainer(pod, c); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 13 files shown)


container.go
```diff
@@ -30,6 +30,7 @@ import (
 type Process struct {
 	Token string
 	Pid   int
+	URL   string
```
sameo: I do not believe this belong to the process structure.
I think the URL that we will pass to the shim is a `Pod` URL, although technically it is a proxy one for now. But the shim should not be aware of which entity it's connecting to.

So what we want is to modify the `Pod` structure to replace the `controlSocket` (which we do not use) with a private URL string. We want a getter for that one.
The proxy will give us that URL when registering the VM and potentially when attaching to a running VM and this is when we will fill that `Pod` specific field. This also means that `EnterContainer()` will return a `Pod` pointer as well.

sameo:
LGTM

PR #148, opened 2017-03-09
utils: Add a script setting up the environment for testing purpose

sboeuf:
This new script helps any user or developer to get the proper
configuration in order to use and test virtcontainers library.
After this script is run, you should be able to use virtc CLI
and to run "go test" properly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ go:
 env:
 - COVERALLS_TOKEN=3NUr5Z9zvjIoGRlKi6WIA3Q9em3x22rZp
 
+services:
+- docker
+
 before_install:
 - go get github.com/mattn/goveralls
 - go get golang.org/x/tools/cmd/cover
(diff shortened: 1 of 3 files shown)


jodh-intel:
Hi @sboeuf - thanks for this: it's going to be very helpful!

However, to avoid duplication, could you make `.travis.yml` call this script?

jodh-intel:
It would be good to reference `utils/setup_virtcontainers_env.sh` in `README.md` too.

sboeuf:
@jodh-intel yes this has to be referenced in the readme update.
About travis calling into this script, it is a good idea, I will try it.

sboeuf:
@jodh-intel It is not really working. go is not known as a binary when executed from the script. And docker command is not known, but do we want to install docker ?

sameo:
@sboeuf Yes, we can install docker from travis. Also we can install cni from there as well, leaving this script with no source installation to do.
Then I think this script could be split into 2 files: One under `.ci/ci-setup.sh` that would be directly used by travis and a second one that could be used for local setup where we'd install virtcontainers and cni, and then call `.ci/ci-setup.sh`.

sboeuf:
@sameo, @jodh-intel I figured out my issue with sudo, and I have now travis able to call into my script. I am wondering about the ci-setup.sh that you mentioned @sameo , what should be in there ? It is not clear to me what should be done in this ci-setup.sh

jodh-intel:
Hi @sboeuf - I'm not really clear on your question about `ci-setup.sh`? For `cc-oci-runtime`, that script is designed to setup the environment before tests are run. It should be CI-system-agnostic (or include code to detect the particular CI system it's running under). But also if possible that script should be usable by users so should not trash the host system if at all possible :smile:

sboeuf:
@jodh-intel I am more confused now, because what you are explaining (which is obviously true in case of cc-oci-runtime) is different from what @sameo said.

> @sboeuf Yes, we can install docker from travis. Also we can install cni from there as well, leaving this script with no source installation to do.

@sameo, @jodh-intel, I am just trying to make sure we agree about what has to be into th(is/ese) script(s) :-)

As is, ci-setup.sh works so that we can use it directly from travis or from our local machine.

sameo:
> As is, ci-setup.sh works so that we can use it directly from travis or from our local machine.

I'm just fine with that, thanks.

LGTM

PR #152, opened 2017-03-10
Add pod annotations and update oci package to use them

sboeuf:
This PR provides a way for a virtcontainers' user to store/fetch private data through virtcontainers management.

### DIFF
--- a/pod.go
+++ b/pod.go
@@ -269,6 +269,10 @@ type PodConfig struct {
 	// This list can be empty and populated by adding containers
 	// to the Pod a posteriori.
 	Containers []ContainerConfig
+
+	// Annotations keys must be unique strings an must be name-spaced
+	// with e.g. reverse domain notation (org.clearlinux.key).
+	Annotations map[string]string
 }
 
 // valid checks that the pod configuration is valid.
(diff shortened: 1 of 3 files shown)


pod.go
```diff
@@ -269,6 +269,12 @@ type PodConfig struct {
 	// This list can be empty and populated by adding containers
 	// to the Pod a posteriori.
 	Containers []ContainerConfig
+
+	// Annotations is a reserved field and can be filled with various
```
sameo: `reserved` fields usually refer to unused and undefined block of a frame or structure. This one is well defined and refers to e.g. OCI annotations. So we should not use that word here.

pod.go
```diff
@@ -269,6 +269,12 @@ type PodConfig struct {
...
+	// different data by each user. This means that virtcontainers will
+	// never try to interpret those data, it is only providing a way to
+	// save some data for users.
```
sameo: We need to be more precise on how `Annotations` should be used as there's nothing preventing different users to add their own annotations to the same pod config.
So we need to add something like:
```
Annotations keys must be unique strings an must be name-spaced with e.g. reverse domain notation (org.clearlinux.key).
```

pod.go
```diff
@@ -348,6 +354,20 @@ func (p *Pod) ID() string {
 	return p.id
 }
 
+// Annotations returns any annotation that a user could have stored through the pod.
+func (p *Pod) Annotations(key string) (string, error) {
+	if len(p.config.Annotations) == 0 {
+		return "", fmt.Errorf("Annotations map is empty")
+	}
+
```
sameo: I don't think this check is necessary. If the map len is empty, the following check (`if exist == false`) will fail.

sboeuf:
@sameo I addressed your comments, please take a look at this new version.

sameo:
LGTM

PR #153, opened 2017-03-10
pkg: oci: Return OCI spec structure from PodConfig()

sboeuf:
OCI spec structure can be needed by the runtime in order to reuse
some data not stored inside PodConfig, but without unmarshalling
the JSON config file again.


sameo:
LGTM

PR #156, opened 2017-03-14
pod: Enable pod console traces

sboeuf:
We had no way to get logs printed inside the VM by our hyperstart
agent because we were missing several parts to be connected properly.

This patch enables the proxy to retrieve hyperstart logs coming from
inside the VM through a new console socket created by the hypervisor.

First it adds a new Console field to the PodConfig so as to be able
to share this socket path with other components such as proxy or
agent implementations.

Then we have to change systemd.unit parameter from "container.target"
to "cc-agent.target" because we need hyperstart output to be
redirected to the right console.

At last, the hypervisor implementation has to create a console socket
from the console path provided by the PodConfig.

### DIFF
--- a/qemu.go
+++ b/qemu.go
@@ -65,6 +65,10 @@ const (
 	defaultMemSlots uint8 = 2
 )
 
+const (
+	defaultConsole = "console.sock"
+)
+
 type qmpGlogLogger struct{}
 
 func (l qmpGlogLogger) V(level int32) bool {
@@ -101,7 +105,7 @@ var kernelDefaultParams = []Param{
 	{"console", "hvc1"},
 	{"initcall_debug", ""},
 	{"init", "/usr/lib/systemd/systemd"},
-	{"systemd.unit", "container.target"},
+	{"systemd.unit", "cc-agent.target"},
 	{"iommu", "off"},
 	{"systemd.mask", "systemd-networkd.service"},
 	{"systemd.mask", "systemd-networkd.socket"},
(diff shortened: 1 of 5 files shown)


sboeuf:
@sameo @dlespiau please let me know what you think about this one ?

qemu.go
```diff
@@ -217,32 +217,46 @@ func (q *qemu) appendFSDevices(devices []ciaoQemu.Device, podConfig PodConfig) [
...
+		if c.Interactive == false || c.Console == "" {
 			console = ciaoQemu.CharDevice{
 				Driver:   ciaoQemu.Console,
 				Backend:  ciaoQemu.Socket,
-				DeviceID: fmt.Sprintf("console%d", i),
-				ID:       fmt.Sprintf("charconsole%d", i),
-				Path:     consolePath,
+				DeviceID: fmt.Sprintf("console%d", idx),
+				ID:       fmt.Sprintf("charconsole%d", idx),
+				Path:     fmt.Sprintf("%s/%s/%s/console.sock", runStoragePath, podConfig.ID, c.ID),
```
mcastelino: @sboeuf is this the same as https://github.com/01org/cc-oci-runtime/blob/045d668b47ba23c8d6a2951c2fd8bdc9b8410f2c/src/oci.h#L66 
We use this for debugging the VM bootup and kernel issues too.

sboeuf: Yes that's the same idea

mcastelino: @sboeuf In that case can we add a debug parameter to our runtime to allow a direct console interface into the guest OS using something along the lines of 
https://github.com/jcvenegas/cc-oci-runtime/commit/3334a2bc7c915e1f631bc85355f1ce20ea984200

This will provide a way to debug any launch or agent failure.

sboeuf: @mcastelino not sure I understand what you suggest here. We can add something to the runtime, but this is not part of virtcontainers. This patch is intended to enable something which was not working until now.

qemu.go
```diff
@@ -217,32 +217,46 @@ func (q *qemu) appendFSDevices(devices []ciaoQemu.Device, podConfig PodConfig) [
...
+func (q *qemu) appendConsoles(devices []ciaoQemu.Device, podConfig *PodConfig) []ciaoQemu.Device {
 	serial := ciaoQemu.SerialDevice{
 		Driver: ciaoQemu.VirtioSerial,
 		ID:     "serial0",
 	}
 
 	devices = append(devices, serial)
 
+	if podConfig.Console == "" {
+		podConfig.Console = fmt.Sprintf("%s/%s/console.sock", runStoragePath, podConfig.ID)
```
jodh-intel: There are quite a few references to `console.sock` now. I'd consider creating a `const` for it to avoid repetition.

sboeuf: Agreed.

sboeuf:
@sameo any comments on this one ?

sameo:
@sboeuf I think this is a very useful patch, but I don't like that we modify the `PodConfig` behind the caller's back. This should be an immutable structure. We already do that for the pod ID, but we should fix that as well.

So, what I'd like to see changed here is the fact that if `PodConfig.Console == ""` then we simply don't add a console device. Passing an empty string just means the caller does not want a console.
With that in mind you don't need to change several API prototypes and change the pod config anymore.

sboeuf:
@sameo Ok with me, I'll do the change.

sboeuf:
@sameo I have made the changes but something is bothering me. I have my VM crashing in case I don't provide the console path, meaning no console is created. It is related to the cc-agent.target used. In cases I use the combo cc-agent.target + console OR container.target + no console, there is no issue.
Should I try to switch between those two systemd.unit targets according to the PodConfig.Console value ? Or should I handle that differently ?

sameo:
@sboeuf The systemd service that we want to run should actually be a hypervisor parameter. Intrinsic knowledge that a hardcoded systemd service that we use expects a console is not something we want to put into virtcontainers. It really is the caller that knows about those dependencies.

sameo:
LGTM

sboeuf:
@sameo I guess this one deserves a follow up PR to make the systemd.unit value as a parameter, right ?

sameo:
@sboeuf Yes. Or more generally, allow for additional kernel parameters (for init, systemd.unit, etc...)

sboeuf:
@sameo Yes makes sense !

PR #157, opened 2017-03-15
hyperstart: Setup the network inside the VM

sboeuf:
Expected network has to be described to the agent running inside the
VM so as to set it up properly. Hyperstart agent implementation needs
to do that during the starting of the pod because network description
is passed through the STARTPOD command.

### DIFF
--- a/hyperstart.go
+++ b/hyperstart.go
@@ -122,6 +122,34 @@ type hyperstartProxyCmd struct {
 	token   string
 }
 
+// HyperPod is the structure corresponding to the format expected
+// by hyperstart to start the pod on the guest.
+type HyperPod struct {
+	Hostname   string                `json:"hostname"`
+	Containers []hyperJson.Container `json:"containers,omitempty"`
+	Interfaces []HyperNetIface       `json:"interfaces,omitempty"`
+	Routes     []hyperJson.Route     `json:"routes,omitempty"`
+	ShareDir   string                `json:"shareDir"`
+}
+
+// HyperNetIface is the structure corresponding to the format expected
+// by hyperstart to setup network interfaces on the guest.
+type HyperNetIface struct {
+	Device      string           `json:"device,omitempty"`
+	NewDevice   string           `json:"newDeviceName,omitempty"`
+	IPAddresses []HyperIPAddress `json:"ipAddresses"`
+	NetMask     string           `json:"netMask"`
+	MTU         string           `json:"mtu"`
+	MACAddr     string           `json:"macAddr"`
+}
+
+// HyperIPAddress is the structure corresponding to the format expected
+// by hyperstart to describe an IP address.
+type HyperIPAddress struct {
+	IPAddress string `json:"ipAddress"`
+	NetMask   string `json:"netMask"`
+}
+
 func (h *hyper) buildHyperContainerProcess(cmd Cmd, terminal bool) (*hyperJson.Process, error) {
 	var envVars []hyperJson.EnvironmentVar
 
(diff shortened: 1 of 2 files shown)


dlespiau:
Urgh! the PR I just merged was targeting a branch that was already a PR (this one). And now pullapprove fails on the merge commit...

Probably not a great idea to do that :) a way to fix it would be to rework that branch to include all commits.

sboeuf:
@dlespiau I have re-submitted this PR so that we have only one commit (no merge commit included). Once it will be merged, I will submit the CNM fix again.

cni.go
```diff
@@ -37,6 +37,14 @@ func (n *cni) addVirtInterfaces(networkNS *NetworkNamespace) error {
 			return err
 		}
 
+		// Interface index found by CNI plugin has to be updated by adding an offset
+		// of 1 because it starts to count interfaces indexes from 0 instead 1.
+		// Network interface standards expect interface indexes to be count from 1,
+		// zero being never used.
+		for i := range result.IPs {
+			result.IPs[i].Interface = result.IPs[i].Interface + 1
```
dlespiau: optional: += 1 works :)

network.go
```diff
@@ -390,6 +390,50 @@ func createNetworkEndpoints(numOfEndpoints int) (endpoints []Endpoint, err error
...
+		return nil
+	})
+	if err != nil {
+		return []net.Interface{}, err
+	}
+
+	return ifaces, err
+}
+
+func getIPNetFromEndpoint(endpoint Endpoint, netIfaces []net.Interface) (net.IPNet, error) {
```
dlespiau: Not familiar with that code at all, but I had a question: does this work when multiple IP addresses are bound to the same interface?

hyperstart.go
```diff
@@ -122,6 +122,25 @@ type hyperstartProxyCmd struct {
 	token   string
 }
 
+// HyperPod is the structure corresponding to the format expected
+// by hyperstart to start the pod on the guest.
+type HyperPod struct {
```
dlespiau: Seems to be missing the routes.

hyperstart.go
```diff
@@ -122,6 +122,25 @@ type hyperstartProxyCmd struct {
...
+	Hostname   string                `json:"hostname"`
+	Containers []hyperJson.Container `json:"containers,omitempty"`
+	Interfaces []HyperNetIface       `json:"interfaces,omitempty"`
+	ShareDir   string                `json:"shareDir"`
+}
+
+// HyperNetIface is the structure corresponding to the format expected
+// by hyperstart to setup the network on the guest.
+type HyperNetIface struct {
```
dlespiau: missing MTU

hyperstart.go
```diff
@@ -122,6 +122,25 @@ type hyperstartProxyCmd struct {
...
+	Device    string `json:"device,omitempty"`
+	NewDevice string `json:"newDevice,omitempty"`
```
dlespiau: The JSON name is actually newDeviceName

hyperstart.go
```diff
@@ -122,6 +122,25 @@ type hyperstartProxyCmd struct {
...
+	IPAddress string `json:"ipAddress"`
```
dlespiau: there's a ipAddresses field that we currently use in our runtime to specify multiple addresses (related to my earlier question)

dlespiau:
An idea could be to look at https://github.com/01org/cc-oci-runtime/blob/master/src/proxy.c#L1192 to see what we do today.

cni.go
sboeuf: yep you're right !

hyperstart.go
sboeuf: Oh yes that's right !

cni.go
dlespiau: and actually ++ as well!

network.go
sboeuf: No we cannot do that but I don't think we want to. We have several endpoints in virtcontainers, describing the veth end point inside the netns. For each of these endpoints, there is only one ipAddress, and we can give hyperstart as many as he wants. We don't have the framework to support this in virtcontainers right now, and I am wondering about the use cases.

hyperstart.go
sboeuf: Ditto.

network.go
dlespiau: ok, that makes more sense. Not totally sure if we need support for multiple IPs on a single interface, my guess is yes for CNM, and don't know for CNI. We'll find out soon enough if we start testing the different plugins.

sboeuf: Oh no wait, we can do that, just adding each ipAddress found for the same interface index.

hyperstart.go
sboeuf: done

hyperstart.go
sboeuf: done

network.go
sboeuf: Will be done in a follow up PR. Results held inside Endpoint.Properties has to be modified to handle this properly.

sboeuf:
@dlespiau I addressed all your comments (including the one about several IPs for an interface ;))

sameo:
LGTM

sameo:
@dlespiau Does that look good to you now ?

PR #158, opened 2017-03-16
cnm: Interface list has to be scanned from inside the network namespace

sboeuf:
There was an error in the code about the scan of interfaces found.
Indeed, this scan has to be performed inside the right netns to be
right. Otherwise, we are scanning the host interfaces.


dlespiau:
lgtm

PR #160, opened 2017-03-17
network: Remove ciao/networking/libsnnet dependency

sboeuf:
Because we need to use netlink Go package, and because libsnnet also
relies on this same package, we cannot have both packages at the same
time. It is caused by the vendoring in ciao while we are following the
upstream in virtcontainers.

This patch removes all dependencies related to libsnnet package and
it allows us to remove the ciao logger introduced to workaround the
same kind of issue we had with glog package. The difference here is
that glog was not a big deal while netlink is really needed. And one
more good reason is that libsnnet was not doing a lot more than what
netlink does.


sameo:
@sboeuf Can we please split the logger removal part into its own patch ?

So, if I understand you correctly here, the reason for removing the libsnnet dependency is because we need to rely on an more recent netlink Go package ? If that's so, have considered changing the ciao vendoring an improving libsnnet instead ?

sboeuf:
@sameo Yes of course I can remove the logger in a separate patch.
The reason for this patch is that we want to rely on the vishvananda/netlink package because I need it to list the routes for a specific namespace in a future PR. And because libsnnet vendor this package, we cannot import it inside virtcontainers. We agreed with Manohar that using libsnnet inside virtcontainers has no real added value compared to use the netlink package directly, and it is so painful to modify libsnnet each time we need something new.
Another good reason is that Manohar suggested that use the netlink package relying on the Handler type instead of doing direct calls to the netlink functions, meaning we are saving performances, avoiding to open/close the netlink socket each time we want to send a command.

sameo:
@sboeuf fair enough. Please split this PR in 2 and we'll merge it.

sameo:
LGTM

PR #161, opened 2017-03-17
cnm: Fill endpoints results with routes

sboeuf:
When using CNM, we were not filling the results of each endpoint
for the Routes field. Now that we can rely on netlink package to
scan the routes inside the specified network namespace, we can
properly fill this field as expected.

As a comment, this is not needed in case of CNI because the plugin
takes care of this for us.

### DIFF
--- a/cnm.go
+++ b/cnm.go
@@ -17,11 +17,15 @@
 package virtcontainers
 
 import (
+	"fmt"
 	"net"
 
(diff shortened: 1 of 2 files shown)


sboeuf:
Blocked by #160

sameo:
LGTM

PR #162, opened 2017-03-17
hyperstart: Fix IP netmask and route gateway

sboeuf:
Gateway parameter given to hyperstart was sometimes not filled
properly in case it was nil. After a specific check, we can be
sure the gateway value is proper.

With the same idea, the netMask needed to be processed from the
hexadecimal value because hyperstart expect it as an integer.

At last, we don't need NetMask field in the hyperstart structure
HyperNetIface because it is already contained into each structure
HyperIPAddress.

### DIFF
--- a/hyperstart.go
+++ b/hyperstart.go
@@ -138,7 +138,6 @@ type HyperNetIface struct {
 	Device      string           `json:"device,omitempty"`
 	NewDevice   string           `json:"newDeviceName,omitempty"`
 	IPAddresses []HyperIPAddress `json:"ipAddresses"`
-	NetMask     string           `json:"netMask"`
 	MTU         string           `json:"mtu"`
 	MACAddr     string           `json:"macAddr"`
 }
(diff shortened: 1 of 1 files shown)


sboeuf:
@sameo @jodh-intel Could you take a look at this one ?

hyperstart.go
```diff
@@ -222,9 +223,14 @@ func (h *hyper) buildNetworkInterfacesAndRoutes(pod Pod) ([]HyperNetIface, []hyp
 		ifaces = append(ifaces, iface)
 
 		for _, r := range endpoint.Properties.Routes {
+			gateway := r.GW.String()
+			if gateway == "<nil>" {
```
jodh-intel: Ouch! Is that a package bug you're having to work around here?

sboeuf: Yeah... I don't like that either. Take a look at https://golang.org/pkg/net/#IP.String

jodh-intel:
@sboeuf - the changes look fine to me. It might be worth adding a comment about / reference to pkg/net regarding that strange `<nil>` gateway string.

Could you add any new tests for these changes?

sboeuf:
@jodh-intel Sure I can add a comment for this, but it is a pretty complex unit test here. Maybe worth to do that in a following PR, breaking down the big function buildNetworkInterfacesAndRoutes().
Indeed, it depends on fetching the network stored internally inside local filesystem, then it scans interfaces from current netns...
Or I can create a function to check this IP.String() return, meaning I can easily add a unit test for this tiny function.

jodh-intel:
I was thinking of something to tests a bit more than `IP.String()`. However, given the difficulties of creating such tests, I'll defer to others on the decision as to whether we need them in this PR.

lgtm.

dlespiau:
Not sure why pullapprove hasn't picked up @jodh-intel lgtm, he's in the maintainers group. I also think we'd like unit tests around those netwokring JSON payloads. We need to at least file an issue on that. Anyway:

lgtm.

PR #163, opened 2017-03-21
Check podID and containerID function arguments

jodh-intel:
This branch switches the logic around so that rather than passing blank `podID` and `containerID`'s around and then magically expanding them at quite a low level, the `podID` and `containerID` are set as soon as possible by `virtc`. This allows all functions to validate the parameters are not blank and return an error if they are [1].

The logic is complicated by the fact that in certain contexts -- for example when creating particular `podResource`'s -- `containerID` *may* be blank. To handle this, I've created a new function called `resourceNeedsContainerID()`.

Note that this is a proof-of-concept branch - the tests have not been updated so will fail.

------
[1] - This allows "fail-fast" behaviour which is less confusing when attempting to debug issues.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -22,6 +22,9 @@ import (
 
 // common error objects used for argument checking
 var (
+	ErrNeedPod         = errors.New("Pod must be specified")
 	ErrNeedPodID       = errors.New("Pod ID cannot be empty")
 	ErrNeedContainerID = errors.New("Container ID cannot be empty")
+	ErrNeedFile        = errors.New("File cannot be empty")
+	ErrNeedState       = errors.New("State cannot be empty")
 )
--- a/filesystem.go
+++ b/filesystem.go
@@ -24,7 +24,10 @@ import (
 	"path/filepath"
 )
 
-// podResource is an int representing a pod resource type
+// podResource is an int representing a pod resource type.
+//
+// Note that some are specific to the pod itself and others can apply to
+// pods and containers.
 type podResource int
 
 const (
@@ -34,13 +37,13 @@ const (
 	// stateFileType represents a state file type
 	stateFileType
 
-	// networkFileType represents a network file type
+	// networkFileType represents a network file type (pod only)
 	networkFileType
 
 	// processFileType represents a process file type
 	processFileType
 
-	// lockFileType represents a lock file type
+	// lockFileType represents a lock file type (pod only)
 	lockFileType
 )
 
--- a/filesystem.go
+++ b/filesystem.go
@@ -145,8 +145,12 @@ func (fs *filesystem) createAllResources(pod Pod) (err error) {
 	return nil
 }
 
-func (fs *filesystem) storeFile(path string, data interface{}) error {
-	f, err := os.Create(path)
+func (fs *filesystem) storeFile(file string, data interface{}) error {
+	if file == "" {
+		return ErrNeedFile
+	}
+
+	f, err := os.Create(file)
 	if err != nil {
 		return err
 	}
@@ -161,8 +165,12 @@ func (fs *filesystem) storeFile(path string, data interface{}) error {
 	return nil
 }
 
-func (fs *filesystem) fetchFile(path string, data interface{}) error {
-	fileData, err := ioutil.ReadFile(path)
+func (fs *filesystem) fetchFile(file string, data interface{}) error {
+	if file == "" {
+		return ErrNeedFile
+	}
+
+	fileData, err := ioutil.ReadFile(file)
 	if err != nil {
 		return err
 	}
@@ -175,12 +183,32 @@ func (fs *filesystem) fetchFile(path string, data interface{}) error {
 	return nil
 }
 
-func resourceDir(podID, containerID string, resource podResource) (string, error) {
-	var path string
+// resourceNeedsContainerID determines if the specified
+// podResource needs a containerID. Since some podResources can
+// be used for both pods and containers, it is necessary to specify
+// whether the resource is being used in a pod-specific context using
+// the podSpecific parameter.
+func resourceNeedsContainerID(podSpecific bool, resource podResource) bool {
 
-	if podID == "" {
-		return "", fmt.Errorf("PodID cannot be empty")
+	switch resource {
+	case lockFileType, networkFileType:
+		// pod-specific resources
+		return false
+	default:
+		return !podSpecific
 	}
+}
+
+func resourceDir(podSpecific bool, podID, containerID string, resource podResource) (string, error) {
+	if podID == "" {
+		return "", ErrNeedPodID
+	}
+
+	if resourceNeedsContainerID(podSpecific, resource) == true && containerID == "" {
+		return "", ErrNeedContainerID
+	}
+
+	var path string
 
 	switch resource {
 	case configFileType:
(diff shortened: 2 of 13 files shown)


errors.go
```diff
@@ -0,0 +1,30 @@
...
+//
+
+package virtcontainers
+
+import (
+	"errors"
+)
+
+// common error objects used for argument checking
+var (
```
sboeuf: Those ones could be const instead, right ?

hack/virtc/main.go
```diff
@@ -227,7 +228,14 @@ func buildPodConfig(context *cli.Context) (vc.PodConfig, error) {
 		Memory: vmMemory,
 	}
 
+	id := context.String("pod-id")
```
sboeuf: I am fine with this change, but could you create the corresponding flag ? Indeed, there is no pod-id flag for the create command. And could you call it "id" instead of "pod-id" ? It is not needed to use pod-id for the naming because we are in the command pod ... thus it is pretty obvious.

filesystem.go
```diff
@@ -175,12 +181,35 @@ func (fs *filesystem) fetchFile(path string, data interface{}) error {
...
+// whether the resource is being used in a pod-specific context using
+// the podSpecific parameter.
+func resourceNeedsContainerID(podSpecific bool, resource podResource) bool {
+
+	switch resource {
+	case lockFileType, networkFileType:
+		// pod-specific resources
+		return false
+	default:
+		if podSpecific == true {
```
sboeuf: You could simplify this with "return !podSpecific"

pod.go
```diff
@@ -280,19 +287,26 @@ type PodConfig struct {
 
 // valid checks that the pod configuration is valid.
 func (podConfig *PodConfig) valid() bool {
-	if _, err := newHypervisor(podConfig.HypervisorType); err != nil {
-		podConfig.HypervisorType = QemuHypervisor
+	if podConfig == nil {
```
sboeuf: Can it happen ? I don't think so because you need podConfig to be valid (not nil) to be able to call this function, right ?

sboeuf:
@jodh-intel I have few comments, let me know what you think about. And as a global comment, could you please add some new lines after all the new closing brackets corresponding to new if statements ?

errors.go
jodh-intel: That would be good, but you can't create `const` `errors` afaik.

pod.go
jodh-intel: Oops - how did that get in there? :)

filesystem.go
jodh-intel: Agreed.

jodh-intel:
@sboeuf - volontiers! I thought the crunched up approach was "the golang way" - I much prefer letting the code breathe! :smile:

hack/virtc/main.go
jodh-intel: Added. This raises the question though of what is considered a valid podID and containerID. Since these end up being encoded in pathnames, we should take care here.

Currently containerID (via `--id=`) is free-form, whereas `podID` default to a `uuid` (although this branch would potentially allow that to change).

@sameo - Could you clarify what the approach should be here by answering a couple of questions:

- should `podID` be restricted to being a `uuid`? I'm guessing yes. If so, maybe we shouldn't allow a user to specify a uuid value manually?
- should `containerID` be restricted to being a `uuid`? I'm guessing no, but how about we restrict the name to only be able to contain characters from the set `[a-ZA-Z0-9-_.]`? (Do we want unicode support?)

errors.go
sboeuf: Ok too bad...

jodh-intel:
Added a new function `commonResourceChecks()` to reduce the cyclomatic complexity of `storeResource()` and thus hopefully appease the gods of Travis.

sboeuf:
@jodh-intel Yeah but you have to fix unit tests now ;)

jodh-intel:
@sboeuf - Yep - on it ;)

hack/virtc/main.go
sameo: @jodh-intel `podID` should be a `UUID`, yes. I agree they should just be assigned by virtcontainers and not exported.

I think `containerID` should also be a `UUID`, not to be confused with a container name (That we don't have atm).

So overall we should stop exporting the `ID` field from both `ContainerConfig` and `PodConfig` and only generate valid `UUID`s for those. 
And we should add a `Name` exported field to both `ContainerConfig` and `PodConfig` that ican be defined by the caller or we'd generate a readable one. Here only `[a-ZA-Z0-9-_.]` would be acceptable. This can be done as a separate PR.

sboeuf: @sameo according to your comment, the generation of the UUID in virtc done in this PR is not valid, right ?

jodh-intel: @sameo - In terms of the command-line API then, are you saying that we should not have a `--id=` for `pod create` or `container create`? In other words, we should have the following behaviour:

- `virtc pod create`
  (generates pod `uuid` automatically).

- `virtc container create --pod-id=$pod_id`
  (generates container `uuid` automatically).

sboeuf: Yes, but he also means that virtcontainers should be responsible for this, and we could stop to export ID in PodConfig.
type PodConfig struct {
id
....
}

sameo: > `virtc pod create`
(generates pod uuid automatically).

Correct.

> `virtc container create --pod-id=$pod_id`
(generates container uuid automatically).

Correct.

jodh-intel: I've raised #163 to make these `ID` fields private.

errors.go
```diff
@@ -22,6 +22,9 @@ import (
 
 // common error objects used for argument checking
 var (
+	ErrNeedPod         = errors.New("Pod must be specified")
 	ErrNeedPodID       = errors.New("Pod ID cannot be empty")
 	ErrNeedContainerID = errors.New("Container ID cannot be empty")
+	ErrNeedFile        = errors.New("file cannot be empty")
```
wcwxyz: Maybe "file" should be "File"?

dlespiau: actually, it seems usual to have errors starting with a lower case character in golang, so the other ones should be changed if any (a separate commit though). For reference I grepped through github.com/golang/go:

```
$ git grep -p 'errors.New.*"[A-Z]' src/ | wc -l
96
$ git grep -p 'errors.New.*"[a-z]' src/ | wc -l
1364
```

jodh-intel: Hi @wcwxyz - thanks for reminding me about this. I had planned for these errors to refer to the name of the object originally. `file` didn't fit because that error is used for a file path. But looking at this list again, it's all rather inconsistent since it should be `containerID` and `podID`. I think as you say, just capitalising `file` would be the simplest fix ;)

wcwxyz: Thanks for looking into it. I figure it's a typo.

jodh-intel: @dlespiau - interesting, but looking at the virtcontainers code:

```
$ git grep -p 'fmt.Errorf.*"[A-Z]' | wc -l
220
$ git grep -p 'fmt.Errorf.*"[a-z]' | wc -l
40
```

Maybe an opportunity here to write a new linter for `Error`'s to plug into https://github.com/alecthomas/gometalinter? :smile:

jodh-intel:
Ready for re-review.

jodh-intel:
Note that once we've decided on a strategy for #187, I can raise a follow-up PR to simplify the test changes (using a test setup fixture).

jodh-intel:
Updated `cleanUp()` function to stop `TestQemuAppendImage` from failing due to a missing `testImage` file. This fix is minimal: I plan to raise follow-up PRs to improve the tests after this PR lands.

sboeuf:
lgtm

PR #164, opened 2017-03-21
cc_proxy: Move to proxy v3.0

sboeuf:
This patch updates the cc-proxy implementation of the proxy, relying
on latest proxy changes (version 3.0).

### DIFF
--- a/cc_proxy.go
+++ b/cc_proxy.go
@@ -17,17 +17,18 @@
 package virtcontainers
 
 import (
+	"encoding/json"
 	"fmt"
 	"net"
 	"net/url"
 
-	"github.com/01org/cc-oci-runtime/proxy/api"
+	"github.com/clearcontainers/proxy/client"
 )
 
 var defaultCCProxyURL = "unix:///run/cc-oci-runtime/proxy.sock"
 
 type ccProxy struct {
-	client *api.Client
+	client *client.Client
 }
 
 // CCProxyConfig is a structure storing information needed for
@@ -36,7 +37,7 @@ type CCProxyConfig struct {
 	URL string
 }
 
-func (p *ccProxy) connectProxy(proxyURL string) (*api.Client, error) {
+func (p *ccProxy) connectProxy(proxyURL string) (*client.Client, error) {
 	if proxyURL == "" {
 		proxyURL = defaultCCProxyURL
 	}
@@ -64,25 +65,7 @@ func (p *ccProxy) connectProxy(proxyURL string) (*api.Client, error) {
 		return nil, err
 	}
 
-	return api.NewClient(conn.(*net.UnixConn)), nil
-}
-
-func (p *ccProxy) allocateProxyInfo() (ProxyInfo, error) {
-	if p.client == nil {
-		return ProxyInfo{}, fmt.Errorf("allocateIOStream: Client is nil, we can't interact with cc-proxy")
-	}
-
-	ioBase, _, err := p.client.AllocateIo(2)
-	if err != nil {
-		return ProxyInfo{}, err
-	}
-
-	proxyInfo := ProxyInfo{
-		StdioID:  ioBase,
-		StderrID: ioBase + 1,
-	}
-
-	return proxyInfo, nil
+	return client.NewClient(conn), nil
 }
 
 // register is the proxy register implementation for ccProxy.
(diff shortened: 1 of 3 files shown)


sboeuf:
I have added do-not-merge label because we have to wait for https://github.com/clearcontainers/proxy/pull/11 to be merged before we merge this one. Indeed, we don't want to break virtcontainers until the proxy 3.0 is completely merged and ready.

cc_proxy.go
```diff
@@ -205,10 +203,21 @@ func (p *ccProxy) sendCmd(cmd interface{}) (interface{}, error) {
...
 	}
 
+	var tokens []string
+	if proxyCmd.token == "" {
+		tokens = nil
+	} else {
+		tokens = append(tokens, proxyCmd.token)
+	}
+
+	if err := p.client.HyperWithTokens(proxyCmd.cmd, tokens, json.RawMessage(data)); err != nil {
```
jodh-intel: This function could just end with:

```
return nil, p.client.HyperWithTokens(proxyCmd.cmd, tokens, json.RawMessage(data))
```

But that makes me wonder, why does this function always return a `nil` `interface`? Do we envisage a proxy that might not return this value at some point? If not, why don't we just return an `error`?

sboeuf: Good question... I cannot remember why we decided to return this interface initially. Let me check this and I will remove it if it is not needed.

dlespiau:
There is a "detail" that needs figuring out still, I talked about it very briefly at the f2f and filed an issue for that: https://github.com/clearcontainers/proxy/issues/21.

Right now, there's no buffering of stdout data in the proxy, so the shim needs to be ready to receive stdout before execcmd is issued.

cc_proxy.go
```diff
@@ -64,25 +65,7 @@ func (p *ccProxy) connectProxy(proxyURL string) (*api.Client, error) {
...
-		return ProxyInfo{}, err
-	}
-
-	proxyInfo := ProxyInfo{
-		StdioID:  ioBase,
-		StderrID: ioBase + 1,
-	}
-
-	return proxyInfo, nil
+	return client.NewClient(conn.(*net.UnixConn)), nil
```
dlespiau: Hum, net.UnixConn isn't a requirement for a client anymore as we don't do fd passing in 3.0. Filed https://github.com/clearcontainers/proxy/issues/23 to remind us.

sboeuf:
@dlespiau about clearcontainers/proxy#21, we are already doing what you described for the create/start sequences in runtime 3.0 + virtcontainers.
But I agree that regarding exec, it is more complicated as we don't have 2 separate commands to:
- first: retrieve the token only
- second: exec the exec command

I am not sure we want to modify virtcontainers so that we split this thing, and maybe the proxy should queue the output for a bit in case of exec (for a bit means not for a long time because shim connection should happen very quick). And thinking about future, maybe one shim will handle a pod as Manohar mentioned it :)

@sameo what is your opinion about this issue ?

cc_proxy.go
sameo: @jodh-intel The returned `interface{}` is for handling proxy implementations that would return something else than an error as part of their send command. `cc_proxy` is not in that category, so it's always nil.

cc_proxy.go
sboeuf: Yes good point, I have made the change, anticipating your change in proxy 3.0.

sameo:
@sboeuf @dlespiau See my comment on clearcontainers/proxy#21

cc_proxy.go
jodh-intel: This is starting to get a bit off-topic for this PR, but every call to `sendCmd()` currently throws away the returned `interface{}`. Is that an oversight, or can we just drop this `interface{}` value until we come across a proxy that actually needs it?

Also, I'm not clear from the `proxy` interface how this value should actually be used. Maybe some more doc would be helpful here (not required as part of this PR though).

dlespiau:
For the record, this looks good to me, the interface definition is orthogonal to the PR as @jodh-intel noted.

lgtm

cc_proxy.go
jodh-intel: Raised #170 to track this.

jodh-intel:
lgtm.

sboeuf:
Now that we have the proxx 3.0 ready, I have rebased this PR and re-uploaded it. We should merge it :)

PR #165, opened 2017-03-22
virtc: Check mandatory command-line arguments.

jodh-intel:
Ensure that all mandatory arguments are checked early.

Note that urfave/cli doesn't actually provide such a feature. The
approach taken attempts to avoid duplicated code by checking centrally,
rather than peppering checks on podID and containerID in lots of
functions.

### DIFF
--- a/hack/virtc/main.go
+++ b/hack/virtc/main.go
@@ -250,6 +250,64 @@ func buildPodConfig(context *cli.Context) (vc.PodConfig, error) {
 	return podConfig, nil
 }
 
+// checkRequiredPodArgs checks to ensure the required command-line
+// arguments have been specified for the pod sub-command specified by
+// the context argument.
+func checkRequiredPodArgs(context *cli.Context) error {
+	if context == nil {
+		return fmt.Errorf("BUG: need Context")
+	}
+
+	// sub-sub-command name
+	name := context.Command.Name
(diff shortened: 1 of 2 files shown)


hack/virtc/main.go
```diff
@@ -250,6 +250,63 @@ func buildPodConfig(context *cli.Context) (vc.PodConfig, error) {
...
+		fallthrough
+	case "list":
+		fallthrough
+	case "run":
+		return nil
+	}
+
+	id := context.String("id")
+	if id != "" {
+		return nil
```
sboeuf: You should test id == "" and return the error here, rather than return nil from here. It would be more consistent with the way we write functions usually, don't you think ?

sboeuf:
@jodh-intel only one comment, and you also need to fix travis, but other than that, we can merge this.

jodh-intel:
Hi @sboeuf - I've fixed the main issue causing the Travis failure. However, I was then seeing test `TestRemoveNetworkFailureNetworkDoesNotExist` fail. This didn't make much sense, but after debugging I can see the failure on `master` locally. This is confirmed by re-running the Travis tests on `master` which are currently failing even though the last build for the same commit passed).

sboeuf:
lgtm

Issue #166, opened 2017-03-23
TestRemoveNetworkFailureNetworkDoesNotExist is failing on master branch


jodh-intel:
Travis had previously passed the code at commit 8362442641ae0a370ba54ebb4d9c22aedd5d240e, but re-running the Travis CI tests now shows it is failing on test `TestRemoveNetworkFailureNetworkDoesNotExist`.

See https://github.com/containers/virtcontainers/pull/165#issuecomment-288696677

dlespiau:
PR #167 seems to have fixed that? Note github supported auto-closing issues based on tags in commit messages: https://help.github.com/articles/closing-issues-via-commit-messages/

PR #167, opened 2017-03-23
pkg: cni: Fix unit testing after CNI plugins update

sboeuf:
In order to follow more closely the CNI specification, some recent changes have been added to the CNI plugins so that they are not failing anymore in case we try to tear down a non existing network.
This patch fixes the unit test which was expecting to see a failure. Now, we are expecting a success in removing a network that we've never created before.

This PR fixes issue #166


sameo:
LGTM

jodh-intel:
lgtm.

PR #173, opened 2017-04-01
epause: remove SIGKILL handler

wcwxyz:
signal SIGKILL (as well as SIGSTOP) cannot be caught or ignored. And we
should remove this.


sameo:
LGTM

PR #174, opened 2017-04-03
Duplicate hyperstart API

sboeuf:
This PR adds a new package to virtcontainers: pkg/hyperstart/api. This package duplicates/modify structures initially provided by github.com/hyperhq/runv/hyperstart/api/json, in order to prevent issues related to upstream modification of this package.
That way, the package provided provided by hyperstart can change according to the changes made to the binary, there is no way it can break our virtcontainers library. The clearcontainers/proxy will also rely on this new package.


sameo:
LGTM

PR #177, opened 2017-04-04
gitignore: Update the .gitignore file

dlespiau:
- Use absolute ignore rules for single binaries
- Ignore objects files (pause.o wasn't ignored)


jodh-intel:
lgtm

PR #178, opened 2017-04-04
virtcontainers: Replace glog by logrus logger

sboeuf:
Glog causes some data race with project using our library and glog at the same time. For that reason, it is safer to move to logrus logger, which is more flexible and which does not cause such issues.

### DIFF
--- a/qemu.go
+++ b/qemu.go
@@ -27,7 +27,7 @@ import (
 
 	ciaoQemu "github.com/01org/ciao/qemu"
 	"github.com/01org/ciao/ssntp/uuid"
-	"github.com/golang/glog"
+	log "github.com/Sirupsen/logrus"
 )
 
 type qmpChannel struct {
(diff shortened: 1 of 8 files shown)


sboeuf:
@sameo @dlespiau please take a look because we need this to avoid go test -race to fail in the proxy.

qemu.go
```diff
@@ -76,19 +76,23 @@ const (
 type qmpGlogLogger struct{}
```
jodh-intel: Shouldn't this be renamed to `qmpLogger` or similar now?

sameo:
LGTM, except for the qemu logger renaming.

dlespiau:
side notes:

-  we should promote more structured logging in virtcontainers instead of just free form messages
- it may be interesting to delegate the actual logging to the user of the virtcontainers library. Something like gokit's logging interface would help a log there.

PR #179, opened 2017-04-05
pkg: hyperstart: Wait for the end of the CTL channel

sboeuf:
This PR mainly adds a new function to our hyperstart package so that a consumer of this package can properly wait for the end of the CTL channel before to go further.
The implementation relies on a Go channel written by the multicast loop when it cannot read anymore on the CTL channel.

### DIFF
--- a/pkg/hyperstart/hyperstart.go
+++ b/pkg/hyperstart/hyperstart.go
@@ -123,6 +123,8 @@ type Hyperstart struct {
 	ctlMutex sync.Mutex
 
 	ctlMulticast *multicast
+
+	ctlChDone chan interface{}
 }
 
 // NewHyperstart returns a new hyperstart structure.
@@ -186,7 +188,8 @@ func (h *Hyperstart) OpenSockets() error {
 		return err
 	}
 
-	h.ctlMulticast = startCtlMonitor(h.ctl)
+	h.ctlChDone = make(chan interface{})
+	h.ctlMulticast = startCtlMonitor(h.ctl, h.ctlChDone)
 
 	return nil
 }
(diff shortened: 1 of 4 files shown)


sboeuf:
@dlespiau please take a look at this and let me know what you think about the way I fixed it. This code is supposed to be used by the proxy here: https://github.com/clearcontainers/proxy/blob/master/vm.go#L560 by adding `vm.hyperHandler.WaitCtlEnd()` before to close the channel.

dlespiau:
I think we should make this wait part of CloseSockets(): OpenSockets() creates this goroutine, makes sense to me we synchronise in CloseSockets(). This way we don't need to add an extra API entry point either.

Note that OpenSockets() and CloseSockets() could be also renamed Open() and Close(), probably more idiomatic.

sboeuf:
@dlespiau I thought about putting this into CloseSockets(), but I think there is a case (the real use case) where it won't work. Indeed, in case of virtcontainers, we first unregisterVM and then we shutdown the VM. Don't you think in that case we will wait for the connection to be broken but it will never happen.

sboeuf:
Well, maybe not. Actually the way you designed the proxy, you are not closing the sockets with the VM when unregisterVM is called, right ?
Then I guess it is okay to put this into CloseSockets().
About changing the names for OpenSockets and CloseSockets to Open and Close, that sounds okay to me, but I'll do that in a following PR.

sboeuf:
This made me change the way unit tests behave but we should be good now.

dlespiau:
lgtm

sameo:
LGTM

pkg/hyperstart/hyperstart.go
```diff
@@ -186,14 +188,19 @@ func (h *Hyperstart) OpenSockets() error {
...
 	return nil
 }
 
 // CloseSockets closes both CTL and IO sockets.
 func (h *Hyperstart) CloseSockets() error {
 	if !h.ctlState.closed() {
+		if h.ctlChDone != nil {
+			<-h.ctlChDone
+		}
+
```
sameo: This is a bit racy as `CloseSockets()` could be closed twice before `h.ctlState.close()` is called.
I would simply add `close (h.ctlChDone)` to `func (c *connState) close()` to fix that issue.

pkg/hyperstart/multicast.go
```diff
@@ -66,6 +66,8 @@ func startCtlMonitor(ctlConn net.Conn) *multicast {
 				break
 			}
 		}
+
+		done <- true
```
sameo: `close(done)`

PR #180, opened 2017-04-05
ci: Fix the build because of cni build name change

sboeuf:
The build script from CNI supposed building all the plugins we need has been renamed from "build" to "build.sh". We have to fix that in our CI script in order to let our build properly pass.


dlespiau:
lgtm

PR #181, opened 2017-04-06
Improve logging

sboeuf:
Because virtcontainers and its packages are all packages, we wanted to be able to specify from the caller (such as the runtime) a different logger than the default one.

### DIFF
--- a/api.go
+++ b/api.go
@@ -19,7 +19,16 @@ package virtcontainers
 import (
 	"os"
 	"syscall"
+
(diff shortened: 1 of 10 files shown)


sboeuf:
@sameo let me know if this one is okay for you ? Please merge it, if it is.

jodh-intel:
I wonder if it would be worth implementing custom formatters for the two new loggers? Other than that,

lgtm.

PR #182, opened 2017-04-06
Make pod console path internal

sboeuf:
There is no reason to configure the pod console path from the PodConfig because it is used only internally by the proxy and it is created by the hypervisor. That's why this PR takes care of removing the Console field from PodConfig, and provide an internal way for both hypervisor and proxy to be able to get this console path.


sboeuf:
@sameo let me know for this PR ! Merge it if you're fine with this.

sameo:
LGTM

PR #188, opened 2017-04-10
proxy: Clarify sendCmd() expected usage

### LINKED ISSUE #170: Clarify or remove interface return value for sendCmd proxy interface method.
the `proxy` interface includes:

```
sendCmd(cmd interface{}) (interface{}, error)
```

However, it is unclear what `interface{}` is for given:

- It is not documented.
- Every call to `sendCmd()` throws away the first return value.

See https://github.com/containers/virtcontainers/pull/164#discussion_r107360822 for further details.

sboeuf:
Because we can pass an interface as a parameter, and receive another interface by using sendCmd() from proxy interface, this was a bit unclear about what we can expect from this function. This commit clarify the usage thanks to some comments.

Fixes #170

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -114,6 +114,14 @@ type proxy interface {
 	// disconnect disconnects from the proxy.
 	disconnect() error
 
-	// sendCmd sends a command to the agent inside the VM through the proxy.
+	// sendCmd sends a command to the agent inside the VM through the
+	// proxy.
+	// This function will always be used from a specific agent
+	// implementation because a proxy type is always tied to an agent
+	// type. That's the reason why it takes an interface as parameter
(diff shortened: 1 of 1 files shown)


jodh-intel:
The comment looks good, but I still don't understand why we always throw away the returned `interface{}`...?

sboeuf:
@jodh-intel we throw it away in case of the pair hyperstart(agent) <=> ccProxy (proxy). But we could have in the future another pair agent <=> proxy using the data returned from sendCmd(), meaning we would not throw them away. Does that make more sense ?

jodh-intel:
OK - in terms of the title, this PR does what it says, so...

lgtm.

PR #191, opened 2017-04-11
virtc: Add cc-shim support

sboeuf:
This PR modifies slightly virtcontainers structures Pod and Process to have something working both with virtc CLI and OCI runtime.
This enables the support for cc-shim inside virtc so that we can use the entire chain (shim+proxy) with virtc CLI.

### DIFF
--- a/hack/virtc/README.md
+++ b/hack/virtc/README.md
@@ -75,6 +75,29 @@ I0410 08:58:49.059044    5384 proxy.go:566] proxy started
 ```
 The proxy socket specified in the example log output has to be used as `virtc`'s `--proxy-url` option.
 
+### Get cc-shim (optional)
+
+If you plan to start `virtc` with the hyperstart agent (implying the use of `cc-proxy` as a proxy), you will have the possibility to enable [cc-shim](https://github.com/clearcontainers/shim) in order to interact with the process running inside your container. First, you will have to perform extra steps to setup your environment.
+
+```
+$ go get github.com/clearcontainers/shim
+$ make
+$ sudo make install
+```
+
+The shim will be installed at the following location: `/usr/libexec/cc-shim`. There will be two cases where you will be able to enable `cc-shim`:
+
+_Start a new container_
+
+```
+# ./virtc container start --id=1 --pod-id=306ecdcf-0a6f-4a06-a03e-86a7b868ffc8 --cc-shim --cc-shim-path="/usr/libexec/cc-shim"
+```
+_Execute a new process on a running container_
+```
+# ./virtc container enter --id=1 --pod-id=306ecdcf-0a6f-4a06-a03e-86a7b868ffc8 --cmd="/bin/ifconfig" --cc-shim --cc-shim-path="/usr/libexec/cc-shim"
+```
+Notice that in both cases, the `--pod-id` and `--id` options have been defined when creating a pod and a container respectively. 
+
 ### Run virtc
 
 All following commands __MUST__ be run as root. By default, and unless you decide to modify it and rebuild it, `virtc` starts empty pods (no container started).
--- a/pod.go
+++ b/pod.go
@@ -53,6 +53,7 @@ const (
 // State is a pod state structure.
 type State struct {
 	State stateString `json:"state"`
+	URL   string      `json:"url,omitempty"`
 }
 
 // valid checks that the pod state is valid.
@@ -353,8 +354,6 @@ type Pod struct {
 	runPath    string
 	configPath string
 
-	url string
-
 	state State
 
 	lockFile *os.File
@@ -377,7 +376,7 @@ func (p *Pod) Annotations(key string) (string, error) {
 
 // URL returns the pod URL for any runtime to connect to the proxy.
 func (p *Pod) URL() string {
-	return p.url
+	return p.state.URL
 }
 
 // GetContainers returns a container config list.
@@ -386,7 +385,8 @@ func (p *Pod) GetContainers() []*Container {
 }
 
 func (p *Pod) createSetStates() error {
-	err := p.setPodState(StateReady)
+	p.state.State = StateReady
+	err := p.setPodState(p.state)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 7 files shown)


sboeuf:
https://github.com/clearcontainers/runtime/pull/70 is the following PR to update the runtime.

hack/virtc/README.md
```diff
@@ -75,6 +75,28 @@ I0410 08:58:49.059044    5384 proxy.go:566] proxy started
...
+$ make
+$ sudo make install
+```
+
+The shim will be installed at the following location: `/usr/libexec/cc-shim`. There will be two cases where you will be able to enable `cc-shim`:
+
+_Start a new container_
+
+```
+# ./virtc container start -id=1 -pod-id=306ecdcf-0a6f-4a06-a03e-86a7b868ffc8 -cc-shim -cc-shim-path="/usr/libexec/cc-shim"
```
jodh-intel: - Although single-dash options might be allowed, I think we should stick with documenting using double-dashes (since that's what `virtc container --help` for example shows).
- Also, how about changing `-pod-id=` to:
  ```
  pod=$(uuidgen -r)
  ./hack/virtc/virtc container start -id=1 --pod-id="$pod" ...
  ```

hack/virtc/README.md
```diff
@@ -75,6 +75,28 @@ I0410 08:58:49.059044    5384 proxy.go:566] proxy started
...
+```
+_Execute a new process on a running container_
+```
+# ./virtc container enter -id=1 -pod-id=306ecdcf-0a6f-4a06-a03e-86a7b868ffc8 -cmd="/bin/ifconfig" -cc-shim -cc-shim-path="/usr/libexec/cc-shim"
```
jodh-intel: `--pod-id="$pod"`.

hack/virtc/README.md
sboeuf: @jodh-intel I agree with those changes, but I think we want them in a follow up PR, right ?

jodh-intel: well, the suggested changes relate to new lines you've added in this PR so surely just as easy to add it to this PR? I think it's clearer to generate a fresh uuid as the (presumably) random one you've chosen may confuse readers (because it's not clear it's random and it might refer to an existing pod on the machine).

pod.go
```diff
@@ -339,8 +339,6 @@ type Pod struct {
 	runPath    string
 	configPath string
 
-	url string
-
```
sboeuf: The only case would be the proxy opening different sockets for different processes. Every time we call Register or Attach to reach the proxy, we receive a URL, we can assume it could be different for some reasons (maybe performance).
The main reason for this change is that in case of starting a container, we need to get the URL to start the shim. But even if I return the pod from StartContainer() API, pod.URL() returns an empty string. Indeed, the URL has been obtained when we did createContainer() from hyperstart.go, meaning there is no way we can get this URL at the time of StartContainer() in case of virtc. The only way would be to store the URL in the PodConfig, but this is not a config coming from the user, and if we set this field as private, the JSON marshalling won't work.
That's why I have thought that using the Process structure would be a good idea because we store it already, and we are covering the possibility to get different URLs from the proxy.

hack/virtc/README.md
sboeuf: Okay I will change the simple dash ;)
About the pod id, I can understand the confusion, but you're not supposed to generate this pod id when you're doing `container start` or `container enter`. I think it will be more confusing. We should add a sentence saying that the pod-id and the container id have already been created, shouldn't we ?

jodh-intel: Ah - in that case yes a comment explaining that the pod-id used in the examples refers to a previously-created pod is fine.

pod.go
sameo: > The only case would be the proxy opening different sockets for different processes. Every time we call Register or Attach to reach the proxy, we receive a URL, we can assume it could be different for some reasons (maybe performance).

As processes from a given pod must be running in the same pod, they all should be handled through the same proxy. There will be one socket per proxy, and thus one URL per pod.

Without digging into the code, `pod.url` should be set by `hyperstart.go` when we start the agent, i.e. when we start a pod. That should be part of a pod state, and we should be storing that anyway.

sboeuf: Ok let's store this as a Pod state!

sboeuf:
This PR depends on https://github.com/clearcontainers/proxy/pull/31

sboeuf:
@sameo let me know your opinion about this new version.

pod.go
```diff
@@ -711,11 +713,7 @@ func (p *Pod) enter(args []string) error {
 	return nil
 }
 
-func (p *Pod) setPodState(state stateString) error {
-	p.state = State{
-		State: state,
-	}
-
+func (p *Pod) setPodState() error {
```
sameo: Either we keep that method name and we keep its argument or we rename it to `storePodState()`. I'd prefer the former.

api.go
```diff
@@ -474,45 +474,45 @@ func DeleteContainer(podID, containerID string) (*Container, error) {
 
 // StartContainer is the virtcontainers container starting entry point.
 // StartContainer starts an already created container.
-func StartContainer(podID, containerID string) (*Container, error) {
+func StartContainer(podID, containerID string) (*Pod, *Container, error) {
```
sameo: I'd prefer if we keep the `StartContainer()` prototype and provide an additional `pod.go:PodFromID()` getter instead.

pod.go
sboeuf: Done.

api.go
sboeuf: If we create a new public function `PodFromID()`, we will have to fetch the pod from this method (which has a performance cost), while we already have this pod from inside the `StartContainer()`.
I think it duplicates things, doesn't it ?

sboeuf: We already did the same (returning the Pod) for the EnterContainer() so that our runtime 3.0 was able to get the pod URL. Either with keep this solution or we need to change EnterContainer also.

sboeuf:
@sameo let me know what you think about my comments, and it this PR needs further changes.

api.go
sameo: @sboeuf so then we should add a `container.URL()` method to fetch the URL a container is reachable through (i.e. its Pod URL). `Container` points back to its Pod, so we should use that to avoid complicating our API.

sboeuf: @sameo This solution sounds good to me :)

sboeuf: Done !

sameo:
LGTM

PR #192, opened 2017-04-13
pkg: oci: fix container rootfs path

wcwxyz:
As OCI spec stated, ocispec.Root.Path is either an absolute path or a
relative path to the bundle.


sboeuf:
@wcwxyz nice catch !
LGTM

jodh-intel:
@wcwxyz - nice! Thanks!

lgtm.

PR #195, opened 2017-04-13
cc_proxy: Don't provide a token for the pause container

sboeuf:
Because we don't need to send/receive any IO to/from the pause container workload, there is no point in passing a token to the proxy.
That way, the proxy does not wait for a token to be provided, and therefore it does not expect a shim to connect.


dlespiau:
lgtm

PR #196, opened 2017-04-13
docs: Replace single dash with double dashes in README.md

sboeuf:
Even if the single dash is supported, it is more generic to use double dashes when command line options are more than one character.


jodh-intel:
lgtm

PR #198, opened 2017-04-14
cc_proxy: Do not marshal twice

sameo:
Our payload will be marshalled by the proxy already.


sboeuf:
LGTM

PR #203, opened 2017-04-20
errors: Make the standard errors private.

jodh-intel:
Only expose the minimum set of standard errors
("virtc" uses ErrNeedPodID and ErrNeedContainerID).


dlespiau:
From an API stand point, I don't think we should go half-way, exposing some errors but not others. Either expose all the errors or none :)

sboeuf:
I agree with @dlespiau that it should be consistent. And I would prefer all those errors to be private. There is no point in having them exposed through the virtcontainers API.

jodh-intel:
ack - branch updated.

sboeuf:
LGTM

Issue #202, opened 2017-04-19
hyperstart start() and stop() implementations are not hyperstart specific


sameo:
Our only agent implementation, `hyperstart.go`, `start()` and `stop()` implementations are not hyperstart specific. They do pod specific initializations that should be factored into `pod.go`. All agent implementations will have to follow the same steps.

Going one step further, we could argue the `start()` and `stop()` agent interface methods are not needed. I would like to see them removed unless someone has a case for keeping them around. 

In any case, the hyperstart implementation for those 2 methods should be empty.

sboeuf:
Fixed by #205.

PR #205, opened 2017-04-20
Move generic code out of hyperstart agent implementation

sboeuf:
A lot of code related to the proxy and the shim was present inside hyperstart agent implementation. The problem is that this code is generic for any pod/container, and deserves to be shared across different agent implementations.

After all generic code has been moved from agent implementations to pod/container code, there is no point in having `start()` and `stop()` functions provided by the agent interface. Our agent implementations are empty for those functions, and we cannot think about a future usage for any future agent.

This PR fixes issue #202. It has been tested with the runtime and everything still works as before.

### DIFF
--- a/container.go
+++ b/container.go
@@ -120,6 +120,43 @@ func (c *Container) URL() string {
 	return c.pod.URL()
 }
 
+func (c *Container) startShim() error {
+	proxyInfo, url, err := c.pod.proxy.connect(*(c.pod), true)
+	if err != nil {
+		return err
+	}
+
+	if err := c.pod.proxy.disconnect(); err != nil {
+		return err
+	}
+
+	if c.pod.state.URL != url {
+		return fmt.Errorf("Pod URL %s and URL from proxy %s MUST be identical", c.pod.state.URL, url)
+	}
+
+	shimParams := ShimParams{
+		Token:   proxyInfo.Token,
+		URL:     url,
+		Console: c.config.Console,
+	}
+
+	pid, err := c.pod.shim.start(*(c.pod), shimParams)
+	if err != nil {
+		return err
+	}
+
+	c.process = Process{
+		Token: proxyInfo.Token,
+		Pid:   pid,
+	}
+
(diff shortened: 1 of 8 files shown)


container.go
```diff
@@ -120,6 +120,43 @@ func (c *Container) URL() string {
...
+	if err != nil {
+		return err
+	}
+
+	if err := c.pod.proxy.disconnect(); err != nil {
+		return err
+	}
+
+	if c.pod.state.URL != url {
+		return fmt.Errorf("Pod URL %s and URL from proxy %s MUST be similar", c.pod.state.URL, url)
```
jodh-intel: Shouldn't this be "identical" rather than "similar" (same comment for identical error below)?

pod.go
```diff
@@ -606,33 +603,76 @@ func (p *Pod) startVM() error {
 		return fmt.Errorf("Did not receive the pod started notification")
 	}
 
-	err := p.agent.start(p)
+	virtLog.Infof("VM started")
+
+	return nil
+}
+
+// startShims starts registers all containers to the proxy and starts
```
jodh-intel: s/starts registers/registers/g

sshd.go
```diff
@@ -105,12 +105,32 @@ func (s *sshd) init(pod *Pod, config interface{}) error {
...
-func (s *sshd) start(pod *Pod) error {
+// exec is the agent command execution implementation for sshd.
+func (s *sshd) exec(pod *Pod, c Container, process Process, cmd Cmd) error {
 	if pod == nil {
 		return errNeedPod
 	}
 
+	session, err := s.client.NewSession()
+	if err != nil {
+		return fmt.Errorf("Failed to create session")
```
jodh-intel: It would be useful to add some more detail to this error: which pod/container for example?

sameo:
Once @jodh-intel are adressed, this LGTM.

sboeuf:
Comments from @jodh-intel have been addressed.

jodh-intel:
lgtm

Issue #204, opened 2017-04-20
Move Console field from ContainerConfig to Cmd structure


sboeuf:
A console path can be provided every time a new process is started, and we know we can start several processes on one container as long as the process on this container is still running.
That's why we have to move Console field to Cmd structure so that every process can be redirected to a different console.

sboeuf:
Fixed by #209

PR #206, opened 2017-04-20
Add the ability to specify a different console for a new process on an existing container

sboeuf:
There is a limitation regarding the ability to provide a different console than container's one in case we are trying to execute a new process on this existing container. Indeed, because the console is tied to the ContainerConfig structure, we have no way to specify a different console for another process.

This PR moves Interactive and Console fields from the structure ContainerConfig to Cmd, solving our current limitation.

This PR fixes issue #204.

### DIFF
--- a/container.go
+++ b/container.go
@@ -44,12 +44,6 @@ type ContainerConfig struct {
 	// RootFs is the container workload image on the host.
 	RootFs string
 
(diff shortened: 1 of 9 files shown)


sboeuf:
Do not merge until #205 is merged.

jodh-intel:
lgtm

PR #207, opened 2017-04-21
api: ListPod() should succeed if pod directory does not exist.

jodh-intel:
Previously ListPod() returned an error if the pod directory did not
exist. However, that is the normal situation on a system where no pods
have been created.

Renamed "TestListPodFailing" test to "TestListPodNoPodDirectory"
for clarity.


sboeuf:
LGTM

dlespiau:
That's definitely nicer!

lgtm

PR #210, opened 2017-04-21
qemu: Prevent udev from creating predictable network interface names

sboeuf:
Because we rely on the MAC address to find the network interface, we don't need to enable the renaming by udev of the interface created.

More than that, because hyperstart agent waits for the interface to be ready before it can continue, we don't want the name of the interface found to change after we received the first uevent.

This PR fixes the issue explained here https://github.com/01org/cc-oci-runtime/issues/815 and fixed with https://github.com/01org/cc-oci-runtime/pull/832. Related to the topic: https://github.com/clearcontainers/hyperstart/pull/20.

### DIFF
--- a/qemu.go
+++ b/qemu.go
@@ -118,6 +118,7 @@ var kernelDefaultParams = []Param{
 	{"systemd.mask", "systemd-networkd.service"},
 	{"systemd.mask", "systemd-networkd.socket"},
 	{"cryptomgr.notests", ""},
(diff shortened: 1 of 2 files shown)


sboeuf:
@sameo can you please take a quick look at this and merge if you're fine with this change.

jodh-intel:
Related to https://github.com/01org/cc-oci-runtime/pull/832.

lgtm.

PR #212, opened 2017-04-24
api: Speed up container status query.

jodh-intel:
StatusContainer() does not need to iterate through all containers after
it has found a matching container to return the status for.

### DIFF
--- a/api.go
+++ b/api.go
@@ -630,8 +630,6 @@ func StatusContainer(podID, containerID string) (ContainerStatus, error) {
 		return ContainerStatus{}, errNeedContainerID
 	}
 
-	var contStatus ContainerStatus
-
 	pod, err := fetchPod(podID)
 	if err != nil {
 		return ContainerStatus{}, err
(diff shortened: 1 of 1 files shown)


api.go
```diff
@@ -645,10 +645,12 @@ func StatusContainer(podID, containerID string) (ContainerStatus, error) {
 				PID:    container.process.Pid,
 				RootFs: container.config.RootFs,
 			}
+			return contStatus, nil
```
dlespiau: Would be more idiomatic to get rid of the contSatus variable altogether with:

````
			return ContainerStatus{
				ID:     container.id,
				State:  container.state,
				PID:    container.process.Pid,
				RootFs: container.config.RootFs,
			}, nil
````

Wondering if returning a pointer instead of coping the struct would be a good idea as well, but that's another discussion though.

jodh-intel: @dlespiau - good call re dropping the variable - branch updated.

dlespiau:
lgtm

sboeuf:
LGTM

PR #214, opened 2017-04-25
Fix oci rootfs+bundle handling

jodh-intel:


sboeuf:
@sameo I would like you to take a look at this change because it does not follow what we did before with the `Annotations` in the `Pod` structure.
The thing is that the bundle path should be retrieved from the annotation stored inside the pod. But in the same time virtcontainers cannot be aware about those specific keys for annotations, meaning we cannot retrieve this from virtcontainers itself.
I can see two options here:
- Either you are okay with the way this PR adds a new field Bundle to `ContainerStatus` and `ContainerConfig`. This means you think a bundle is generic enough to be re-used by another runtime implementation.
- Or you think this should be runtime specific, and retrieved from the runtime itself, meaning that we need StatusPod() to return a handle on `*Pod`. We can keep the `PodStatus` structure returned which is generic, and the handle to `*Pod` would be helpful to get annotations previously stored.

sameo:
@jodh-intel @sboeuf The `ContainerConfig` and `PodConfig` structures should be kept as specification agnostic as possible. So for passing OCI specific data down to virtcontainers from `oci/utils.go` we should use `Annotations`, as we did for `Pods`.
At the same time we want to avoid returning raw `Pod` pointers back to the status callers because they're still at least one step away from being able to interpret the status call results. So I don't mind adding container specifications sections to the status structures. For example we'd have:

```Go
type OCIStatus struct {
       Bundle string
}

type ContainerStatus {
[...]
     OCI OCIStatus
}
```

The `PodConfig` and `ContainerConfig` structures are input to virtcontainers and we should avoid putting e.g. OCI specific stuff in there as it could be quite confusing. `ContainerStatus` is an output from virtcontainers and callers can only worry about the non empty sections/fields.

sboeuf:
@sameo I like the idea, but there is still one thing which is not consistent, that is the annotation key to use. Indeed, the key in the hash table to retrieve an annotation has been set by the runtime implementation. In our case, and to be more specific, it is set by the pkg/oci/utils.go from virtcontainers repo. This key is private to the oci package, therefore I guess we should make this public.
But globally, this is not the purpose with annotations I think. I mean that Annotations have been added to allow any runtime to store some specific data that virtcontainers is not even aware of. In that case, it would be asking to the runtime to store those annotations but then virtcontainers would be responsible to retrieve them. What do you think ?

sameo:
@jodh-intel pointed out that my initial proposal would lead to either:

* A circular dependency as virtcontainers would have to use the `pkg/oci/utils.go` package

or 

* Having to implement some private but OCI specific functions in virtcontainers.

@jodh-intel suggested that we add some spec callback to our `PodConfig` structure. I think it's the right approach but I'd like to extend it a bit further:

1. In virtcontainers, define a `spec.go:SpecHandler` interface. For example:

```Go
type SpecHandler interface {
        PodStatus(pod Pod) interface{}
        ContainerStatus(pod Pod, container Container) interface{}
}
```

2. Add a `SpecHandler` field to `PodConfig`
3. Add a `SpecStatus interface{}` field to `ContainerStatus`
4. When calling `StatusContainer()`, do something along those lines:
```Go
func StatusContainer(podID, containerID string) (ContainerStatus, error) {
[...]
if pod.SpecHandler != nil {
   contStatus.SpecStatus = pod.SpecHandler.ContainerStatus(pod, container)
}

```
5. In `pkg/oci/utils.go`, implement an OCI specific virtcontainers `SpecHandler`. The `OCIStatus` structure could also be kept in `pkg/oci/utils.go`
6. From any OCI runtime implementation, optionally add `pkg/oci/utils.go:OCISpecHandler` to the `PodConfig` it creates.

jodh-intel:
@sameo - this sounds good to me. I'll start work on a `SpecHandler` PR based on your design ...

sameo:
After some productive IRC brainstorming with @jodh-intel, we came up to the conclusion that an easier and simpler solution would be to add `Pod` and `Container` Annotations to respectively the `PodStatus` and `ContainerStatus` structures.

sboeuf:
@jodh-intel I think this PR is not relevant anymore because #228 has been merged. Please close it if this is the case.

jodh-intel:
Yep - it is merged now, so closing.

PR #215, opened 2017-04-25
hyperstart: Copy pause binary instead of creating a symbolic link

sboeuf:
Initially we used a hardlink to create the pause binary to the needed location so that our pause container could use it from inside the VM. But hardlink had a limitation regarding cross partitions, and that's the reason we moved to symbolic link.

But symbolic links are not evaluated when the pause container rootfs was mounted, resulting in having the pause binary missing.

The only way to properly deal with those issues is to copy the binary to the expected location.

### DIFF
--- a/hyperstart.go
+++ b/hyperstart.go
@@ -195,7 +195,7 @@ func (h *hyper) buildNetworkInterfacesAndRoutes(pod Pod) ([]hyperstart.NetworkIf
 	return ifaces, routes, nil
 }
 
-func (h *hyper) linkPauseBinary(podID string) error {
+func (h *hyper) copyPauseBinary(podID string) error {
 	pauseDir := filepath.Join(defaultSharedDir, podID, pauseContainerName, rootfsDir)
 
 	if err := os.MkdirAll(pauseDir, dirMode); err != nil {
@@ -204,10 +204,10 @@ func (h *hyper) linkPauseBinary(podID string) error {
 
 	pausePath := filepath.Join(pauseDir, pauseBinName)
 
-	return os.Symlink(h.config.PauseBinPath, pausePath)
+	return fileCopy(h.config.PauseBinPath, pausePath)
 }
 
-func (h *hyper) unlinkPauseBinary(podID string) error {
+func (h *hyper) removePauseBinary(podID string) error {
 	pauseDir := filepath.Join(defaultSharedDir, podID, pauseContainerName)
 
 	return os.RemoveAll(pauseDir)
@@ -392,7 +392,7 @@ func (h *hyper) startPauseContainer(podID string) error {
 		Process: process,
 	}
 
-	if err := h.linkPauseBinary(podID); err != nil {
+	if err := h.copyPauseBinary(podID); err != nil {
 		return err
 	}
 
@@ -454,7 +454,7 @@ func (h *hyper) stopPauseContainer(podID string) error {
 		return err
 	}
 
-	if err := h.unlinkPauseBinary(podID); err != nil {
+	if err := h.removePauseBinary(podID); err != nil {
 		return err
 	}
 
--- a/utils.go
+++ b/utils.go
@@ -1,0 +1,43 @@
+//
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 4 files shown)


utils.go
```diff
@@ -0,0 +1,59 @@
...
+func fileCopy(srcPath, dstPath string) error {
+	if srcPath == "" {
+		return fmt.Errorf("Source path cannot be empty")
+	}
+
+	if dstPath == "" {
+		return fmt.Errorf("Destination path cannot be empty")
+	}
+
+	srcInfo, err := os.Stat(srcPath)
```
jodh-intel: You don't need the `Stat()` call as `ReadFile()` will handle `ENOENT`.

utils.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		return err
+	}
+
+	srcContent, err := ioutil.ReadFile(srcPath)
+	if err != nil {
+		return err
+	}
+
+	if err := os.RemoveAll(dstPath); err != nil {
```
jodh-intel: I'd be tempted to call `RemoveAll()` before `ReadFile()` since if you cannot remove the destination, there isn't much point in reading the source (first).

utils.go
```diff
@@ -0,0 +1,59 @@
...
+		return err
+	}
+
+	dstFile, err := os.OpenFile(dstPath, os.O_WRONLY|os.O_CREATE, srcInfo.Mode())
```
jodh-intel: It would be better (atomic) if you wrote the file to a temporary name first, then called `os.Rename()` on it since that way you can never have a partially-written `dstPath`. Also, since `Rename()` handles an existing file automatically, I'd probably just drop the `RemoveAll()` call.

hyperstart.go
```diff
@@ -204,7 +204,7 @@ func (h *hyper) linkPauseBinary(podID string) error {
 
 	pausePath := filepath.Join(pauseDir, pauseBinName)
 
-	return os.Symlink(h.config.PauseBinPath, pausePath)
+	return fileCopy(h.config.PauseBinPath, pausePath)
```
jodh-intel: The function name should change to reflect the changed behaviour: `copyPauseBinary()` maybe.

jodh-intel: Also, particularly since this function has been problematic in the past, how about introducing a `TestCopyPauseBinary()` function? We could even check copying to a new fs with a tmp mount and also ensure the `src` and `dest` have the expected different inode numbers.

sboeuf: I wanted to create a generic function, and not something specific to the pause binary. Does that make sense ?

sboeuf: I have already added TestFileCopySuccessful(), have you seen it ?

utils.go
sboeuf: Agreed.

utils.go
sboeuf: I need the `Stat()` to get file information, so that I can determine file mode later in the code.

sboeuf:
@jodh-intel I have made some modifications based on your comments, please let me know if that sounds better now ?

utils.go
```diff
@@ -0,0 +1,64 @@
...
+
+package virtcontainers
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+)
+
+func fileCopy(srcPath, dstPath string) error {
```
jodh-intel: A thought occurred - extended attributes and security contexts. This doesn't appear to be performance-critical code so it would honestly be easier and safer (and avoids reinventing the wheel ;) to call out to `cp` (`exec.LookPath()`, *etc*) as that will handle such complexities for us. 

Mr Pike would seem to agree: https://go-review.googlesource.com/c/1591/9/src/io/ioutil/ioutil.go#77.

sboeuf: @jodh-intel I honestly don't know what is the best here. If @sameo agrees with this executing `cp` directly from our go code, then I am fine with this.
Let's just try to get this PR merged by end of this week at least, because it virtcontainers which is currently broken.

sameo: @jodh-intel @sboeuf I'm fine with using exec'ing `cp` directly, it's probably safer.

sboeuf:
@jodh-intel I used exec.Run() to `cp` the file, and I changed the hyperstart functions names, and I have written a basic unit test for this. Does that sound good ?

jodh-intel:
I think it would be useful to add a comment for `cpBinaryName` explaining *why* we call out to the system utility command (mentioning extended attributes and security contexts), but that's a nit, so...

lgtm

sameo:
LGTM

PR #216, opened 2017-04-26
oci: Fix annotation namespacing.

jodh-intel:
The OCI spec states for Annotations that "Keys SHOULD be named using a
reverse domain notation". This is sensible to minimise the chance of
collision (Annotations have a shared namespace).


pkg/oci/utils.go
```diff
@@ -32,8 +32,8 @@ var (
 	// ErrNoLinux is an error for missing Linux sections in the OCI configuration file.
 	ErrNoLinux = errors.New("missing Linux section")
 
-	// ociConfigPathKey is the annotation key to fetch the OCI config.json file path.
-	ociConfigPathKey = "oci/config_path"
+	// ociConfigPathKey is the annotation key to fetch the OCI configuration file path.
+	ociConfigPathKey = "com.github.containers.virtcontainers.pkg.oci.config_path"
```
sboeuf: I think you're talking about a different thing here. The OCI specs defines some annotations but our Annotations field in virtcontainers is not OCI specific. It is intended to provide any runtime implementation, a way to store extra data that are specific to the specification.
I think we should not merge this change.

jodh-intel:
@sboeuf - ah - the commit message needs to be tweaked, but:

https://github.com/containers/virtcontainers/blob/master/pod.go#L285..L286

jodh-intel:
Branch updated.

sboeuf:
@jodh-intel Sounds better like this :)
LGTM

PR #218, opened 2017-04-27
qemu: Fix benchmarks issues related to VM root cause

### LINKED ISSUE #199: Qemu/QMP fails sporadically when running benchmarks
In case we run some benchmarks using Qemu hypervisor such as:
```
go test -bench=CreateStartStopDeletePodQemuHypervisorHyperstartAgentNetworkNoop -benchtime=60s
```
We get the following error:
```
WARN[0022] Unable to connect to unix socket (/tmp/virtcontainers-tmp-009901593/virtcontainers/pods/run/7f49d00d-1995-4156-8c79-5f5ab24ce138/monitor.sock): dial unix /tmp/virtcontainers-tmp-009901593/virtcontainers/pods/run/7f49d00d-1995-4156-8c79-5f5ab24ce138/monitor.sock: connect: no such file or directory 
ERRO[0022] Failed to connect to QEMU instance dial unix /tmp/virtcontainers-tmp-009901593/virtcontainers/pods/run/7f49d00d-1995-4156-8c79-5f5ab24ce138/monitor.sock: connect: no such file or directory 
--- FAIL: BenchmarkCreateStartStopDeletePodQemuHypervisorHyperstartAgentNetworkNoop-4
        api_test.go:1393: Could not create and start pod: Did not receive the pod started notification
FAIL
```

I think this is caused by Qemu stopping very soon for unknown reasons. I have added a timeout before we call QMPStart() because I thought it was the monitor.sock creation that was taking some time, but obviously this is the opposite.
This issue will be solved by finding the reason why Qemu is stopping unexpectedly.

sboeuf:
After we sent the QUIT command through QMP socket to the VM, we were just returning, assuming the VM was stopped. But in some cases, it can take a little bit more time for the VM to stop, causing issues with our benchmarks. Indeed, running the same VM with the same UUID was conflicting and the QMP sockets were not created because the previous VM with the same UUID was not stopped. This patch waits for a go channel indicating QMP socket has been disconnected, meaning the "real" end of the VM.

Fixes #199.

### DIFF
--- a/qemu.go
+++ b/qemu.go
@@ -25,6 +25,7 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+	"time"
 
 	ciaoQemu "github.com/01org/ciao/qemu"
 	"github.com/01org/ciao/ssntp/uuid"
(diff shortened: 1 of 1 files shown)


sameo:
LGTM

PR #221, opened 2017-04-28
Add QEMU Q35 machine type support

sameo:
`Q35` is a more standard and upstream QEMU machine type that is a reasonably good alternative to `pc-lite`.
We now support it through virtcontainers, although the default machine type is still `pc-lite`.


sameo:
cc @jodh-intel

jodh-intel:
lgtm

PR #223, opened 2017-05-01
network: Fix benchmarks sporadic issues with netns

### LINKED ISSUE #219: cni: Benchmarks sporadically fail because we are not ending in the expected netns
We should have benchmarks consistently passing, but instead they fail sporadically because of CNI errors:
```
BenchmarkCreateStartStopDeletePodQemuHypervisorHyperstartAgentNetworkCNI-4    	--- FAIL: BenchmarkCreateStartStopDeletePodQemuHypervisorHyperstartAgentNetworkCNI-4
	api_test.go:1405: Could not delete pod: exit status 2: iptables v1.4.21: Couldn't load target `CNI-2e3a43835ef481fda21efc6f':No such file or directory
		
		Try `iptables -h' or 'iptables --help' for more information.
BenchmarkCreateStartStopDeletePodQemuHypervisorNoopAgentNetworkCNI-4          	--- FAIL: BenchmarkCreateStartStopDeletePodQemuHypervisorNoopAgentNetworkCNI-4
	api_test.go:1405: Could not delete pod: exit status 2: iptables v1.4.21: Couldn't load target `CNI-b0b4f04a417de990fd3ddeed':No such file or directory
		
		Try `iptables -h' or 'iptables --help' for more information.
```

The root cause is about our code being in the wrong netns and unable to see what it is supposed to delete. Still need to figure out the reason for this mixing between netns.

sboeuf:
This commit gets rid of doNetNS() function as it is too complex and it introduces some sporadic issues related to the use of go routines and network namespaces at the same time. The cause of those issues
is about some processes/threads changing during the runtime, causing some confusion with the network namespaces used.

The only way is to really make sure everything is controlled by our library, meaning we make all of our network implementations rely on the same "safe" function safeDoNetNS(). This patch also improves this
function by calling into runtime.LockOSThread(). This implies that no matter who is the caller of this function, it will end up in the expected network namespace with the guarantee the thread won't change
during the execution.

This is solving all sporadic issues related to the usage of network namespaces that have been found with the benchmarks.

Fixes #219

### DIFF
--- a/network.go
+++ b/network.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"runtime"
 
 	"github.com/01org/ciao/ssntp/uuid"
 	"github.com/containernetworking/cni/pkg/ns"
@@ -302,24 +303,13 @@ func setNetNS(netNSPath string) error {
 	return n.Set()
 }
 
-// doNetNS makes some subsequent calls to a go routine and LockOSThread
-// in order to protect the current process from other thread switching
-// to different netns. Unless we have to make sure the thread ID has to
-// stay the same, this function should be used by default.
+// doNetNS is free from any call to a go routine, and it calls
+// into runtime.LockOSThread(), meaning it won't be executed in a
+// different thread than the one expected by the caller.
 func doNetNS(netNSPath string, cb func(ns.NetNS) error) error {
-	n, err := ns.GetNS(netNSPath)
-	if err != nil {
-		return err
-	}
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
 
-	return n.Do(cb)
-}
-
-// safeDoNetNS is free from any call to a go routine, meaning it will
-// not be executed in a different thread than the one expected by the
-// caller. This is used in case of CNM network, because we need to
-// make sure the process switched to the given netns has PID == TID.
-func safeDoNetNS(netNSPath string, cb func(ns.NetNS) error) error {
 	currentNS, err := ns.GetCurrentNS()
 	if err != nil {
 		return err
--- a/cnm.go
+++ b/cnm.go
@@ -181,7 +181,7 @@ func (n *cnm) run(networkNSPath string, cb func() error) error {
 		return fmt.Errorf("networkNSPath cannot be empty")
 	}
 
-	return safeDoNetNS(networkNSPath, func(_ ns.NetNS) error {
+	return doNetNS(networkNSPath, func(_ ns.NetNS) error {
 		return cb()
 	})
 }


jodh-intel:
lgtm

dlespiau:
> LockOSThread wires the calling goroutine to its current operating system thread. Until the calling goroutine exits or calls UnlockOSThread, it will always execute in that thread, and no other goroutine can.

I haven't checked if that function is called by a short-lived goroutine, but if that isn't the case, it should be a good idea to call `UnlockOSThread` so we don't end up with 1 goroutine take a full thread and not sharing it with any other.

network.go
```diff
@@ -302,24 +303,12 @@ func setNetNS(netNSPath string) error {
...
-	}
-
-	return n.Do(cb)
-}
-
-// safeDoNetNS is free from any call to a go routine, meaning it will
-// not be executed in a different thread than the one expected by the
-// caller. This is used in case of CNM network, because we need to
-// make sure the process switched to the given netns has PID == TID.
+// safeDoNetNS is free from any call to a go routine, and it calls
```
sameo: Could we call that one `doNetNS` ? `safeDoNetNS` implies there's an unsafe one.

sboeuf: Oh yes you're right, this will make more sense !

sboeuf:
@dlespiau I agree and I will add a defer call to the function. But just want to let you know that we can do this because in the CreatePod() function, the PrestartHook() function is the first one where we can expect safeDoNetNS() to be called, meaning that we already achieved our goal to keep the PID == TID. Look at api.go, I had added an init() function calling into LockOSThread(). The reason for this is about keeping the consumer of the library with its process ID the same than the thread ID until we get into the hook. Indeed, dockerd inspects only /proc/PID/ns/net, meaning we cannot have a different thread ID.
I will submit a second patch to this PR, in order to Unlock all threads for every APIs function, and one Unlock right after we don't need it anymore in the case of CreatePod().

dlespiau:
This is definitely a bit fiddly, it also means the consumer of the library can't call API functions in a goroutine (it could be scheduled in a different thread than the main thread). They need to call API functions from the main thread (probably worth documenting).

I'm not sure I follow the reason why your patch work. runtime.LockOSThread will lock the calling goroutine on its *current* thread, so I don't see any guarantee that the calling goroutine will be bound to the main thread unless it's already running onto the main thread.

api.go
```diff
@@ -64,6 +64,9 @@ func CreatePod(podConfig PodConfig) (*Pod, error) {
 		return nil, err
 	}
 
+	// Unlock the thread here after the network hook was run.
+	runtime.UnlockOSThread()
```
dlespiau: It's a bit awkward to call that from that from every other entry points. Can't we try to call it when we know we're finished with the namespace setup?

sboeuf: This call is about the locking function from the init() function.

sboeuf:
@dlespiau Forcing the lockOSThread() inside the init() is the way we force the main thread to have the same ID than the process. When the runtime import virtcontainers library, init() is executed, making sure the thread is locked.
I agree this is not nice... But dockerd forces us to do that because of the wrong implementation...

api.go
dlespiau: If we know that we don't need the lockOSThread after a certain point, we should call unlock() at that point, not in every single entry point?

dlespiau:
I get the init() point, what I don't get is why putting a LockOSThread thread inside the function works. This depends on the thread the function is being called from. if the function is called form a goroutine with tid != pid, LockOSThread locks that function on to that thread, not the main thread.

dlespiau:
After a brief discussion on IRC, lgtm for the first patch.

The second patch though looks a bit ugly. I'd rather not have it in that form if possible.

mcastelino:
@sboeuf go 1.10 may have a fix for this in the longer term https://github.com/golang/go/issues/20676

PR #224, opened 2017-05-01
Fix readme

amshinde:


jodh-intel:
lgtm

dlespiau:
The semaphoreci test is failing but still young. No way a change to README could have caused that test failure. Merging.

PR #226, opened 2017-05-02
vendor: Update vendoring with latest dep tool version

sboeuf:
Now that dep tool includes a "prune" option, we want to use it to reduce our vendoring. This means we have to initialiaze and update our previous vendoring.


sboeuf:
@sameo @dlespiau if this patch is fine with you, please merge it.

sameo:
LGTM

PR #228, opened 2017-05-04
Add annotations to statuses

jodh-intel:


api.go
```diff
@@ -638,10 +639,11 @@ func StatusContainer(podID, containerID string) (ContainerStatus, error) {
...
 			return ContainerStatus{
-				ID:     container.id,
-				State:  container.state,
-				PID:    container.process.Pid,
-				RootFs: container.config.RootFs,
+				ID:          container.id,
+				State:       container.state,
+				PID:         container.process.Pid,
+				RootFs:      container.config.RootFs,
+				Annotations: container.pod.config.Annotations,
```
sameo: This will work in the 1container/pod OCI context, but is not generic enough.

The container status annotations map should carry the container config annotations, not the pod ones, in order to carry container specific annotations in a context where we have more than 1 container per pod.

That means adding an `Annotations` field to the `ContainerConfig` structure and copying it to `ContainerStatus`

pkg/oci/utils.go
```diff
@@ -220,11 +241,12 @@ func StatusToOCIState(status vc.PodStatus) (spec.State, error) {
...
-		Version: spec.Version,
-		ID:      status.ID,
-		Status:  stateToOCIState(status.ContainersStatus[0].State),
-		Pid:     status.ContainersStatus[0].PID,
-		Bundle:  status.ContainersStatus[0].RootFs,
+		Version:     spec.Version,
+		ID:          status.ID,
+		Status:      stateToOCIState(status.ContainersStatus[0].State),
+		Pid:         status.ContainersStatus[0].PID,
+		Bundle:      status.Annotations[ociBundlePathKey],
```
sameo: This then would become `status.ContainerStatus[0].Annotations[ociBundlePathKey]`

pkg/oci/utils.go
```diff
@@ -220,11 +241,12 @@ func StatusToOCIState(status vc.PodStatus) (spec.State, error) {
...
+		Annotations: status.Annotations,
```
sameo: and here `status.ContainerStatus[0].Annotations`

jodh-intel:
Hi @sameo - branch ready for re-review.

pkg/oci/utils.go
```diff
@@ -140,10 +143,34 @@ func networkConfig(ocispec spec.Spec) (vc.NetworkConfig, error) {
...
+		}
+
+		finalPodAnnotations[k] = v
+	}
+
+	// copy container annotations
+	for k, v := range containerAnnotations {
+		finalContainerAnnotations[k] = v
+	}
+
```
sameo: Shouldn't this simply be `finalContainerAnnotations = containerAnnotations` ?

pkg/oci/utils.go
```diff
@@ -140,10 +143,34 @@ func networkConfig(ocispec spec.Spec) (vc.NetworkConfig, error) {
 
 // PodConfig converts an OCI compatible runtime configuration file
 // to a virtcontainers pod configuration structure.
-func PodConfig(runtime RuntimeConfig, bundlePath, cid, console string) (*vc.PodConfig, *spec.Spec, error) {
+func PodConfig(runtime RuntimeConfig, bundlePath, cid, console string, podAnnotations, containerAnnotations map[string]string) (*vc.PodConfig, *spec.Spec, error) {
```
sameo: So here you want the caller (e.g. our 3.0 runtime implementation) to be responsible for creating both the pod and container annotations ?

I may be missing something, but the idea of this routine is to transform OCI specific data into virtcontainers pointers. So I wonder what is the rationale behind adding pod and container annotations as arguments here. Since this is being done in the `cc-runtime list` implementation context, there may be some annotations that you'd want to pass from the runtime through a pod or container config. In that case, this should be explicitly done from the runtime _after_ calling this routine by setting `Annotations` directly (or through some setter routine).

pkg/oci/utils.go
```diff
@@ -140,10 +143,34 @@ func networkConfig(ocispec spec.Spec) (vc.NetworkConfig, error) {
...
+	for k, v := range containerAnnotations {
+		finalContainerAnnotations[k] = v
+	}
+
 	configPath := filepath.Join(bundlePath, "config.json")
 	ociLog.Debugf("converting %s", configPath)
 
+	// supplement pod annotations
+	finalPodAnnotations[ociConfigPathKey] = configPath
+	finalPodAnnotations[ociBundlePathKey] = bundlePath
```
sameo: Ok, I _believe_ that's where the confusion is coming from: The current implementation is wrong, both the bundle and config file paths are container specific artifacts and should be carried through the single container annotations map. Not through pod ones.

At the moment, this routine should return a PodConfig with an empty pod annotations map.

pkg/oci/utils.go
```diff
@@ -176,9 +203,10 @@ func PodConfig(runtime RuntimeConfig, bundlePath, cid, console string) (*vc.PodC
 	}
 
 	containerConfig := vc.ContainerConfig{
-		ID:     cid,
-		RootFs: rootfs,
-		Cmd:    cmd,
+		ID:          cid,
+		RootFs:      rootfs,
+		Cmd:         cmd,
+		Annotations: finalContainerAnnotations,
```
sameo: And this should container the bundle and config paths.

pkg/oci/utils.go
sameo: And this comment is still valid.

jodh-intel:
Hi @sameo - thanks for reviewing. Branch updated.

sameo:
LGTM

PR #230, opened 2017-05-04
networking: Turn off multicast snooping on integration bridge

### LINKED ISSUE #229: Turn off multicast snooping on integration bridge
This has caused issues related to kernel panics due to a bug in the multicast timer expiry.
See: https://github.com/01org/cc-oci-runtime/issues/581
We would like to turn off as many attributes as possible in general to dumb down the bridge.

amshinde:
Multicast snooping may be turned on the bridge depending on the
distro kernel config. Disable this explicitly to avoid kernel issues
related to multicast timer expiration.

Fixes #229

### DIFF
--- a/network.go
+++ b/network.go
@@ -132,7 +132,8 @@ func createLink(netHandle *netlink.Handle, name string, expectedLink netlink.Lin
 	switch expectedLink.Type() {
 	case (&netlink.Bridge{}).Type():
 		newLink = &netlink.Bridge{
-			LinkAttrs: netlink.LinkAttrs{Name: name},
+			LinkAttrs:         netlink.LinkAttrs{Name: name},
+			MulticastSnooping: expectedLink.(*netlink.Bridge).MulticastSnooping,
 		}
 	case (&netlink.Tuntap{}).Type():
 		newLink = &netlink.Tuntap{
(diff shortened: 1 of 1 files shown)


sameo:
cc @mcastelino

sameo:
LGTM

sboeuf:
@amshinde I am totally fine with this change. But wouldn't be clearer to use this function from the netlink package instead:
```
func (h *Handle) BridgeSetMcastSnoop(link Link, on bool) error
```
Link: https://godoc.org/github.com/vishvananda/netlink#Handle.BridgeSetMcastSnoop

amshinde:
@sboeuf Using that function is just a matter of convenience. Plus the createLink function is a generic function and it should be able to handle both true/false values for multicast snooping for bridges.

sboeuf:
Well I think it will help to understand the code easily. When you look at the code, you can see the sequence:
```
// Create the bridge
bridgeLink, err := createLink(netHandle, netPair.Name, &netlink.Bridge{})
if err != nil {
	return fmt.Errorf("Could not create bridge: %s", err)
}

// Disable multicast snooping
if err := netHandle.BridgeSetMcastSnoop(bridgeLink, false); err != nil {
	return fmt.Errorf("Could not disable the multicast snooping on the bridge: %s", err)
}
```

amshinde:
I'd rather have the createLink function support all possible attributes present on a link/bridge while creating the link, rather than have extra calls to set attributes later on.

sboeuf:
Ok fine.
LGTM

PR #232, opened 2017-05-05
network: Fix wrong MAC address configuration

sboeuf:
We were not able to ping the network plugin bridge from inside the container, meaning something was not properly configured. Indeed, we were assigning the same MAC address to the veth inside the network namespace and to the network interface inside the VM.

The right thing to do is to retrieve the MAC address of the veth after it has been created by the network plugin, then assign it to the network interface inside the VM, and reset the veth MAC to a random one. That way, we make sure the traffic will be seen by the network bridge on the host (cni0 or docker0 for instance) as if it was coming from the veth pair MAC address. But the reality is that those packets actually come from inside the VM.

### DIFF
--- a/cni.go
+++ b/cni.go
@@ -17,7 +17,6 @@
 package virtcontainers
 
 import (
-	"github.com/containernetworking/cni/pkg/ns"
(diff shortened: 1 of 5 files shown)


sameo:
LGTM

Issue #233, opened 2017-05-05
Add support for additional groups while running a container


amshinde:
Ref : https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go#L89

sboeuf:
We should implement this by changing the Cmd structure from
```go
type Cmd struct {
    ...
    Group string
    ...
}
```
to
```go
type Cmd struct {
    ...
    Groups []string
    ...
}
```

In case our agent support to set only one group, we just need to take `cmd.Groups[0]`.

PR #234, opened 2017-05-05
Add support for additional groups

amshinde:


sboeuf:
@amshinde I would prefer to go with the proposal here: https://github.com/containers/virtcontainers/issues/233
The field AdditionalGroups sounds too OCI specific to me, and we can achieve the same thing by providing a list of groups. The first item of the list being what is considered as Group by OCI.

amshinde:
@sboeuf We need to maintain a clear distinction between user's primary group and his supplementary groups. Having a separate field to store this makes more sense to me and makes it clearer in terms of storing the state of a container. We gain no advantage from changing the group from int to a slice of integers, especially when the supplementary groups may not always be provided. We now have this implied logic that the group at index 0 is the primary group of the user and pass the rest of the groups separately in the "additionalGroups" field expected by hyperstart agent as well.

pod.go
```diff
@@ -224,8 +224,9 @@ type Cmd struct {
 	Envs    []EnvVar
 	WorkDir string
 
-	User  string
-	Group string
+	User             string
+	Group            string
+	AdditionalGroups []string
```
sameo: I am fine with a having a clear distinction between a process primary and supplementary groups as this is a POSIX definition. But then the naming should match POSIX and not OCI:

```
PrimaryGroup        string
SupplementaryGroups []string

**Comment by amshinde** [comment]
@sameo @sboeuf I have renamed additionalGroups to SupplementaryGroups to make it less oci specific.

**Comment by sboeuf** [comment]
@amshinde you forgot to rename Group to PrimaryGroup. You should do that as the first commit and then apply your 2 other patches on top of that.
Those 2 patches are fine.

**Comment by amshinde** [comment]
@sboeuf I was planning to do that in a separate PR as it is technically not part of this functionality. 
How does that sound?

**Comment by sboeuf** [comment]
Sounds good to me !

File: pkg/oci/utils.go [hunk 2]
```diff
@@ -170,6 +170,13 @@ func PodConfig(runtime RuntimeConfig, bundlePath, cid, console string) (*vc.PodC
 		Console:     console,
 	}
 
+	if len(ocispec.Process.User.AdditionalGids) > 0 {
+		cmd.SupplementaryGroups = []string{}
```
sboeuf: you can do simpler here, right ? Just loop over AdditionalGids array. And define cmd.SupplementaryGroups before so that you append only if there is actually some additional groups.

amshinde: The difference is between passing nil vs an empty slice. I just preferred the earlier.

amshinde: I can change this if the code appears simpler

sameo:
lgtm

PR #240, opened 2017-05-09
running containers standalone

devimc:
running containers standalone

### DIFF
--- a/cc_shim.go
+++ b/cc_shim.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"syscall"
 )
 
 type ccShim struct{}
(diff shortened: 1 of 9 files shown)


sboeuf:
@devimc @dlespiau @sameo According to the discussion on IRC, I think this is not the way we want to implement this.
Only "run" command expects to wait for the container process to terminate (from runc definition), and because we don't have a run implementation right now, it uses create + start by default. But we have to implement a new run.go file corresponding to the case of "run" command. There is no need to make a difference between standalone vs Docker case, because run is always expected to block until the container process exits.
We have to implement this run.go the same way we have implemented create.go + start.go, and wait for the shim PID provided through virtcontainers API.
I don't think we need any modification for virtcontainers at this point.

PR #242, opened 2017-05-10
api: Make StatusPod() call StatusContainer().

jodh-intel:
StatusPod() was constructing ContainerStatus objects, which is the job
of StatusContainer().

Previously, StatusContainer() was returning an array of
ContainerStatus's *without* the required container Annotations.

Updated TestStatusPodSuccessful() to test full PodStatus returned by
StatusPod().


sameo:
LGTM

PR #250, opened 2017-05-11
api: Add HypervisorConfig to PodStatus.

jodh-intel:
Added hypervisor configuration to PodStatus. The latter already held the
hypervisor type but adding the configuration will allow runtimes to
implement richer "list" and "state" commands which include hypervisor
configuration details.


jodh-intel:
Specifically the hypervisor config is needed for `runtime state` and `runtime list --all`. The latter is an extension (not defined by the OCI specification) but is already available for https://github.com/01org/clr-oci-runtime.

devimc:
lgtm

sboeuf:
LGTM

PR #255, opened 2017-05-15
Improve build and ci scripts

sboeuf:
In order to reduce the amount of commands needed inside semaphore CI settings, those scripts factorize all those commands and provide an easy way to configure semaphore CI.
Also, the makefile is now able to handle more things, meaning that we can simplify .travis.yml.
At last, we have added a mock package in order to retrieve specific binary mock paths that can be now provided when starting `go test` for more flexibility.
Globally, this PR is a way to improve the CI on virtcontainers, making it more modular.


ci/semaphore-jobs.sh
```diff
@@ -0,0 +1,32 @@
...
+#
+
+#!/bin/bash
+
+set -e
+
+go_packages=$(go list ./... | grep -v vendor)
+
+echo -e "go test $go_packages -cover"
+sudo -E go test $go_packages -cover
```
jodh-intel: Since your log statements are identical to the commands being run, I'd go for one of these two approaches:

- Use `eval`
  ```
  cmd="sudo go test $go_packages -cover"
  echo "$cmd"
  eval $cmd
  ```
- Use `set -x`
  Remove all the `echo`'s and add `set -x` at the top of the script.

sboeuf: Thanks for the tip, I have replaced with `set -x`.

jodh-intel:
Could you rename those files too - there is nothing obviously semaphore-specific in them, so they should just be named `ci-setup.sh` or similar, as we do for `cc-oci-runtime`:

- https://github.com/01org/cc-oci-runtime/tree/master/.ci/

That gives us flexibility to run them under multiple CI systems and/or switch CI systems without confusion.

sboeuf:
@jodh-intel I have reorganized the scripts so that they comply with what you expect.

.ci/ci-setup.sh
```diff
@@ -0,0 +1,78 @@
...
+
+echo "Install linux-container kernel"
+chronic sudo apt-get install -y --force-yes linux-container
+
+echo "Install qemu-lite binary"
+chronic sudo apt-get install -y --force-yes qemu-lite
+
+echo "Download clear containers image"
+latest_version=$(curl -sL https://download.clearlinux.org/latest)
+chronic curl -LO "https://download.clearlinux.org/current/clear-${latest_version}-containers.img.xz"
```
jodh-intel: This could do with a variable as you're using that URL a lot. Also, how about building up the URL as you go? The checksum file is just `${file}-SHA512SUMS` for example?

.ci/ci-setup.sh
```diff
@@ -0,0 +1,78 @@
...
+go get github.com/fzipp/gocyclo
+go get github.com/gordonklaus/ineffassign
+go get github.com/golang/lint/golint
+go get github.com/client9/misspell/cmd/misspell
+
+echo "Setup virtcontainers environment"
+chronic sudo -E bash utils/virtcontainers-setup.sh
+
+echo "Check and build virtcontainers"
+go_packages=$(go list ./... | grep -v vendor)
```
jodh-intel: You might want to do what we did with cc-oci-runtime and introduce a `ci-common.sh` for things like this and source it from the other scripts as this is the 2nd time this variable has been defined (with the same value).

utils/virtcontainers-setup.sh
```diff
@@ -16,4 +16,62 @@
 
 #!/bin/bash
 
-bash $GOPATH/src/github.com/containers/virtcontainers/ci/ci-setup.sh
+if [[ $EUID -ne 0 ]]; then
+	echo "This script must be run as root"
+	exit 1
+fi
+
+echo "Create temporary build directory /tmp/virtcontainers/build"
```
jodh-intel: There is a *lot* or repetition in this file. Could you add some variables to clean it up a bit? This sort of thing:

```
tmpdir=$(mktemp)
echo "blah blah $tmpdir"
rm -rf "$tmpdir"
```

Note the unique temporary directory. Can we switch to that (safer). If not, I think this hard-coded path deserves a comment, particularly as it would be useful to be able to run these scripts on a test system.

.ci/ci-setup.sh
sboeuf: done

.ci/ci-setup.sh
sboeuf: done

utils/virtcontainers-setup.sh
sboeuf: done

jodh-intel:
There are still 2 refs to download.clearlinux.org (could do with 1 more variable to ease maintenance), but with or without that further change:

lgtm.

jodh-intel:
@sboeuf - Semaphore seems unhappy.

sboeuf:
@jodh-intel I made the changes you suggested. Could you remove the request for changes because you gave me LGTM but you haven't discarded the "Changes requested".

sboeuf:
@sameo please let me know what you think about this PR.

sameo:
LGTM

PR #256, opened 2017-05-17
pod: add function to set or add annotations

devimc:
SetOrAddAnnotation is a function to set or add
annotations to a pod

### DIFF
--- a/pod.go
+++ b/pod.go
@@ -403,6 +403,11 @@ func (p *Pod) SetAnnotations(annotations map[string]string) error {
 	return nil
 }
 
+// GetAnnotations returns pod's annotations
+func (p *Pod) GetAnnotations() map[string]string {
+	return p.config.Annotations
+}
+
 // URL returns the pod URL for any runtime to connect to the proxy.
 func (p *Pod) URL() string {
 	return p.state.URL
--- a/pod.go
+++ b/pod.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 )
@@ -372,6 +373,8 @@ type Pod struct {
 	state State
 
 	lockFile *os.File
+
+	annotationsLock *sync.RWMutex
 }
 
 // ID returns the pod identifier string.
@@ -391,6 +394,9 @@ func (p *Pod) Annotations(key string) (string, error) {
 
 // SetAnnotations sets or adds an annotations
 func (p *Pod) SetAnnotations(annotations map[string]string) error {
+	p.annotationsLock.Lock()
+	defer p.annotationsLock.Unlock()
+
 	for k, v := range annotations {
 		p.config.Annotations[k] = v
 	}
@@ -405,6 +411,9 @@ func (p *Pod) SetAnnotations(annotations map[string]string) error {
 
 // GetAnnotations returns pod's annotations
 func (p *Pod) GetAnnotations() map[string]string {
+	p.annotationsLock.RLock()
+	defer p.annotationsLock.RUnlock()
+
 	return p.config.Annotations
 }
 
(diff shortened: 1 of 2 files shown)


pod.go
```diff
@@ -536,7 +541,6 @@ func fetchPod(podID string) (*Pod, error) {
 	if err != nil {
 		return nil, err
 	}
-
```
sboeuf: No need to remove this blank line

pod.go
```diff
@@ -389,6 +389,11 @@ func (p *Pod) Annotations(key string) (string, error) {
 	return value, nil
 }
 
+// SetOrAddAnnotation sets or adds an annotation
+func (p *Pod) SetOrAddAnnotation(key, value string) {
```
sboeuf: I would prefer `SetAnnotation()` instead of `SetOrAddAnnotation`, because the way you have implemented this, there is no way to know if it was added or replaced. But keep the comment like this because it explains what is done.

pod.go
```diff
@@ -389,6 +389,11 @@ func (p *Pod) Annotations(key string) (string, error) {
...
+	p.config.Annotations[key] = value
```
sboeuf: I think you should save the PodConfig locally after you added a new Annotation. You can refer to this: https://github.com/containers/virtcontainers/blob/master/api.go#L424-L429
Indeed, depending on where you call this function, you may lose this Annotation if the config is not stored afterwards.
But this will cost a bit of performances because we will store every time we add a new annotation. I would suggest a second function in case you want to add/set several Annotations:
```go
func (p *Pod) SetAnnotations(annotationList map[string]string)
``` 
@sameo @jodh-intel what do you think ?

pod.go
jodh-intel: Also, how about making this function return the previous value for `key` if it was already set. That would require 2 return values though:

```
func (p *Pod) SetOrAddAnnotation(key, value string) (string, bool) {}
```

The `bool` being set if there was a previous value (as returning `""` would be ambiguous [either the key was not set, or it was set to `""`]).

pod.go
jodh-intel: @sboeuf - agreed on both points.

@devimc - I think it makes sense to have an API to set annotations (much safer than simply altering the state of the objects directly), but do you have a particular use-case in mind by adding this (if so, that might help us all to hone in on the best design)?

pod.go
jodh-intel: Of course, once we've reached a consensus on the API, it will need some new tests :)

devimc:
remove ```do-no-merge``` label please

pod.go
```diff
@@ -389,6 +389,20 @@ func (p *Pod) Annotations(key string) (string, error) {
 	return value, nil
 }
 
+// SetAnnotations sets or adds an annotations
+func (p *Pod) SetAnnotations(annotations map[string]string) error {
+	for k, v := range annotations {
+		p.config.Annotations[k] = v
```
sameo: We need to lock the `Annotations` map, with a RW lock. This could be done as the next patch for this PR.

sameo: Once we lock `Annotations`, we also need to provide a `GetAnnotation()` getter and unexport `Annotations`.

devimc:
@sameo  @sboeuf fixed

pod.go
```diff
@@ -36,6 +37,9 @@ const controlSocket = "ctrl.sock"
 // to understand if the VM is still alive or not.
 const monitorSocket = "monitor.sock"
 
+// annotationsLock is a RWMutex used to lock Annotations map
+var annotationsLock sync.RWMutex
```
sameo: We want to have a lock per pod, not a global one.

pod.go
```diff
@@ -391,6 +395,9 @@ func (p *Pod) Annotations(key string) (string, error) {
 
 // SetAnnotations sets or adds an annotations
 func (p *Pod) SetAnnotations(annotations map[string]string) error {
+	annotationsLock.Lock()
```
sboeuf: We want this lock to be part of the pod handler. In case you have several pods, you don't want to wait the lock for another pod's annotations.

devimc:
done, I just added ```annotationsLock``` to ```Pod struct```

sboeuf:
Sounds good to me :)

sboeuf:
LGTM

sameo:
LGTM

PR #259, opened 2017-05-18
vendor: Update runtime-spec package for compatibility with CRI-O

sboeuf:
This PR updates the vendored version of the runtime-spec to v1.0.0-rc5. In order to support both Docker and CRI-O (using different runtime-spec versions), it also introduces a new compatibility structure inheriting from the one defined by runtime-spec.


pkg/oci/utils.go
```diff
@@ -39,6 +39,16 @@ var (
 	BundlePathKey = "com.github.containers.virtcontainers.pkg.oci.bundle_path"
 )
 
+type CompatOCIProcess struct {
```
jodh-intel: I don't understand why this is being introduced on a vendoring change? It isn't actively being used so can't (shouldn't?) this be handled on a follow-up PR (along with a new test or two)?

sameo: @jodh-intel I think @sboeuf is trying to implement a structure that would allow us to be both rc4 and rc5 compatible.
Those compat structures replace the `spec.Process` ones in the below calls, they are used.

jodh-intel: Right, after some digging I found it.

@sboeuf - it would have been helpful to reference this: https://github.com/opencontainers/runtime-spec/pull/675 and/or https://github.com/opencontainers/runtime-spec/commit/37391fb66ffc7f2bc87c1b85349b68b561c53627 explicitly.

sameo: Agreed. In particular, a mention to opencontainers/runtime-spec@37391fb in a comment would be helpful.
@sboeuf, can you quickly do that to make things a little clearer ?

sboeuf:
@sameo @jodh-intel I have added a short explanation as a comment in each Compat structure. Let me know if that's better !

sameo:
@sboeuf One more question: Why do you want to keep the rc4 compatibility ?

sboeuf:
@sameo Because the config.json given by Docker is relying on this.

sameo:
Makes sense. Docker is based on rc4.

sameo:
LGTM

PR #262, opened 2017-05-19
rootfs: Support mounting the rootfs as readonly

amshinde:
This adds support for the oci specification to mount
the rootfs as readonly.

### DIFF
--- a/syscall_test.go
+++ b/syscall_test.go
@@ -28,7 +28,7 @@ func TestBindMountInvalidSourceSymlink(t *testing.T) {
 	source := filepath.Join(testDir, "fooFile")
 	os.Remove(source)
 
-	err := bindMount(source, "")
+	err := bindMount(source, "", false)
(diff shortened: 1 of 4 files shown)


sameo:
LGTM

syscall_test.go
```diff
@@ -77,14 +77,50 @@ func TestBindMountSuccessful(t *testing.T) {
...
+	}
+
+	// should not be able to create file in read-only mount
+	destFile := filepath.Join(dest, "foo")
+	_, err = os.OpenFile(destFile, os.O_CREATE, mountPerm)
+	if err == nil {
+		t.Fatal(err)
+	}
+
+	syscall.Unmount(dest, 0)
```
sboeuf: Should be called through a `defer` call after we do the `bindMount()`. We won't unmount in case the call to `os.OpenFile()` fails.

PR #266, opened 2017-05-22
oci: Pass the read-only config from the oci spec file

amshinde:
Now that read-only rootfs support has been added pass this to
the container config.
Should have been part of earlier PR, missed adding this.


sboeuf:
LGTM

PR #268, opened 2017-05-23
Add CRI-O support for the runtime

sboeuf:
This PR provides the support to be able to implement CRI-O into the runtime.


pkg/oci/utils.go
```diff
@@ -37,6 +37,34 @@ var (
 
 	// BundlePathKey is the annotation key to fetch the OCI configuration file path.
 	BundlePathKey = "com.github.containers.virtcontainers.pkg.oci.bundle_path"
+
+	// ContainerTypeKey is the annotation key to fetch container type.
+	ContainerTypeKey = "com.github.containers.virtcontainers.pkg.oci.container_type"
+
+	// CRIOContainerTypeKey is the annotation key to fetch CRIO container type.
+	CRIOContainerTypeKey = "crio/container_type"
```
sboeuf: This is defined by CRI-O, not us.... Take a look here: https://github.com/kubernetes-incubator/cri-o/blob/master/server/sandbox_run.go#L263, an here: https://github.com/kubernetes-incubator/cri-o/blob/master/server/container_create.go#L541

pkg/oci/utils.go
```diff
@@ -37,6 +37,34 @@ var (
...
+
+	// CRIOSandboxNameKey is the annotation key to fetch CRIO sandbox/pod name.
+	CRIOSandboxNameKey = "crio/sandbox_name"
+
+	// ContainerTypeContainer defines a regular container.
+	// Keep in sync with CRI-O.
```
sboeuf: Here: https://github.com/kubernetes-incubator/cri-o/blob/master/server/container.go#L10-L13

sboeuf:
@jodh-intel I have already covered those following functions with unit tests:
- `GetContainerType()`
- `GetContainerTypeFromConfig()`
- `IsPod()`
- `PodContainerPodID()`
- `GetOCIConfig()`

`GetContainerInfo()` is a bit more complex and I think it will be done in a PR dedicated to increase the overall coverage, because we have to setup several things.

pkg/oci/utils.go
jodh-intel: Hmm, it would have been nicer if they had created a `const` but how about we add those links into the code? I can see this breaking rather easily (through no fault of ours), but atleast if we add a reference, there is some hope of finding the new values when they change them ;)

pkg/oci/utils.go
```diff
@@ -161,57 +189,111 @@ func networkConfig(ocispec CompatOCISpec) (vc.NetworkConfig, error) {
...
-			ConfigPathKey: configPath,
-			BundlePathKey: bundlePath,
-		},
+	return UnknownContainerType, fmt.Errorf("Could not find container type")
+}
+
+// GetContainerTypeFromConfig returns the type of container based on CRI-O
+// annotations.
+func GetContainerTypeFromConfig(ociSpec CompatOCISpec) (ContainerType, error) {
+	containerType, ok := ociSpec.Annotations[CRIOContainerTypeKey]
```
jodh-intel: How about adding a log call here showing that we failed to find the CRIO key (specify it)?

pkg/oci/utils.go
sboeuf: Yes I agree it would  be even better that they export such values so that we don't need to define them ;)
As you've suggested, I will add those links to the comment !

jodh-intel:
@sboeuf  - Gah - I've been "githubed" (why would I *not* want to see the entire diff?). Seems I forgot to click "load diff" wrt those tests.

pkg/oci/utils.go
sboeuf: done

pkg/oci/utils.go
sboeuf: done

pkg/oci/utils.go
sboeuf: done

sboeuf:
@jodh-intel PR updated.

pkg/oci/utils.go
```diff
@@ -161,57 +196,113 @@ func networkConfig(ocispec CompatOCISpec) (vc.NetworkConfig, error) {
...
-	containerConfig := vc.ContainerConfig{
-		ID:             cid,
-		RootFs:         rootfs,
-		ReadonlyRootfs: ocispec.Spec.Root.Readonly,
-		Cmd:            cmd,
-		Annotations: map[string]string{
-			ConfigPathKey: configPath,
-			BundlePathKey: bundlePath,
-		},
+	return UnknownContainerType, fmt.Errorf("Could not find container type")
```
jodh-intel: Just one more log call here maybe to show that `ContainerTypeKey` couldn't be found.

sboeuf: done

jodh-intel:
lgtm

pkg/oci/utils.go
```diff
@@ -161,57 +196,115 @@ func networkConfig(ocispec CompatOCISpec) (vc.NetworkConfig, error) {
...
-			BundlePathKey: bundlePath,
-		},
+	ociLog.Errorf("Annotations[%s] not found, cannot determine the container type",
+		ContainerTypeKey)
+	return UnknownContainerType, fmt.Errorf("Could not find container type")
+}
+
+// GetContainerTypeFromConfig returns the type of container based on CRI-O
+// annotations.
+func GetContainerTypeFromConfig(ociSpec CompatOCISpec) (ContainerType, error) {
```
sameo: This should be a `CompatOCISpec` method:
```
func (spec CompatOCISpec) Type() (ContainerType, error)
```

Moreover the above routine should not be CRI-O specific and should be able to support other CRI servers. In other words, we should be looking through a set of annotations keys and not only the CRI-O one.

pkg/oci/utils.go
```diff
@@ -161,57 +196,115 @@ func networkConfig(ocispec CompatOCISpec) (vc.NetworkConfig, error) {
...
+		return true
+	}
+
+	return false
+}
+
+// PodContainerPodID determines the pod ID related to an OCI configuration from a specific
+// annotation. This function is expected to be called only when the container
+// type is "PodContainer".
+func PodContainerPodID(annotations map[string]string) (string, error) {
```
sameo: here as well, this would be cleaner as a `CompatOCISpec` method.

pkg/oci/utils.go
```diff
@@ -37,6 +37,41 @@ var (
...
+// ContainerType defines a type of container.
+type ContainerType string
+
+// List different types of containers
+var (
+	RegularContainer     ContainerType = "regular_container"
+	PodContainer         ContainerType = "pod_container"
+	PodSandbox           ContainerType = "pod_sandbox"
+	UnknownContainerType ContainerType = "unknown_container_type"
 )
```
sameo: A container type (sandbox or pod container) is not OCI specific. I would prefer if this definition and all its related method (e.g. getting a container type from a container ID) was part of the virtcontainers API itself.

pkg/oci/utils.go
```diff
@@ -37,6 +37,41 @@ var (
...
+	// https://github.com/kubernetes-incubator/cri-o/blob/v0.3/server/container.go#L11
+	ContainerTypePod = "sandbox"
+)
+
+// ContainerType defines a type of container.
+type ContainerType string
+
+// List different types of containers
+var (
+	RegularContainer     ContainerType = "regular_container"
```
sboeuf: For most cases, that works to have only a pod or a container. But in case of `delete`, I need to be able to make a difference between Docker and CRI-O because checking if the process is stopped here: https://github.com/clearcontainers/runtime/pull/147/files#diff-635f3cde55281671a8db5283d1597a71R86 actually does not work in case of CRI-O. Indeed, conmon is still monitoring the pod container even after we stopped it, meaning that we have a defunct/zombie shim process until we return from delete.

sameo: For the delete case, look at the code if there was no `oci.RegularContainer` definition and use `oci.PodSandbox` instead. It should just work, and if --force is not provided, it will prevent a Docker container or a CRI sandbox from being deleted/removed.

pkg/oci/utils.go
```diff
@@ -161,57 +196,115 @@ func networkConfig(ocispec CompatOCISpec) (vc.NetworkConfig, error) {
...
 
-	cmd.SupplementaryGroups = []string{}
-	for _, gid := range ocispec.Process.User.AdditionalGids {
-		cmd.SupplementaryGroups = append(cmd.SupplementaryGroups, strconv.FormatUint(uint64(gid), 10))
+	return vc.ContainerStatus{}, "", fmt.Errorf("Could not find any pod/container ID %s", containerID)
+}
+
+// GetContainerType determines which type of container matches the annotations
+// table provided.
+func GetContainerType(annotations map[string]string) (ContainerType, error) {
```
sboeuf: There is a main issue about doing this kind of thing, that is we will end up looping over all pods and containers while we already have the annotations (or the ociSpec handler) with us. I mean from a runtime perspective, this will lead us to loop many times over the list of pods and containers while we already have the information in one of our structure.

pkg/oci/utils.go
sboeuf: Yes agreed.

pkg/oci/utils.go
```diff
@@ -161,57 +196,115 @@ func networkConfig(ocispec CompatOCISpec) (vc.NetworkConfig, error) {
...
+	case ContainerTypeContainer:
+		return PodContainer, nil
+	}
+
+	return UnknownContainerType, fmt.Errorf("Unknown container type %s", containerType)
+}
+
+// IsPod determines if the container type can be considered as a pod.
+// We can consider a pod in case we have a PodSandbox or a RegularContainer.
+func IsPod(cType ContainerType) bool {
```
sboeuf: Yes.

pkg/oci/utils.go
```diff
@@ -161,57 +196,115 @@ func networkConfig(ocispec CompatOCISpec) (vc.NetworkConfig, error) {
...
+	if !ok {
+		return "", fmt.Errorf("Could not find pod ID: Unknown Annotations[%s]", CRIOSandboxNameKey)
 	}
 
+	return podID, nil
+}
+
+// PodConfig converts an OCI compatible runtime configuration file
+// to a virtcontainers pod configuration structure.
+func PodConfig(ocispec CompatOCISpec, runtime RuntimeConfig, bundlePath, cid, console string) (vc.PodConfig, error) {
```
sboeuf: If you do that, you will parse several times the config.json... I made those changes specifically to make sure we load the ociSpec only once, and then we can deduce everything from this.

pkg/oci/utils.go
```diff
@@ -244,24 +337,65 @@ func PodConfig(runtime RuntimeConfig, bundlePath, cid, console string) (*vc.PodC
...
 
-// StatusToOCIState translates a virtcontainers pod status into an OCI state.
-func StatusToOCIState(status vc.PodStatus) (spec.State, error) {
-	if len(status.ContainersStatus) != 1 {
-		return spec.State{},
-			fmt.Errorf("ContainerStatus list from PodStatus is wrong, expecting only one container status, got %v",
-				status.ContainersStatus)
+// ContainerConfig converts an OCI compatible runtime configuration
+// file to a virtcontainers container configuration structure.
+func ContainerConfig(ocispec CompatOCISpec, bundlePath, cid, console string) (vc.ContainerConfig, error) {
```
sboeuf: Ditto.

pkg/oci/utils.go
sboeuf: Yes agreed.

pkg/oci/utils.go
sameo: Fair enough. I feel the API is a bit redundant now, but it's for the good cause.

pkg/oci/utils.go
```diff
@@ -37,6 +37,42 @@ var (
...
+	// the container type from annotations in the config.json.
+	CRIContainerTypeKeyList = []string{CRIOContainerTypeKey}
+
+	// CRISandboxNameKeyList lists all the CRI keys that could define
+	// the sandbox name (pod ID) from annotations in the config.json.
+	CRISandboxNameKeyList = []string{CRIOSandboxNameKey}
+
+	// PodIDPrefix is the prefix added to the container ID when the caller
+	// does not define the container type explicitely.
+	PodIDPrefix = "pod_"
```
sameo: Where is that one used ? I can't find it on the runtime PR.

sboeuf: It is used in this pkg. Look at `ContainerConfig()`

pkg/oci/utils.go
```diff
@@ -161,61 +197,94 @@ func networkConfig(ocispec CompatOCISpec) (vc.NetworkConfig, error) {
...
 
-	rootfs := ocispec.Root.Path
-	if !filepath.IsAbs(rootfs) {
-		rootfs = filepath.Join(bundlePath, ocispec.Root.Path)
+	return ocispec, nil
+}
+
+// GetContainerType determines which type of container matches the annotations
+// table provided.
+func GetContainerType(annotations map[string]string) (vc.ContainerType, error) {
```
sboeuf: Sounds good, I'll make this change.

pkg/oci/utils.go
```diff
@@ -161,61 +197,94 @@ func networkConfig(ocispec CompatOCISpec) (vc.NetworkConfig, error) {
...
-		Interactive:  ocispec.Process.Terminal,
-		Console:      console,
+	ociLog.Errorf("Annotations[%s] not found, cannot determine the container type",
+		ContainerTypeKey)
+	return vc.UnknownContainerType, fmt.Errorf("Could not find container type")
+}
+
+// ContainerType returns the type of container and if the container type was
+// found from CRI servers annotations.
+func (spec *CompatOCISpec) ContainerType() (vc.ContainerType, bool, error) {
```
sboeuf: It is here to make the difference between a case where you get a `PodSandbox` because no annotation has been found, from a case where a CRI annotation was set in order to determine the container type. This is helpful for `ContainerConfig()` in order to determine which podID we assign to the pod. In case of CRI, podID == containerID, and in case of Docker podID != containerID. That's the way we can make the difference for the delete case.

sameo: I see. We need to know if both the container and the pod need to be deleted together or not when receiving a `delete` command for a given container id...In the Docker case, we have to delete the pod (and hence the container) whereas in the CRI one we may want to simply delete a container process running inside a Pod (and not the pod itself).

So tell me what I'm missing: Given that a Docker created container will be tagged by virtcontainers as a `PodSandbox` container type, we will always delete the pod in the Docker case. And that's what we want... There's no need to have specific container ID to distinguish the 2 use cases.

sboeuf: Well now I think this is not needed anymore because according to the recent changes into CRI-O getting the shim properly terminated (and not remaining as a zombie process), we don't need to make a difference between those two cases. And it is in my opinion better if we can avoid to make a difference between a CRI-O and a Docker pod sandbox.
I am removing this `bool` return.

sboeuf: Fixed !

pkg/oci/utils.go
sameo: Can you make that change as well ?

sboeuf: I thought we talked in private about this one, and we agreed that we cannot move this as a virtcontainers method because that means virtcontainers would be aware about all annotations. And we decided that we wanted virtcontainers to be annotations agnostic.

sameo: Ah, too many context switches...Yes, you're right.

sameo:
LGTM

sameo:
@mcastelino Could you please look at this patch ? It modifies some of the networking files and thus need your approval.

sboeuf:
@mcastelino please could you review this ?

api.go
```diff
@@ -51,21 +51,21 @@ func CreatePod(podConfig PodConfig) (*Pod, error) {
...
 	// Initialize the network.
-	err = p.network.init(&(p.config.NetworkConfig))
+	netNsPath, netNsCreated, err := p.network.init(p.config.NetworkConfig)
 	if err != nil {
 		return nil, err
 	}
 
 	// Execute prestart hooks inside netns
-	err = p.network.run(p.config.NetworkConfig.NetNSPath, func() error {
+	err = p.network.run(netNsPath, func() error {
```
mcastelino: @sboeuf The pre-start hooks do not need to be executed from within the namespace. At least with docker the pid of the ns is sent as a parameter. So do we need to run within the namespace?

hyperstart.go
```diff
@@ -422,7 +425,7 @@ func (h *hyper) startOneContainer(pod Pod, c Container) error {
 		Process: process,
 	}
 
-	if err := h.bindMountContainerRootfs(pod.id, c.id, c.rootFs, c.config.ReadonlyRootfs); err != nil {
+	if err := h.bindMountContainerRootfs(pod.id, c.id, c.rootFs, false); err != nil {
```
mcastelino: Why do we default to false?

sboeuf: Because hyperstart actually does not support this to be set to true. Docker does not set this to true, but I noticed that CRI-O does that, and we end up with hyperstart failure when it tries to do `mkdir /lib/modules`. This should be fixed inside hyperstart before we can re-enable this.

api.go
sboeuf: Well I agree it does not need to be executed from the netns, but we are passing a PID, and this PID has to be in the expected netns so that Docker can deduce the netns from this PID. And because we're passing the PID of our running thread, we need this thread to be in the netns when we're calling into the hook.

mcastelino: Given the docker pid vs tid and netns issue, I wonder if we should execute the hook using reexec. I am not sure if docker tracks the pid for later use (I think not). That will eliminate the issue we have with docker. It may also simplify our code longer term.

sboeuf: How are you planning to make the hook working if we don't reexec ?
I am pretty sure Docker does not track the pid for later use, it is only using it for netns determination.

mcastelino: I meant we reexec ourself before we exec the hook. That way we ensure tid==pid for the caller of the hook.

sboeuf: I am not sure to understand all consequences here. How does that solve the fact that we have to be in the netns ? We still need to get this PID in the right netns, so that docker can retrieve the right netns when it is looking for /proc/PID/ns/net, right ? Otherwise I am missing something that I don't understand :)

PR #269, opened 2017-05-23
 agent: implement ps command

devimc:
ps command is used to list the processes running inside
the container, this command is called by ```docker top```

partially fixes clearcontainers/runtime#95

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -26,6 +26,22 @@ import (
 // AgentType describes the type of guest agent a Pod should run.
 type AgentType string
 
+// ProcessListOptions contains the options used to list running
+// processes inside the container
+type ProcessListOptions struct {
+	// Format describes the output format to list the running processes.
+	// Formats are unrelated to ps(1) formats, only two formats can be specified:
+	// "json" and "table"
+	Format string
+
+	// Args contains the list of arguments to run ps(1) command.
+	// If Args is empty the agent will use "-ef" as options to ps(1).
+	Args []string
+}
+
+// ProcessList represents the list of running processes inside the container
+type ProcessList []byte
+
 const (
 	// NoopAgentType is the No-Op agent.
 	NoopAgentType AgentType = "noop"
@@ -146,4 +162,7 @@ type agent interface {
 	// container related to a Pod. If all is true, all processes in
 	// the container will be sent the signal.
 	killContainer(pod Pod, c Container, signal syscall.Signal, all bool) error
+
+	// processListContainer will list the processes running inside the container
+	processListContainer(pod Pod, c Container, options ProcessListOptions) (ProcessList, error)
 }
--- a/api.go
+++ b/api.go
@@ -678,3 +678,34 @@ func PausePod(podID string) (VCPod, error) {
 func ResumePod(podID string) (VCPod, error) {
 	return togglePausePod(podID, false)
 }
+
+// ProcessListContainer is the virtcontainers entry point to list
+// processes running inside a container
+func ProcessListContainer(podID, containerID string, options ProcessListOptions) (ProcessList, error) {
+	if podID == "" {
+		return nil, errNeedPodID
+	}
+
+	if containerID == "" {
+		return nil, errNeedContainerID
+	}
+
+	lockFile, err := lockPod(podID)
+	if err != nil {
+		return nil, err
+	}
+	defer unlockPod(lockFile)
+
+	p, err := fetchPod(podID)
+	if err != nil {
+		return nil, err
+	}
+
+	// Fetch the container.
+	c, err := fetchContainer(p, containerID)
+	if err != nil {
+		return nil, err
+	}
+
+	return c.processList(options)
+}
--- a/container.go
+++ b/container.go
@@ -707,6 +707,24 @@ func (c *Container) kill(signal syscall.Signal, all bool) error {
 	return nil
 }
 
+func (c *Container) processList(options ProcessListOptions) (ProcessList, error) {
+	state, err := c.fetchState("ps")
+	if err != nil {
+		return nil, err
+	}
+
+	if state.State != StateRunning {
+		return nil, fmt.Errorf("Container not running, impossible to list processes")
+	}
+
+	if _, _, err := c.pod.proxy.connect(*(c.pod), false); err != nil {
+		return nil, err
+	}
+	defer c.pod.proxy.disconnect()
+
+	return c.pod.agent.processListContainer(*(c.pod), *c, options)
+}
+
 func (c *Container) createShimProcess(token, url string, cmd Cmd) (*Process, error) {
 	if c.pod.state.URL != url {
 		return &Process{}, fmt.Errorf("Pod URL %s and URL from proxy %s MUST be identical", c.pod.state.URL, url)
--- a/hyperstart.go
+++ b/hyperstart.go
@@ -649,3 +649,32 @@ func (h *hyper) killOneContainer(cID string, signal syscall.Signal, all bool) er
 
 	return nil
 }
+
+func (h *hyper) processListContainer(pod Pod, c Container, options ProcessListOptions) (ProcessList, error) {
+	return h.processListOneContainer(pod.id, c.id, options)
+}
+
+func (h *hyper) processListOneContainer(podID, cID string, options ProcessListOptions) (ProcessList, error) {
+	psCmd := hyperstart.PsCommand{
+		Container: cID,
+		Format:    options.Format,
+		PsArgs:    options.Args,
+	}
+
+	proxyCmd := hyperstartProxyCmd{
+		cmd:     hyperstart.PsContainer,
+		message: psCmd,
+	}
+
+	response, err := h.proxy.sendCmd(proxyCmd)
+	if err != nil {
+		return nil, err
+	}
+
+	msg, ok := response.([]byte)
+	if !ok {
+		return nil, fmt.Errorf("failed to get response message from container %s pod %s", cID, podID)
+	}
+
+	return msg, nil
+}
--- a/pkg/hyperstart/hyperstart.go
+++ b/pkg/hyperstart/hyperstart.go
@@ -50,6 +50,7 @@ const (
 	SetupInterface  = "setupinterface"
 	SetupRoute      = "setuproute"
 	RemoveContainer = "removecontainer"
+	PsContainer     = "pscontainer"
 )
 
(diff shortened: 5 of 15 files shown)


api.go
```diff
@@ -698,3 +698,44 @@ func KillContainer(podID, containerID string, signal syscall.Signal, all bool) e
 
 	return nil
 }
+
+// PsContainer is the virtcontainers entry point to list
+// processes running inside a container
+func PsContainer(podID, containerID string, psArgs []string) error {
```
sboeuf: Maybe I am missing something here but you never return anything, right ? How do you plan to get the data back ?

Also I think it could be part of `StatusContainer()` instead of adding a new function to the exported API. @sameo any thoughts on this one ?

jodh-intel:
Related to https://github.com/clearcontainers/runtime/issues/95.

api.go
devimc: probably I have to return a string

sboeuf: I would say that you want to return a new dedicated structure. Something like:
```go
type ProcessInfo struct {
        Name string
        PID string
        PPID string
        ...
}

type PsInfo struct {
        ID string
        ProcessesList []ProcessInfo
}
```

pkg/hyperstart/hyperstart.go
```diff
@@ -73,6 +74,7 @@ var CodeList = map[string]uint32{
 	SetupInterface:  SetupInterfaceCode,
 	SetupRoute:      SetupRouteCode,
 	RemoveContainer: RemoveContainerCode,
+	PsContainer:     PsContainerCode,
```
sameo: Where in hyperstart is this defined ?

devimc: It is not defined, not yet I'm working on that https://github.com/clearcontainers/hyperstart/pull/30

devimc: https://github.com/clearcontainers/hyperstart/pull/30/files#diff-b94e58fcb93efa0dafb0674fda336bacR31

sameo:
Depends on an hyperstart PR

jodh-intel:
Related: https://github.com/clearcontainers/proxy/pull/83

devimc:
@sameo  @sboeuf  @jodh-intel or somebody with superpowers , please remove ```do-not-merge``` tag

agent.go
```diff
@@ -146,4 +146,7 @@ type agent interface {
 	// container related to a Pod. If all is true, all processes in
 	// the container will be sent the signal.
 	killContainer(pod Pod, c Container, signal syscall.Signal, all bool) error
+
+	// psContainer will list the processes running inside the container
+	psContainer(pod Pod, c Container, format string, psArgs []string) ([]byte, error)
```
sboeuf: `processListContainer` seems like a more generic name IMO.
And also I would prefer a function prototype a bit different:
```go
processListContainer(pod Pod, c Container, processListOptions []Param) (ProcessList, error)
```
The reason here is that I would like to get the same processList structure from any agent implementation and get this structure formatted as expected from the runtime implementation.

api.go
```diff
@@ -660,3 +660,34 @@ func PausePod(podID string) (*Pod, error) {
 func ResumePod(podID string) (*Pod, error) {
 	return togglePausePod(podID, false)
 }
+
+// PsContainer is the virtcontainers entry point to list
+// processes running inside a container
+func PsContainer(podID, containerID string, format string, psArgs []string) ([]byte, error) {
```
sboeuf: `s/PsContainer/ProcessListContainer/`
Prototype
```go
ProcessListContainer(podID, containerID string, processListOptions []Param) (ProcessList, error)
```

container.go
```diff
@@ -607,6 +607,24 @@ func (c *Container) kill(signal syscall.Signal, all bool) error {
 	return nil
 }
 
+func (c *Container) ps(format string, psArgs []string) ([]byte, error) {
```
sboeuf: Replace
```go
ps(format string, psArgs []string) ([]byte, error)
```
with
```go
processList(processListOptions []Param) (ProcessList, error)
```

devimc:
@sboeuf changes applied

sboeuf:
@devimc Thanks but you've addressed only half of the comments. Here is what you missed:
> And also I would prefer a function prototype a bit different:
> ```go
> processListContainer(pod Pod, c Container, processListOptions []Param) (ProcessList, error)
> ```
> The reason here is that I would like to get the same processList structure from any agent implementation and get this structure formatted as expected from the runtime implementation.

And
> Replace
> ```go
> ps(format string, psArgs []string) ([]byte, error)
> ```
> with
> ```go
> processList(processListOptions []Param) (ProcessList, error)
> ```

The important point here is that I think it would be better to return a generic structure that we define inside virtcontainers so that it can be reused by any runtime implementation. That means that I would expect to see the runtime doing the conversion between this structure and the expected format.
Does that make sense ?
@sameo any thought on that ?

devimc:
@sboeuf  imo does not make sense to create a struct just for 1 element, I agree with you I'd expect the conversion of types in the runtime, for this reason I defined the type ProcessList as []byte

hyperstart.go
```diff
@@ -634,3 +634,27 @@ func (h *hyper) killOneContainer(cID string, signal syscall.Signal, all bool) er
...
+	}
+
+	proxyCmd := hyperstartProxyCmd{
+		cmd:     hyperstart.PsContainer,
+		message: psCmd,
+	}
+
+	response, err := h.proxy.sendCmd(proxyCmd)
+	if msg, ok := response.([]byte); ok {
+		return msg, err
```
jodh-intel: This would be clearer as follows (currently it is unclear if it's an error or not :):

```
return msg, nil
```

sboeuf: yes I think this should return `return msg, nil` since this is the __right__ case.

jodh-intel: @sboeuf - Once that's done, is this good to merge?

sboeuf: @jodh-intel @devimc Yes after this is cleared up, I think we're good to merge this PR. Please could you implement like this in order to get all this clearer:
```go
response, err := h.proxy.sendCmd(proxyCmd)
if err != nil {
        return nil, err
}
if msg, ok := response.([]byte); !ok {
        return nil, fmt.Errorf("failed to get response message")
}

return msg, nil
```

sboeuf: @devimc any progress on this one ?

devimc:
@jodh-intel  @sboeuf  changes applied , thanks

sboeuf:
LGTM

agent.go
```diff
@@ -26,6 +26,19 @@ import (
 // AgentType describes the type of guest agent a Pod should run.
 type AgentType string
 
+// ProcessListOptions contains the options used to list running
+// processes inside the container
+type ProcessListOptions struct {
+	// Format describes the output format to list the running processes
+	Format string
+
+	// Args contains the list of arguments to run ps command
```
jodh-intel: Again, the agent PR shows that if this isn't set, it default to sane values so I wonder if it would be worth adding "(optional)" at the end of the comment here?

api.go
```diff
@@ -678,3 +678,34 @@ func PausePod(podID string) (VCPod, error) {
 func ResumePod(podID string) (VCPod, error) {
 	return togglePausePod(podID, false)
 }
+
+// ProcessListContainer is the virtcontainers entry point to list
+// processes running inside a container
+func ProcessListContainer(podID, containerID string, options ProcessListOptions) (ProcessList, error) {
```
jodh-intel: Since this changes the API, you'll need to also do the following:

- `interfaces.go`: Add this function to the `VC` interface).
- `implementation.go`: Add a simple wrapper like the one for `SetLogger()`).
-  `pkg/vcMock/types.go`: Again, add a function similar to `SetLoggerFunc`).
- `pkg/vcMock/mock.go`: Add a simple wrapper like the one for `SetLogger()`).
- `pkg/vcMock/mock_test.go`: Add a simple `TestProcessListContainer()`.

container.go
```diff
@@ -707,6 +707,24 @@ func (c *Container) kill(signal syscall.Signal, all bool) error {
...
 }
 
+func (c *Container) processList(options ProcessListOptions) (ProcessList, error) {
+	state, err := c.fetchState("ps")
+	if err != nil {
+		return nil, err
+	}
+
+	if state.State != StateRunning {
+		return nil, fmt.Errorf("Container not running, impossible to list processes")
```
jodh-intel: Can you add `c.id` to the error here? Also, this is racy as the container might be running here, but by the time you call `processListContainer()` below it might have stopped.

hyperstart.go
```diff
@@ -649,3 +649,32 @@ func (h *hyper) killOneContainer(cID string, signal syscall.Signal, all bool) er
...
+	}
+
+	response, err := h.proxy.sendCmd(proxyCmd)
+	if err != nil {
+		return nil, err
+	}
+
+	msg, ok := response.([]byte)
+	if !ok {
+		return nil, fmt.Errorf("failed to get response message")
```
jodh-intel: Would be useful to have `podID` and `cID` in the error here.

jodh-intel:
I should have said that above are addressed, I think this is good to land [1].

---
[1] - I suppose this _could_ land before https://github.com/clearcontainers/agent/pull/90 as the agent will just reject the unknown command.

agent.go
```diff
@@ -26,6 +26,19 @@ import (
 // AgentType describes the type of guest agent a Pod should run.
 type AgentType string
 
+// ProcessListOptions contains the options used to list running
+// processes inside the container
+type ProcessListOptions struct {
+	// Format describes the output format to list the running processes
```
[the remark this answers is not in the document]
devimc: sure, lemme update the comment to specify format is unrelated to ps
default value is set in the runtime [1]

[1] - https://github.com/clearcontainers/runtime/pull/444/files#diff-b985e723bb136b291af1a569f76616d4R32

agent.go
devimc: default value is set in the runtime https://github.com/clearcontainers/runtime/pull/444/files#diff-b985e723bb136b291af1a569f76616d4R67

api.go
devimc: done, thanks

devimc:
@jodh-intel changes applied, thanks

api.go
jodh-intel: Wow that was fast! :smile:

PR #271, opened 2017-05-26
utils: Fix the script installing CNI plugins

sboeuf:
Because CNI recently moved all their plugins to a different repo, we have to update our scripts to make sure we can still build CNI plugins from sources. This is needed for our CI to continue to be functional.
The new repo is: github.com/containernetworking/plugins

### DIFF
--- a/utils/virtcontainers-setup.sh
+++ b/utils/virtcontainers-setup.sh
@@ -59,7 +59,7 @@ cp -r ${busybox_bundle}/* ${pause_bundle}/
 echo -e "Move to ${tmpdir}/${virtcontainers_build_dir}"
 pushd ${tmpdir}/${virtcontainers_build_dir}
 echo "Clone cni"
-git clone https://github.com/containernetworking/cni.git
+git clone https://github.com/containernetworking/plugins.git
 
 echo "Copy CNI config files"
 cp $GOPATH/src/github.com/containers/virtcontainers/test/cni/10-mynet.conf ${ETCDIR}/cni/net.d/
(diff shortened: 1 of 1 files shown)


utils/virtcontainers-setup.sh
```diff
@@ -59,7 +59,7 @@ cp -r ${busybox_bundle}/* ${pause_bundle}/
 echo -e "Move to ${tmpdir}/${virtcontainers_build_dir}"
 pushd ${tmpdir}/${virtcontainers_build_dir}
 echo "Clone cni"
-git clone https://github.com/containernetworking/cni.git
+git clone https://github.com/containernetworking/plugins.git
```
jodh-intel: Since we vendor this package, shouldn't this script really be using the output of `dep status`?

sboeuf: We only vendor this package because of the libcni we need inside virtcontainers. But what we need from this script are the plugins. And since the plugins have been moved to a different repo, there are no reason we want to follow a specific version, does that make sense ?

jodh-intel: The point is that this change *should* have broken our `master` builds, but somehow didn't.

I've re-triggered builds on `master` with Travis and sure enough they fail as a result of the upstream change.

If we make `virtcontainers-setup.sh` use the same version of that repo as we use for vendoring, we're immune from upstream changes and know that what we are testing with is in sync with what virtcontainers is using.

sboeuf: Vendoring the script would have made sense if they didn't move the plugins to a different repo. But now that they are in a separate one, there is no way to track a kind of synchronization between both repos (cni and plugins).

jodh-intel: Frankly, I'm still not happy with this, however I see now that this change is "no worse" than what was there before (which was also just pulling from the tip). So in the interests of unbreaking the build let's merge this. However, I think we need to do better going forward and consider whether we should just pull the bleeding edge here.

jodh-intel:
for the sake of unborking the `master` build...

lgtm.

PR #272, opened 2017-05-26
Generalise *logrus.Logger to the logrus.FieldLogger interface

dlespiau:
We don't really want a Logger object, but either a Logger or an Entry. Using the FieldLogger interface leaves a choice to users.

Took the opportunity to rename SetLog to SetLogger as well, looks slightly better to me.

### DIFF
--- a/api.go
+++ b/api.go
@@ -28,10 +28,10 @@ func init() {
 	runtime.LockOSThread()
 }
 
-var virtLog = logrus.New()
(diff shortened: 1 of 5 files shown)


jodh-intel:
Travis is complaining due to #271 not having landed yet.

jodh-intel:
lgtm

sboeuf:
LGTM

PR #275, opened 2017-06-06
shim: Stop the shim if container has only been created

sboeuf:
In case a shim has been created because a pod or a container has been created, we don't want to end up with a call to DeleteContainer() or DeletePod() that could end up in one or several shims running with no container related.

### DIFF
--- a/pod.go
+++ b/pod.go
@@ -756,6 +756,20 @@ func (p *Pod) stopSetStates() error {
 	return nil
 }
 
+// stopShims stops all remaining shims corresponfing to not started/stopped
+// containers.
+func (p *Pod) stopShims() error {
+	for _, c := range p.containers {
+		if err := stopShim(c.process.Pid); err != nil {
+			return err
+		}
+	}
+
+	virtLog.Infof("Shim(s) stopped")
+
+	return nil
+}
+
 // stopVM stops the agent inside the VM and shut down the VM itself.
 func (p *Pod) stopVM() error {
 	if _, _, err := p.proxy.connect(*p, false); err != nil {
--- a/shim.go
+++ b/shim.go
@@ -18,6 +18,7 @@ package virtcontainers
 
 import (
 	"fmt"
+	"syscall"
 
 	"github.com/mitchellh/mapstructure"
 )
@@ -94,6 +95,20 @@ func newShimConfig(config PodConfig) interface{} {
 	default:
 		return nil
 	}
+}
+
+func stopShim(pid int) error {
+	if pid <= 0 {
+		return nil
+	}
+
+	virtLog.Infof("Stopping shim PID %d", pid)
+
+	if err := syscall.Kill(pid, syscall.SIGKILL); err != nil && err != syscall.ESRCH {
+		return err
+	}
+
+	return nil
 }
 
 // shim is the virtcontainers shim interface.
--- a/shim_test.go
+++ b/shim_test.go
@@ -17,8 +17,13 @@
 package virtcontainers
 
 import (
+	"os/exec"
 	"reflect"
 	"testing"
+)
+
+var (
+	testRunningProcess = "sleep"
 )
 
 func testSetShimType(t *testing.T, value string, expected ShimType) {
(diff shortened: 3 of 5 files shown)


shim.go
```diff
@@ -96,6 +98,36 @@ func newShimConfig(config PodConfig) interface{} {
...
+
+	if err := process.Signal(syscall.Signal(0)); err != nil {
+		return false, nil
+	}
+
+	return true, nil
+}
+
+func stopShim(pid int) error {
+	running, err := shimRunning(pid)
```
jodh-intel: This is racy. Go doesn't exactly make this easy, but I think what you need to do is:

- Call `syscall.RawSyscall` directly.
- Check the returned value (which might be `syscall.ESRCH`).

sboeuf: @jodh-intel sorry but I am not clear on what could happen here and how to solve this ?

jodh-intel: This branch is effectively doing:

```
kill(pid, 0)
<< check for error denoting ESRCH >>
kill(pid, SIGKILL)
```
There is no point doing two kills and its racy since the process might exist when you `kill(pid, 0)`, then die before you `kill(pid, SIGKILL)` it. All you need to do is the actual `SIGKILL` and if the process doesn't exist at that point, you'll get a `ESRCH` back which you can check for.

sboeuf: Got it, makes sense !

sboeuf:
@jodh-intel Please take a look, this is now fixed.

shim_test.go
```diff
@@ -147,3 +152,31 @@ func TestNewShimConfigFromUnknownShimPodConfig(t *testing.T) {
 
 	testNewShimConfigFromPodConfig(t, podConfig, nil)
 }
+
+func TestStopShimSuccessfulProcessNotRunning(t *testing.T) {
+	cmd := exec.Command(testRunningProcess, "0.001")
```
jodh-intel: Nit: that could be just `sleep 0`.

shim_test.go
```diff
@@ -147,3 +152,31 @@ func TestNewShimConfigFromUnknownShimPodConfig(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	if err := stopShim(pid); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestStopShimSuccessfulProcessRunning(t *testing.T) {
+	cmd := exec.Command(testRunningProcess, "1")
```
jodh-intel: Maybe make it `999` to minimise any change of it finishing before the `stopShim()` call.

shim_test.go
```diff
@@ -17,10 +17,15 @@
 package virtcontainers
 
 import (
+	"os/exec"
 	"reflect"
 	"testing"
 )
 
+var (
+	testRunningProcess = "/bin/sleep"
```
jodh-intel: Might be safer to use `exec.LookPath()` here?

pod.go
```diff
@@ -756,6 +756,20 @@ func (p *Pod) stopSetStates() error {
...
+// stopShims stops all remaining shims corresponfing to not started/stopped
+// containers.
+func (p *Pod) stopShims() error {
+	for _, c := range p.containers {
+		if err := stopShim(c.process.Pid); err != nil {
+			return err
+		}
+	}
+
+	virtLog.Infof("Shim(s) stopped")
```
jodh-intel: Rather than this message (or maybe in combination) how about adding a `stopping shim pid %d` log call before the `stopShim()` call?

shim.go
```diff
@@ -96,6 +97,18 @@ func newShimConfig(config PodConfig) interface{} {
 	}
 }
 
+func stopShim(pid int) error {
+	if pid <= 0 {
```
jodh-intel: Maybe this should be `pid <= 1` :smile: Probably not reasonable for this function to make too many assumptions though.

shim_test.go
sboeuf: done

shim_test.go
sboeuf: done

shim_test.go
sboeuf: done

pod.go
sboeuf: done in shim.go directly.

shim.go
sboeuf: Yeah let's keep `pid <= 0` ;)

jodh-intel:
lgtm

sboeuf:
So funny, the coverage decreased -6.0e-05%, it is like 0 !

sameo:
LGTM

PR #276, opened 2017-06-06
Handle bind-mounts and use devicemapper device when available

amshinde:
This PR handles the mounts passed with the OCI spec file. It handles the bind-mounts by passing them to the VM through 9pfs and handling the bind-mounting within the VM to work for both cases where we plan to use overlay as rootfs or a block device.
When devicemapper is the storage driver for the container rootfs, use the underlying block device instead.

### DIFF
--- a/container.go
+++ b/container.go
@@ -102,6 +102,9 @@ type Container struct {
 
 	rootFs string
 
+	// File system type of rootfs block device
+	fstype string
+
 	config *ContainerConfig
 
 	pod *Pod
--- a/container.go
+++ b/container.go
@@ -294,6 +294,11 @@ func createContainers(pod *Pod, contConfigs []ContainerConfig) ([]*Container, er
 			c.process = process
 		}
 
+		mounts, err := c.fetchMounts()
+		if err == nil {
+			c.mounts = mounts
+		}
+
 		containers = append(containers, c)
 	}
 
@@ -330,6 +335,11 @@ func createContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 	process, err := c.fetchProcess()
 	if err == nil {
 		c.process = process
+	}
+
+	mounts, err := c.fetchMounts()
+	if err == nil {
+		c.mounts = mounts
 	}
 
 	state, err := c.pod.storage.fetchContainerState(c.podID, c.id)
@@ -428,6 +438,8 @@ func (c *Container) start() error {
 		c.stop()
 		return err
 	}
+
+	c.storeMounts()
 
 	err = c.setContainerState(StateRunning)
 	if err != nil {
--- a/container.go
+++ b/container.go
@@ -45,6 +45,18 @@ type ContainerStatus struct {
 	Annotations map[string]string
 }
 
+// Mount describes a container mount.
+type Mount struct {
+	Source      string
+	Destination string
+
+	// Type specifies the type of filesystem to mount.
+	Type string
+
+	// Options list all the mount options of the filesystem.
+	Options []string
+}
+
 // ContainerConfig describes one container runtime configuration.
 type ContainerConfig struct {
 	ID string
@@ -62,6 +74,8 @@ type ContainerConfig struct {
 	// for example to add additional status values required
 	// to support particular specifications.
 	Annotations map[string]string
+
+	Mounts []Mount
 }
 
 // valid checks that the container configuration is valid.
@@ -96,6 +110,8 @@ type Container struct {
 	state State
 
 	process Process
+
+	mounts []Mount
 }
 
 // ID returns the container identifier string.
@@ -254,6 +270,7 @@ func createContainers(pod *Pod, contConfigs []ContainerConfig) ([]*Container, er
 			containerPath: filepath.Join(pod.id, contConfig.ID),
 			state:         State{},
 			process:       Process{},
+			mounts:        contConfig.Mounts,
 		}
 
 		state, err := c.pod.storage.fetchContainerState(c.podID, c.id)
--- a/container.go
+++ b/container.go
@@ -55,6 +55,9 @@ type Mount struct {
 
 	// Options list all the mount options of the filesystem.
 	Options []string
+
+	// HostPath used to store host side bind mount path
+	HostPath string
 }
 
 // ContainerConfig describes one container runtime configuration.
--- a/container.go
+++ b/container.go
@@ -186,6 +186,14 @@ func (c *Container) fetchProcess() (Process, error) {
 	return c.pod.storage.fetchContainerProcess(c.podID, c.id)
 }
 
+func (c *Container) storeMounts() error {
+	return c.pod.storage.storeContainerMounts(c.podID, c.id, c.mounts)
+}
+
+func (c *Container) fetchMounts() ([]Mount, error) {
+	return c.pod.storage.fetchContainerMounts(c.podID, c.id)
+}
+
 // fetchContainer fetches a container config from a pod ID and returns a Container.
 func fetchContainer(pod *Pod, containerID string) (*Container, error) {
 	if pod == nil {
--- a/filesystem.go
+++ b/filesystem.go
@@ -45,6 +45,9 @@ const (
 
 	// lockFileType represents a lock file type (pod only)
 	lockFileType
+
+	// mountsFileType represents a mount file type
+	mountsFileType
 )
 
 // configFile is the file name used for every JSON pod configuration.
@@ -61,6 +64,8 @@ const processFile = "process.json"
 
 // lockFile is the file name locking the usage of a pod.
 const lockFileName = "lock"
+
+const mountsFile = "mounts.json"
 
 // dirMode is the permission bits used for creating a directory
 const dirMode = os.FileMode(0750)
@@ -103,6 +108,8 @@ type resourceStorage interface {
 	fetchContainerState(podID, containerID string) (State, error)
 	fetchContainerProcess(podID, containerID string) (Process, error)
 	storeContainerProcess(podID, containerID string, process Process) error
+	fetchContainerMounts(podID, containerID string) ([]Mount, error)
+	storeContainerMounts(podID, containerID string, mounts []Mount) error
 }
 
 // filesystem is a resourceStorage interface implementation for a local filesystem.
@@ -217,7 +224,7 @@ func resourceDir(podSpecific bool, podID, containerID string, resource podResour
 	case configFileType:
 		path = configStoragePath
 		break
-	case stateFileType, networkFileType, processFileType, lockFileType:
+	case stateFileType, networkFileType, processFileType, lockFileType, mountsFileType:
 		path = runStoragePath
 		break
 	default:
@@ -257,6 +264,9 @@ func (fs *filesystem) resourceURI(podSpecific bool, podID, containerID string, r
 		filename = processFile
 	case lockFileType:
 		filename = lockFileName
+		break
+	case mountsFileType:
+		filename = mountsFile
 		break
 	default:
 		return "", "", fmt.Errorf("Invalid pod resource")
@@ -351,6 +361,17 @@ func (fs *filesystem) storeResource(podSpecific bool, podID, containerID string,
 		}
 
 		return fs.storeFile(processFile, file)
+	case []Mount:
+		if resource != mountsFileType {
+			return fmt.Errorf("Invalid pod resource")
+		}
+
+		mountsFile, _, err := fs.resourceURI(podSpecific, podID, containerID, mountsFileType)
+		if err != nil {
+			return err
+		}
+
+		return fs.storeFile(mountsFile, file)
 
 	default:
 		return fmt.Errorf("Invalid resource data type")
@@ -413,8 +434,16 @@ func (fs *filesystem) fetchResource(podSpecific bool, podID, containerID string,
 		}
 
 		return process, nil
-	}
 
+	case mountsFileType:
+		mounts := []Mount{}
+		err = fs.fetchFile(path, &mounts)
+		if err != nil {
+			return nil, err
+		}
+
+		return mounts, nil
+	}
 	return nil, fmt.Errorf("Invalid pod resource")
 }
 
@@ -568,6 +597,32 @@ func (fs *filesystem) fetchContainerProcess(podID, containerID string) (Process,
 
 func (fs *filesystem) storeContainerProcess(podID, containerID string, process Process) error {
 	return fs.storeContainerResource(podID, containerID, processFileType, process)
+}
+
+func (fs *filesystem) fetchContainerMounts(podID, containerID string) ([]Mount, error) {
+	if podID == "" {
+		return []Mount{}, errNeedPodID
+	}
+
+	if containerID == "" {
+		return []Mount{}, errNeedContainerID
+	}
+
+	data, err := fs.fetchResource(false, podID, containerID, mountsFileType)
+	if err != nil {
+		return []Mount{}, err
+	}
+
+	switch mounts := data.(type) {
+	case []Mount:
+		return mounts, nil
+	default:
+		return []Mount{}, fmt.Errorf("Unknown mounts type : [%T]", mounts)
+	}
+}
+
+func (fs *filesystem) storeContainerMounts(podID, containerID string, mounts []Mount) error {
+	return fs.storeContainerResource(podID, containerID, mountsFileType, mounts)
 }
 
 func (fs *filesystem) deleteContainerResources(podID, containerID string, resources []podResource) error {
--- a/filesystem_test.go
+++ b/filesystem_test.go
@@ -325,6 +325,115 @@ func TestFilesystemFetchContainerConfigFailingPodIDEmpty(t *testing.T) {
 	}
 }
 
+func TestFilesystemFetchContainerMountsSuccessful(t *testing.T) {
+	fs := &filesystem{}
+	contID := "100"
+
+	contMountsDir := filepath.Join(runStoragePath, testPodID, contID)
+	os.MkdirAll(contMountsDir, dirMode)
+
+	path := filepath.Join(contMountsDir, mountsFile)
+	os.Remove(path)
+
+	f, err := os.Create(path)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	source := "/dev/sda1"
+	dest := "/root"
+	mntType := "ext4"
+	options := "rw"
+	hostPath := "/tmp/root"
+
+	mountData := fmt.Sprintf(`
+		[
+		  {
+			"Source":"%s",
+			"Destination":"%s",
+			"Type":"%s",
+			"Options": ["%s"],
+			"HostPath":"%s"
+		  }
+		]
+	`, source, dest, mntType, options, hostPath)
+
+	n, err := f.WriteString(mountData)
+	if err != nil || n != len(mountData) {
+		f.Close()
+		t.Fatal(err)
+	}
+	f.Close()
+
+	data, err := fs.fetchContainerMounts(testPodID, contID)
+	if err != nil {
+		data, _ := ioutil.ReadFile(path)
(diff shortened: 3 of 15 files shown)


amshinde:
This PR requires hyperstart PR to be merged first and made available in the container image.
https://github.com/clearcontainers/hyperstart/pull/32

filesystem.go
```diff
@@ -570,6 +599,32 @@ func (fs *filesystem) storeContainerProcess(podID, containerID string, process P
...
+	if err != nil {
+		return []Mount{}, err
+	}
+
+	switch mounts := data.(type) {
+	case []Mount:
+		return mounts, nil
+	}
+
+	return []Mount{}, fmt.Errorf("Unknown mounts type")
```
jodh-intel: Could you add the type (`%T`) to the error?

filesystem_test.go
```diff
@@ -325,6 +326,79 @@ func TestFilesystemFetchContainerConfigFailingPodIDEmpty(t *testing.T) {
...
+	}
+
+	source := "/dev/sda1"
+	dest := "/root"
+	mntType := "ext4"
+	options := "rw"
+	ignoreMount := false
+	hostPath := "/tmp/root"
+
+	mountData := fmt.Sprintf("[{\"Source\":\"%s\",\"Destination\":\"%s\",\"Type\":\"%s\",\"Options\": [\"%s\"],\"IgnoreMount\":%s, \"HostPath\":\"%s\"}]", source, dest, mntType, options, strconv.FormatBool(ignoreMount), hostPath)
```
jodh-intel: This would look better and be easier to read if you used backticks:

```
mountData := fmt.Sprintf(`
  [
     {
        "Source": "%s",
  ...
`, source, ...)

File: mount.go [hunk 3]
```diff
@@ -16,5 +16,78 @@
...
+	// stat.Dev points to the underlying device containing the file
+	major := major(stat.Dev)
+	minor := minor(stat.Dev)
+
+	mountPoint := path
+
+	// We get the mount point by recursively peforming stat on the path
+	// The point where the device changes indicates the mountpoint
+	for {
+		if mountPoint == "/" {
```
mcastelino: We should still check if / results in a switch. If it does not, then we do not have a mount point.

mount.go
```diff
@@ -91,3 +95,43 @@ func getDeviceForPath(path string) (*device, error) {
...
+		if err == io.EOF {
+			break
+		}
+
+		fields := strings.Fields(line)
+		if len(fields) != fieldsPerLine {
+			return "", "", fmt.Errorf("Incorrect no of fields (expected %d, got %d)) :%s", fieldsPerLine, len(fields), line)
+		}
+
+		// fields[0] => Device, fields[1] => Path , fields[2] => Type
```
mcastelino: would having enums here help vs 0, 1, 2? You already have a constant definition above

mount.go
```diff
@@ -91,3 +95,43 @@ func getDeviceForPath(path string) (*device, error) {
...
 	return dev, nil
 }
+
+const (
+	procMountsFile = "/proc/mounts"
+
+	fieldsPerLine = 6
+)
+
+func getDevicePathAndFsType(mountPoint string) (string, string, error) {
```
mcastelino: Nit, named return values would help here.

hyperstart.go
```diff
@@ -437,6 +481,10 @@ func (h *hyper) startPod(pod Pod) error {
 		}
 	}
 
+	if err := h.addDrives(&pod); err != nil {
```
mcastelino: Do you still pass the mounted devicemapper mount pass via 9pfs? I did not see code to disable that? Did I miss it. I am worried about block device corruption if we continue to pass it in via 9p. Even thing like accidental atime update

Also is there a easy way to toggle the behaviour. i.e, switch back to 9pfs even for devicemapper. Offering this option may be required if there are users who may care about the lack of support for dynamic file copy in/out of the container via 

```
docker cp [OPTIONS] CONTAINER:SRC_PATH DEST_PATH|-
docker cp [OPTIONS] SRC_PATH|- CONTAINER:DEST_PATH
```

Also we need to document this limitation with the direct assignment approach.

pkg/oci/utils_test.go
```diff
@@ -86,18 +86,24 @@ func TestMinimalPodConfig(t *testing.T) {
...
 		},
 		{
 			Source:      "devpts",
 			Destination: "/dev/pts",
 			Type:        "devpts",
 			Options:     []string{"nosuid", "noexec", "newinstance", "ptmxmode=0666", "mode=0620", "gid=5"},
+			IgnoreMount: false,
+			HostPath:    "",
 		},
 	}
```
amshinde: @mcastelino  I though /dev/shm needs to be created inside the VM, is that right. I am planning to send all the mount info for the mounts that need to be created within the VM (including /dev/shm) so that hyperstart can create those. Planning to send a separate PR for this.(for virtcontainers and hyperstart)

hyperstart.go
```diff
@@ -220,6 +222,79 @@ func (h *hyper) bindMountContainerRootfs(podID, cID, cRootFs string, readonly bo
...
+		}
+	}
+	return false
+}
+
+// bindMountContainerMounts handles bind-mounts by bindmounting to the host shared directory
+// which is mounted through 9pfs in the VM.
+// Hyperstart uses "fsmap" struct to bind mount bind these mounts in the hypertsart shared directory
+// to the correct mountpoint within the container rootfs.
+func (h *hyper) bindMountContainerMounts(podID string, c Container, mounts *[]Mount) ([]*hyperstart.FsmapDescriptor, error) {
```
amshinde: Yes, I dont need a pointer here. Will fix this.

filesystem_test.go
amshinde: Didnt know about the backticks. Will fix this.

filesystem.go
amshinde: Fixing this

mount.go
amshinde: Fixing.

pkg/oci/utils_test.go
mcastelino: Yes. It has to be created inside. But the size needs to be picked up from the outside. So I thought sending this value in via the same mechanism would make sense.

hyperstart.go
amshinde: Yes, we are still passing the mounted devicemapper mount pass via 9pfs. I didnt think this would cause any issues. I'll disable the 9pfs mount.
Will add the documentation too.
I suppose we could add a configuration option in  the global config file for the runtime to toggle the behaviour. I'll add that to the runtime.

pkg/oci/utils_test.go
amshinde: Yes, I was planning to send this is a separate structure say "mounts" to hyperstart. But I suppose i could get away with extending the current fsmap type to handle this as well. I'll add the system mounts handling in a separate PR.

hyperstart.go
```diff
@@ -220,6 +222,79 @@ func (h *hyper) bindMountContainerRootfs(podID, cID, cRootFs string, readonly bo
...
+	var fsmap []*hyperstart.FsmapDescriptor
+
+	// TODO: We need to handle system mounts by having the agent create them inside the VM.
+	// Handle just bind mounts for now
+
+	for ind := range *mounts {
+		m := &(*mounts)[ind]
+
+		m.IgnoreMount = true
+		if isPartOf(systemMounts, m.Destination) {
```
amshinde: Yes makes sense.

pkg/oci/utils_test.go
```diff
@@ -80,6 +80,27 @@ func TestMinimalPodConfig(t *testing.T) {
...
 
+	expectedMounts := []vc.Mount{
+		{
+			Source:      "proc",
+			Destination: "/proc",
+			Type:        "proc",
+			Options:     nil,
+		},
+		{
```
amshinde: @mcastelino  These are just the tests, but I am currently doing the bind-mounts in the same namespace as host. Need to add the change to enter a separate mount namespace for 3.0

mount.go
```diff
@@ -0,0 +1,20 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package virtcontainers
+
+// These mounts need to be created by the agent within the VM
+var systemMounts = []string{"/proc", "/dev", "/dev/pts", "/dev/shm", "/dev/mqueue", "/sys", "/sys/fs/cgroup"}
```
amshinde: The /dev would have been sufficient here, basically I ignore mounts with that prefix.We use bind inside the VM. I think the fsmap structure needs to be extended to include extra mount options.

mount.go
amshinde: fixing

mount.go
amshinde: @mcastelino I dont quite understand what you mean, I am using "/" as the termination case while walking up the directory tree. When i reach "/", that should be the mount point.

mcastelino: @amshinde let us take two cases.

If you have /mydisk where mydisk is just a directory on / 
vs /mydisk where you have mounted something on /mydisk

In the first case /mydisk and / will have the same stat
In the second case /mydisk and / will have different stats.

So in the first case we error out saying we did not find a mount point.
In the second case /mydisk is the mount point.

The only time / is a valid mount point will be in the case of volumes where the user may have done something along the lines of -v /:/hostdisk.
In such a case yes / is the mount point. But that is a special case, where the source is /.

hyperstart.go
```diff
@@ -292,6 +292,7 @@ func (h *hyper) bindMountContainerMounts(podID string, c Container, mounts []Mou
 		fsmap = append(fsmap, fsmapDesc)
 	}
 
+	c.storeMounts()
```
sameo: The container mount storage is generic operation and should be done at the container or pod level. We don't want each potential hypervisor interface implementation to explicitly call this one for each container it creates.

hyperstart.go
```diff
@@ -296,6 +296,24 @@ func (h *hyper) bindMountContainerMounts(podID string, c Container, mounts []Mou
 	return fsmap, nil
 }
 
+func (h *hyper) bindUnmountContainerMounts(podID string, c *Container) error {
```
sameo: This is a container generic routine and should be implemented as a private `Container` method instead. Again, we want that one to be usable by all potential hypervisor implementations.

hyperstart.go
```diff
@@ -305,7 +323,9 @@ func (h *hyper) bindUnmountContainerRootfs(podID, cID string) error {
 
 func (h *hyper) bindUnmountAllRootfs(pod Pod) {
 	for _, c := range pod.containers {
+		h.bindUnmountContainerMounts(pod.id, c)
 		h.bindUnmountContainerRootfs(pod.id, c.id)
+		h.bindUnmountContainerMounts(pod.id, c)
```
sameo: Why the double unmount ?

pod.go
```diff
@@ -212,6 +212,14 @@ func (s *Sockets) String() string {
 	return strings.Join(sockSlice, " ")
 }
 
+// Drive represents a block storage drive in the case where the graph storage
+// driver has an underlying block storage device.
+type Drive struct {
```
sameo: You may want to document this structure fields.

mount.go
```diff
@@ -16,5 +16,78 @@
...
+
+func major(dev uint64) int {
+	return int((dev >> 8) & 0xfff)
+
+}
+
+func minor(dev uint64) int {
+	return int((dev & 0xff) | ((dev >> 12) & 0xfff00))
+}
+
```
sameo: Something like:

```Go
var (	
        MINORBITS = uint(20)
        MINORMASK = uint((uint(1) << MINORBITS) - 1)
)

func major(dev uint) int{
	return int(dev >> MINORBITS)
}

func minor(dev uint) int{
	return int(dev & MINORMASK)
}

```

would be simpler.

mount.go
```diff
@@ -16,5 +16,78 @@
...
+type device struct {
+	major      int
+	minor      int
+	mountPoint string
+}
+
+// getDeviceForPath gets the underlying device containing the file specified by path.
+// The device type constitutes the major-minor number of the device and the dest mountPoint for the device
+func getDeviceForPath(path string) (*device, error) {
```
sameo: Please just return a `device`

mount.go
```diff
@@ -16,5 +16,78 @@
...
+}
+
+type device struct {
+	major      int
+	minor      int
+	mountPoint string
+}
+
+// getDeviceForPath gets the underlying device containing the file specified by path.
+// The device type constitutes the major-minor number of the device and the dest mountPoint for the device
```
sameo: Could you give an example of what this function is supposed to do ? Given path `/foo/bar/` what would be the device ?
For example, let's say that `/dev/sda1` is mounted on `/a/b/c/`. If I pass `/a/b/c/file1` to this routine it should return:
```Go
device{
     major: major(/dev/sda3)
     minor: minor(/dev/sda3)
     mountPoint: /a/b/c/
}
```

Correct ?

mount.go
```diff
@@ -91,3 +95,56 @@ func getDeviceForPath(path string) (*device, error) {
...
+	fieldsPerLine = 6
+)
+
+const (
+	procDeviceIndex = iota
+	procPathIndex
+	procTypeIndex
+)
+
+func getDevicePathAndFsType(mountPoint string) (devicePath, fsType string, err error) {
```
sameo: So there's a kubernetes package for doing a bunch of mount and device operations:
`import "k8s.io/kubernetes/pkg/util/mount"`

And then you can do things like:

```Go
mounter := mount.Mounter{}
mountPoints, err := mounter.List()
```

where `mountPoints` is an array of:

```Go
// This represents a single line in /proc/mounts or /etc/fstab.
type MountPoint struct {
	Device string
	Path   string
	Type   string
	Opts   []string
	Freq   int
	Pass   int
}
```

It also provides some other useful routines like `mounter.PathIsDevice()` or `mount.IsNotMountPoint()` that you could use in your previous patch to reduce code duplication.

mount.go
```diff
@@ -148,3 +148,20 @@ func getDevicePathAndFsType(mountPoint string) (devicePath, fsType string, err e
...
+func isDeviceMapper(major, minor int) (bool, error) {
+
+	//Check if /sys/dev/block/${major}-${minor}/dm exists
+	sysPath := fmt.Sprintf("/sys/dev/block/%d:%d/dm", major, minor)
+
+	_, err := os.Stat(sysPath)
+
+	if err == nil {
+		return true, nil
+	} else if os.IsNotExist(err) {
```
sameo: Why do we make that not an error ?

hyperstart.go
```diff
@@ -403,6 +403,50 @@ func (h *hyper) exec(pod *Pod, c Container, process Process, cmd Cmd) error {
 	return nil
 }
 
+func (h *hyper) addDrives(pod *Pod) error {
```
sameo: Here as well, this should be a `Pod` private method for other hypervisor implementations to use the same logic.

sameo:
This mostly looks good to me, I only have requests for minor changes.
@amshinde One thing I'd like you to do though is to split this PR in two: One for the bind mount part and another one for the devicemapper support.

hyperstart.go
amshinde: One is for the container rootfs and the other is for the bind-mounts that we created in the shared directory

container.go
```diff
@@ -55,6 +55,12 @@ type Mount struct {
 
 	// fstab style mount options
 	Options []string
+
+	// IgnoreMount denotes these need to be handled explicitly by the agent within the VM
+	IgnoreMount bool
```
amshinde: Will do

hyperstart.go
amshinde: The implementation for the bind-mounting is agent/hyperstart specific. That's why I put it here, along the same lines of bindUnmountContainerRootfs function that was defined for hyperstart

mount.go
amshinde: Thats right, I'll add  example to the function documentation

mount.go
amshinde: We want to check if the path exists or not to determine if its a devicemapper device. If it exists, we return true else false if it does not. All other errors are treated as errors

mount.go
amshinde: @sameo The above code does not seem to give the correct major-minor numbers. I know thats the how its documented in libc. This is what I followed:
https://github.com/golang/go/issues/8106

mount.go
```diff
@@ -0,0 +1,200 @@
...
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"syscall"
+)
+
+// These mounts need to be created by the agent within the VM
+var systemMounts = []string{"/proc", "/dev", "/dev/pts", "/dev/shm", "/dev/mqueue", "/sys", "/sys/fs/cgroup"}
```
sboeuf: Should be defined in `pkg/hyperstart` instead.

sameo: This is not really an hyperstart specific list, it can potentially apply to any other agent.

hyperstart.go
```diff
@@ -220,6 +222,98 @@ func (h *hyper) bindMountContainerRootfs(podID, cID, cRootFs string, readonly bo
...
+			Source:       filename,
+			Path:         m.Destination,
+			ReadOnly:     readonly,
+			DockerVolume: false,
+		}
+
+		fsmap = append(fsmap, fsmapDesc)
+	}
+
+	c.storeMounts()
```
amshinde: No, we updated the HostPath for the mount here.
I have moved the storeMount outside at the pod start in any case, will push an update

utils.go
```diff
@@ -41,3 +42,22 @@ func fileCopy(srcPath, dstPath string) error {
...
+	}
+
+	return b, nil
+}
+
+func reverseString(s string) string {
+	r := []rune(s)
+	for i, j := 0, len(r)-1; i < len(r)/2; i, j = i+1, j-1 {
+		r[i], r[j] = r[j], r[i]
+	}
```
amshinde: Fixing

hyperstart.go
```diff
@@ -220,6 +222,98 @@ func (h *hyper) bindMountContainerRootfs(podID, cID, cRootFs string, readonly bo
...
+		filename := fmt.Sprintf("%s-%s-%s", c.id, hex.EncodeToString(randBytes), filepath.Base(m.Destination))
+		mountDest := filepath.Join(defaultSharedDir, podID, filename)
+
+		err = bindMount(m.Source, mountDest, false)
+		if err != nil {
+			return nil, err
+		}
+
+		m.IgnoreMount = false
+		m.HostPath = mountDest
```
amshinde: Because I am passing a slice here, the underlying elements can be modified.

amshinde:
Closing this PR. Opening 2 others for mounts and devicemapper

filesystem.go
```diff
@@ -570,6 +599,32 @@ func (fs *filesystem) storeContainerProcess(podID, containerID string, process P
...
+	data, err := fs.fetchResource(false, podID, containerID, mountsFileType)
+	if err != nil {
+		return []Mount{}, err
+	}
+
+	switch mounts := data.(type) {
+	case []Mount:
+		return mounts, nil
+	default:
+		return []Mount{}, fmt.Errorf("Unknown mounts type : [%T]", mounts)
```
amshinde: Used this since I wanted to display the type here

sboeuf: You can display the type outside this switch since you assigned the `mounts` variable before entering the switch, right ?

sboeuf: I am sorry I have deleted your comment by mistake. But I agree that you need this way to display the type.

mount.go
amshinde: sameo had a look at that package, there is one function that I can use for parsing /proc/mounts file as of now. I'd rather just do that, then import that package.

PR #278, opened 2017-06-08
Bind mounts

amshinde:


container.go
```diff
@@ -410,6 +448,8 @@ func (c *Container) start() error {
 		return err
 	}
 
+	c.storeMounts()
```
sboeuf: I think you should remove that and store the mount points from the agent itself.

pod.go
```diff
@@ -718,6 +718,10 @@ func (p *Pod) start() error {
 		return err
 	}
 
+	for _, c := range p.containers {
+		c.storeMounts()
```
sboeuf: I think you should remove that and store the mount points from the agent itself.

hyperstart.go
```diff
@@ -430,6 +512,18 @@ func (h *hyper) startOneContainer(pod Pod, c Container) error {
 		return err
 	}
 
+	//TODO : Enter mount namespace
+
+	// Handle container mounts
+	fsmap, err := h.bindMountContainerMounts(pod.id, c.id, c.config.Mounts)
```
sboeuf: We should pass `c.mounts` instead of `c.config.Mounts` because we don't want the initial config to be modified here.

hyperstart.go
```diff
@@ -430,6 +512,18 @@ func (h *hyper) startOneContainer(pod Pod, c Container) error {
...
+	if err != nil {
+		h.bindUnmountAllRootfs(pod)
+		return err
+	}
+
```
sboeuf: We should call `c.storeMounts()` from here.
I think it should really be stored after we know that we actually modified the container mounts list. Otherwise, doing that into container.go or pod.go is not obvious at all.

container.go
amshinde: No, I think this should be done at the container/pod level from having every agent implementation having to do this.
@sameo What do you think, you were mentioning the same right?

hyperstart.go
amshinde: Agreed

mount.go
```diff
@@ -0,0 +1,36 @@
...
+
+func isSystemMount(m string) bool {
+	for _, p := range systemMountPrefixes {
+		if m == p {
+			return true
+		}
+		if strings.HasPrefix(m, p+"/") {
+			return true
+		}
+	}
```
amshinde: fixed

mount.go
```diff
@@ -0,0 +1,36 @@
...
+import "strings"
+
+// These mounts need to be created by the agent within the VM
+var systemMounts = []string{"/proc", "/dev", "/dev/pts", "/dev/shm", "/dev/mqueue", "/sys", "/sys/fs/cgroup"}
+
+var systemMountPrefixes = []string{"/proc", "/dev", "/sys"}
+
+func isSystemMount(m string) bool {
+	for _, p := range systemMountPrefixes {
+		if m == p {
```
amshinde: fixed

hyperstart.go
```diff
@@ -467,12 +561,12 @@ func (h *hyper) stopPauseContainer(podID string) error {
 
 // stopContainer is the agent Container stopping implementation for hyperstart.
 func (h *hyper) stopContainer(pod Pod, c Container) error {
-	return h.stopOneContainer(pod.id, c.id)
+	return h.stopOneContainer(pod.id, &c)
 }
 
-func (h *hyper) stopOneContainer(podID, cID string) error {
+func (h *hyper) stopOneContainer(podID string, c *Container) error {
```
amshinde: fixed

hyperstart.go
amshinde: fixed

container.go
sboeuf: I am not saying this is not a good idea to do that at the pod/container level, but according to the way we modify this list of mounts from inside the agent, the agent itself should store the new list. I would be more than happy to see this modification (list of mounts I mean) out of the agent implementation but I think it is a bit of work to refactor this.
Therefore, according to the way it is implemented here (modified inside the agent implementation), I think the agent implementation should take care of storing this mount list.

sameo:
@amshinde Please provide a PR description. We need to know what you did, why and how.

hyperstart.go
sameo: I'd rather do it from a more generic place, to avoid code duplication across hypervisors. I know we only support one for now, but it's an architecture discipline we should follow.

container.go
sameo: @sboeuf What would we need to refactor ? Once the container is started, we can safely store the bind mount information from outside the agent. Please elaborate.

sameo:
> This means that if we want everything out of the hyperstart agent implementation, this will be done in another PR because we have to change a lot of things I guess.

@sboeuf I may be missing your point, but the mount list is somehow agent specific and should be created there most likely. It can be stored outside of the agent code though as I think we can safely assume the bind mount list is created once the container is started.

sboeuf:
@sameo It is currently agent specific but I guess I was hoping that it could be different in the future. Anyway, my main concern here is that it is not clear that we modify the mount list from a call to the agent, and then we store this from the container/pod level. But if you think this is clear enough, then I am fine with this.

sameo:
LGTM

PR #280, opened 2017-06-09
Make state functions update in memory state too

jodh-intel:


pod.go
```diff
@@ -833,6 +854,20 @@ func (p *Pod) setPodState(state State) error {
 	return nil
 }
 
+func (p *Pod) getContainer(containerID string) (*Container, error) {
+	if containerID == "" {
+		return &Container{}, errNeedContainerID
+	}
+
+	for _, c := range p.containers {
+		if c.ID() == containerID {
```
sboeuf: You don't need the getter here because we are in the virtcontainers lib itself, you can do `c.id` instead of `c.ID()`.

sboeuf:
@jodh-intel I like the PR, it makes things clearer. Have you tested that with our runtime to make sure things are still working and it does not introduce a regression. I don't think it would but it always better to check !

sameo:
LGTM

jodh-intel:
Hi @sboeuf - yes, tested with https://github.com/clearcontainers/runtime/commit/66b54638c5125167fe852dce4a45c2a1d01c9e27.

sboeuf:
LGTM

PR #281, opened 2017-06-10
Use block device for container rootfs in case of devicemapper

amshinde:
In case of device mapper storage driver, the overlay file system is backed by a block device.
With this PR, we directly use the block device for container rootfs, instead of the overlay. As we do not go through 9pfs, this gives us much better I/O performance. We go through 9pfs only for the bind-mounts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,3 +183,23 @@ __Runtime network setup with CNI__
 4. Create bridge, TAP, and link all together with network interface previously created ([code](https://github.com/containers/virtcontainers/blob/0.5.0/network.go#L123-L205))
 
 5. Start VM inside the netns and start the container ([code](https://github.com/containers/virtcontainers/blob/0.5.0/api.go#L66-L70))
+
+## Storage
+
+Container workloads are shared with the virtualized environment through 9pfs.
+The devicemapper storage driver is a special case. The driver uses dedicated block devices rather than formatted filesystems, and operates at the block level rather than the file level. This knowledge has been used to directly use the underlying block device instead of the overlay file system for the container root file system. The block device maps to the top read-write layer for the overlay. This approach gives much better I/O performance compared to using 9pfs to share the container file system.
+
+The approach above does introduce a limitation in terms of dynamic file copy in/out of the container via `docker cp` operations.
+The copy operation from host to container accesses the mounted file system on the host side. This is not expected to work and may lead to inconsistencies as the block device will be simultaneously written to, from two different mounts.
+The copy operation from container to host will work, provided the user calls `sync(1)` from within the container prior to the copy to make sure any outstanding cached data is written to the block device.
+
+```
+docker cp [OPTIONS] CONTAINER:SRC_PATH HOST:DEST_PATH
+docker cp [OPTIONS] HOST:SRC_PATH CONTAINER:DEST_PATH
+```
+
+The devicemapper block device can only be used when creating a pod. If a container with a devicemapper rootfs is added to a pod, after the VM has started, the devicemapper block device will not be used. The container will fallback to using the overlay file system instead. This should be fixed once capability to hot-plug virtio block devices is added.
+
+#### How to check if container uses devicemapper block device as its rootfs
+
+Start a container. Call mount(8) within the container. You should see '/' mounted on /dev/vda device.
--- a/container.go
+++ b/container.go
@@ -102,9 +102,6 @@ type Container struct {
 
 	rootFs string
 
-	// File system type of rootfs block device
-	fstype string
-
 	config *ContainerConfig
 
 	pod *Pod
--- a/container.go
+++ b/container.go
@@ -152,6 +152,28 @@ func (c *Container) SetPid(pid int) error {
 	return c.storeProcess()
 }
 
+func (c *Container) setStateBlockIndex(index int) error {
+	c.state.BlockIndex = index
+
+	err := c.pod.storage.storeContainerResource(c.pod.id, c.id, stateFileType, c.state)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *Container) setStateFstype(fstype string) error {
+	c.state.Fstype = fstype
+
+	err := c.pod.storage.storeContainerResource(c.pod.id, c.id, stateFileType, c.state)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
 // URL returns the URL related to the pod.
 func (c *Container) URL() string {
 	return c.pod.URL()
--- a/container.go
+++ b/container.go
@@ -102,6 +102,9 @@ type Container struct {
 
 	rootFs string
 
+	// File system type of rootfs block device
+	fstype string
+
 	config *ContainerConfig
 
 	pod *Pod
--- a/container.go
+++ b/container.go
@@ -290,7 +290,7 @@ func newContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 
 	state, err := c.pod.storage.fetchContainerState(c.podID, c.id)
 	if err == nil {
-		c.state.State = state.State
+		c.state = state
 	}
 
 	process, err := c.pod.storage.fetchContainerProcess(c.podID, c.id)
--- a/container.go
+++ b/container.go
@@ -242,12 +242,10 @@ func (c *Container) setContainerState(state stateString) error {
 	}
 
 	// update in-memory state
-	c.state = State{
-		State: state,
-	}
+	c.state.State = state
 
 	// update on-disk state
-	err := c.pod.storage.storeContainerResource(c.podID, c.id, stateFileType, c.state)
+	err := c.pod.storage.storeContainerResource(c.pod.id, c.id, stateFileType, c.state)
 	if err != nil {
 		return err
 	}
--- a/hyperstart.go
+++ b/hyperstart.go
@@ -323,7 +323,9 @@ func (h *hyper) bindUnmountContainerRootfs(podID, cID string) error {
 func (h *hyper) bindUnmountAllRootfs(pod Pod) {
 	for _, c := range pod.containers {
 		h.bindUnmountContainerMounts(c.mounts)
-		h.bindUnmountContainerRootfs(pod.id, c.id)
+		if c.fstype == "" {
+			h.bindUnmountContainerRootfs(pod.id, c.id)
+		}
 	}
 }
 
@@ -541,11 +543,12 @@ func (h *hyper) startOneContainer(pod Pod, c Container) error {
 		container.Fstype = c.fstype
 		container.Image = driveName
 		diskIndex = diskIndex + 1
-	}
+	} else {
 
-	if err := h.bindMountContainerRootfs(pod.id, c.id, c.rootFs, false); err != nil {
-		h.bindUnmountAllRootfs(pod)
-		return err
+		if err := h.bindMountContainerRootfs(pod.id, c.id, c.rootFs, false); err != nil {
+			h.bindUnmountAllRootfs(pod)
+			return err
+		}
 	}
 
 	//TODO : Enter mount namespace
@@ -618,8 +621,10 @@ func (h *hyper) stopOneContainer(podID string, c Container) error {
 		return err
 	}
 
-	if err := h.bindUnmountContainerRootfs(podID, c.id); err != nil {
-		return err
+	if c.fstype == "" {
+		if err := h.bindUnmountContainerRootfs(podID, c.id); err != nil {
+			return err
+		}
 	}
 
 	return nil
--- a/hyperstart.go
+++ b/hyperstart.go
@@ -517,6 +517,8 @@ func (h *hyper) startPauseContainer(podID string) error {
 	return nil
 }
 
+var diskIndex = int(0)
+
 func (h *hyper) startOneContainer(pod Pod, c Container) error {
 	process, err := h.buildHyperContainerProcess(c.config.Cmd)
 	if err != nil {
@@ -528,6 +530,17 @@ func (h *hyper) startOneContainer(pod Pod, c Container) error {
 		Image:   c.id,
 		Rootfs:  rootfsDir,
 		Process: process,
+	}
+
+	if c.fstype != "" {
+		driveName, err := getVirtDriveName(diskIndex)
+		if err != nil {
+			return err
+		}
+
+		container.Fstype = c.fstype
+		container.Image = driveName
+		diskIndex = diskIndex + 1
 	}
 
 	if err := h.bindMountContainerRootfs(pod.id, c.id, c.rootFs, false); err != nil {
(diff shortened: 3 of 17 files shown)


utils.go
```diff
@@ -41,3 +43,37 @@ func fileCopy(srcPath, dstPath string) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return b, nil
+}
+
+func reverseString(s string) string {
+	r := []rune(s)
+	for i, j := 0, len(r)-1; i < len(r)/2; i, j = i+1, j-1 {
```
jodh-intel: Why not only call `len(r)` once?

container.go
```diff
@@ -45,6 +45,21 @@ type ContainerStatus struct {
...
 
+// Mount describes a container mount.
+type Mount struct {
+	Source      string
+	Destination string
+
+	// Type specifies the type of filesystem to mount.
+	Type string
+
+	// Options list all the mount options of the filesystem.
```
jodh-intel: What is the order? Sorted or original ordering? Are duplicate options handled?

jodh-intel:
@amshinde - I appreciate you have added some tests, but this is a big PR and it's not ideal that the test coverage is falling. Could you find a way to add some more to keep the coverage change `>= 0`? :-)

mount.go
```diff
@@ -0,0 +1,232 @@
...
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"syscall"
+)
+
+// These mounts need to be created by the agent within the VM
+var systemMounts = []string{"/proc", "/dev", "/dev/pts", "/dev/shm", "/dev/mqueue", "/sys", "/sys/fs/cgroup"}
```
sameo: The order does not really matter, but I agree it could be better indented for easier maintenance.

sameo:
Could you please rebase ? It will make the PR look smaller at least :)

amshinde:
Rebased this PR.

amshinde:
Added documentation pointing out the changes introduced with devicemapper and limitations.

pod.go
```diff
@@ -998,6 +998,10 @@ func (p *Pod) checkContainersState(state stateString) error {
 	return nil
 }
 
+// Drive names are dependent on the order in which they are added to the hypervisor options.
+// driveIndex is an incrementing counter used to track the index at which current drive is added to the hypervisor.
+var driveIndex = int(0)
```
sboeuf: We can do better than this, by adding this index to the hypervisor structure maybe. There is clearly no need for a global variable here. And more than that, keep in mind that we don't need to store this data because this will only be used by CreatePod(). Maybe this will be interesting to store in case we consider hotplugging some new block devices when the VM is already started.

container.go
```diff
@@ -152,6 +152,34 @@ func (c *Container) SetPid(pid int) error {
...
+	if err != nil {
+		return err
+	}
+
+	c.state = tState
+
+	return nil
+}
+
+func (c *Container) setStateFstype(fstype string) error {
```
sboeuf: Ditto for this function.

README.md
```diff
@@ -183,3 +183,17 @@ __Runtime network setup with CNI__
 4. Create bridge, TAP, and link all together with network interface previously created ([code](https://github.com/containers/virtcontainers/blob/0.5.0/network.go#L123-L205))
 
 5. Start VM inside the netns and start the container ([code](https://github.com/containers/virtcontainers/blob/0.5.0/api.go#L66-L70))
+
+## Storage
```
sboeuf: Thanks for updating the doc here. Could you also add the limitation that devicemapper block device can only be used when creating a new pod. Adding a new container (with a devicemapper rootfs) after we started the VM won't work.

mount.go
```diff
@@ -122,3 +125,56 @@ func getDeviceForPath(path string) (device, error) {
...
+
+	file, err = os.Open(procMountsFile)
+	if err != nil {
+		return
+	}
+
+	defer file.Close()
+
+	reader := bufio.NewReader(file)
+	for {
```
amshinde: @sboeuf /proc/mounts output can change, but the mount info for this device will not change(It has been mounted by docker and will be unmounted by it as well) Using /proc/mounts is a standard way to get mount info.

container.go
```diff
@@ -152,6 +152,34 @@ func (c *Container) SetPid(pid int) error {
 	return c.storeProcess()
 }
 
+func (c *Container) setStateBlockIndex(index int) error {
+	tState := c.state
+	tState.BlockIndex = index
```
amshinde: I am just updating a temporary object here first. That way if the write to filesystem fails, the object in memory and filesystem are in sync

amshinde:
@sboeuf I have pushed all your changes except for the call to addDrives(). Will wait for @sameo to comment on it.

README.md
```diff
@@ -183,3 +183,19 @@ __Runtime network setup with CNI__
...
+
+Container workloads are shared with the virtualized environment through 9pfs.
+The devicemapper storage driver is a special case. The driver uses dedicated block devices rather than formatted filesystems, and operates at the block level rather than file level. We use this knowledge to directly use the underlying block device instead of the overlay file system for the container root file system. The block device maps to the top read-write layer for the overlay. This approach gives much better I/O performance as compared to using 9pfs to share the container file system.
```
jodh-intel: s/rather than file level/rather than the file level/

s/performance as compared/performance compared/

Also, this change mixes familiar and passive voices. I'd personally reword to allow you to drop all the "We"'s.

README.md
```diff
@@ -183,3 +183,19 @@ __Runtime network setup with CNI__
...
+
+The approach above does introduce a limitation in terms of dynamic file copy in/out of the container via docker cp operations:
```
jodh-intel: formatting: `docker cp`.

What limitation? I think it would help to explain it first, then give examples.

README.md
```diff
@@ -183,3 +183,19 @@ __Runtime network setup with CNI__
...
+
+```
+docker cp [OPTIONS] CONTAINER:SRC_PATH DEST_PATH|-
+docker cp [OPTIONS] SRC_PATH|- CONTAINER:DEST_PATH
```
jodh-intel: What are the pipe characters for?

README.md
```diff
@@ -183,3 +183,19 @@ __Runtime network setup with CNI__
...
+```
+
+While the copy to container from host is not expected to work, it is possible to copy files from the container to host. We do need to sync the files to the block device before the copy operation.
```
jodh-intel: "from host ... to container" sounds more natural to me.

Overall, I find these two sentences confusing as it is unclear to the reader if they need to do anything "special": who needs to sync the files? virtcontainers, or is it up to the user to call `sync(1)` potentially multiple times to ensure the data is persisted?

README.md
```diff
@@ -183,3 +183,19 @@ __Runtime network setup with CNI__
...
+
+The devicemapper block device can only be used when creating a pod. Adding a container with a devicemapper rootfs after the VM has started does not work currently.
```
jodh-intel: - What behaviour will the user see if they try to add a devicemapper rootfs after pod creation? Is it handled elegantly, or do you get a crash / odd error message?
- Will it work in the future? (Is there an issue for this?)
- How can you tell if a pod was created with a devicemapper rootfs after the event?

hyperstart.go
```diff
@@ -323,7 +323,9 @@ func (h *hyper) bindUnmountContainerRootfs(podID, cID string) error {
 func (h *hyper) bindUnmountAllRootfs(pod Pod) {
 	for _, c := range pod.containers {
 		h.bindUnmountContainerMounts(c.mounts)
-		h.bindUnmountContainerRootfs(pod.id, c.id)
+		if c.state.Fstype == "" {
+			h.bindUnmountContainerRootfs(pod.id, c.id)
```
jodh-intel: Missing error check on this call.

hyperstart.go
```diff
@@ -342,6 +344,23 @@ func (h *hyper) init(pod *Pod, config interface{}) (err error) {
 	// Override pod agent configuration
 	pod.config.AgentConfig = h.config
 
+	created, err := pod.isPodCreated()
+	if err != nil {
+		return err
+	}
+
+	if !created {
+		virtLog.Infof("Pod %s has not competed creation yet, adding hypervisor devices", pod.id)
```
jodh-intel: typo: "completed". I'd write as, `Pod %s not fully created yet, adding hypervisor devices`

mount.go
```diff
@@ -32,3 +41,189 @@ func isSystemMount(m string) bool {
...
+		var line string
+
+		line, err = reader.ReadString('\n')
+		if err == io.EOF {
+			err = fmt.Errorf("Mount %s not found", mountPoint)
+			break
```
jodh-intel: This still looks a little odd to me - I'd just `return fmt.Errorf(...)` for all these cases as a return cleary breaks out of the loop.

mount.go
```diff
@@ -32,3 +41,189 @@ func isSystemMount(m string) bool {
...
+		}
+
+		if mountPoint == fields[procPathIndex] {
+			devicePath = fields[procDeviceIndex]
+			fsType = fields[procTypeIndex]
+			return
+		}
+	}
+
+	return
```
jodh-intel: If you immediately return an error at the call site, this would presumably need to become a catch-all error.

mount.go
```diff
@@ -32,3 +41,189 @@ func isSystemMount(m string) bool {
...
+
+// getVirtBlockDriveName returns the disk name format for virtio-blk
+// Reference: https://github.com/torvalds/linux/blob/master/drivers/block/virtio_blk.c @c0aa3e0916d7e531e69b02e426f7162dfb1c6c0
+func getVirtDriveName(index int) (string, error) {
+	if index < 0 {
+		return "", fmt.Errorf("Index cannot be negative for drive")
+	}
+
+	// Prefix used for virtio-block devices
+	prefix := "vd"
```
jodh-intel: This could be `const`

mount_test.go
```diff
@@ -41,3 +49,132 @@ func TestIsSystemMount(t *testing.T) {
...
+
+	expected := "/proc"
+
+	if dev.mountPoint != expected {
+		t.Fatalf("Expected %s mountpoint, got %s", expected, dev.mountPoint)
+	}
+}
+
+func TestGetDeviceForPathEmptyPath(t *testing.T) {
+	_, err := getDeviceForPath("")
```
jodh-intel: I think there is scope to add some more tests for this function:

- `getDeviceForPath("///")`
- `getDeviceForPath("/../../.././././../.")`
- what if the specified path has spaces / special characters / unicode?
- create a test that creates a bind mount manually and see if `getDevicePath()` returns the expected value.

mount.go
```diff
@@ -32,3 +41,189 @@ func isSystemMount(m string) bool {
...
+			fsType = fields[procTypeIndex]
+			return
+		}
+	}
+
+	return
+}
+
+// isDeviceMapper checks if the device with the major and minor numbers is a devicemapper block device
+func isDeviceMapper(major, minor int) (bool, error) {
```
jodh-intel: Missing tests. You could create a variable this function uses (something like, `var formatTemplate  := /sys/dev/block/%d:%d/dm`), and have the tests modify that so you can check the expected return value.

mount_test.go
```diff
@@ -41,3 +49,132 @@ func TestIsSystemMount(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	if path != "proc" || fstype != "proc" {
+		t.Fatal(err)
+	}
+}
+
+func TestGetVirtDriveNameInvalidIndex(t *testing.T) {
+	index := -1
```
jodh-intel: You could drop this var and just pass `-1` directly as it's only used once.

hyperstart.go
amshinde: I have just wrapped the existing call in an if condition. This will require changes up the stack, something that I want to do in a separate PR.

amshinde:
Rebased this PR, addressed review comments, fixed documentation and added some more tests.
@jodh-intel , @sboeuf , @sameo Please have a look.

README.md
```diff
@@ -183,3 +183,23 @@ __Runtime network setup with CNI__
...
+
+## Storage
+
+Container workloads are shared with the virtualized environment through 9pfs.
+The devicemapper storage driver is a special case. The driver uses dedicated block devices rather than formatted filesystems, and operates at the block level rather than the file level. This knowledge has been used to directly use the underlying block device instead of the overlay file system for the container root file system. The block device maps to the top read-write layer for the overlay. This approach gives much better I/O performance compared to using 9pfs to share the container file system.
+
+The approach above does introduce a limitation in terms of dynamic file copy in/out of the container via `docker cp` operations.
+The copy operation from host to container accesses the mounted file system on the host side. This is not expected to work and may lead to inconsistencies as the block device will be simultaneously written to, from two different mounts.
+The copy operation from container to host will work, provided the user calls `sync(1)` prior to the copy to make sure any outstanding cached data is written to the block device.
```
jodh-intel: Where should the `sync` be called? In the VM, or on the host? Should the container be paused, then the `sync` call made I wonder?

hyperstart.go
jodh-intel: Is there an issue to track this? Not checking a return value for an operation like this sounds dangerous to me. I'd add a reference to an issue in a comment for this line.

README.md
amshinde: @jodh-intel I have added "from within the container" to make this clearer. And no, we do not need to pause the container.

jodh-intel:
lgtm

amshinde:
@jodh-intel I have created a separate issue for checking the return value of unmount: https://github.com/containers/virtcontainers/issues/295.
Please take a look.

jodh-intel:
Thanks @amshinde.

mount_test.go
```diff
@@ -41,3 +51,227 @@ func TestIsSystemMount(t *testing.T) {
...
+
+		data := bytes.Split(output, []byte(","))
+		if len(data) < 2 {
+			t.Fatal()
+		}
+
+		majorStr := strings.TrimSpace(string(data[0]))
+		minorStr := strings.TrimSpace(string(data[1]))
+
+		majorNo, err := strconv.Atoi(majorStr)
```
sboeuf: mount_test.go:75:12: ineffectual assignment to err

You should do something with err, or just use `_`

mount_test.go
```diff
@@ -41,3 +51,227 @@ func TestIsSystemMount(t *testing.T) {
...
+		minorNo, err := strconv.Atoi(minorStr)
```
sboeuf: mount_test.go:76:12: ineffectual assignment to err

You should do something with err, or just use `_`

container.go
```diff
@@ -238,13 +260,10 @@ func (c *Container) setContainerState(state stateString) error {
 		return errNeedState
 	}
 
-	// update in-memory state
```
amshinde: fixed

mount_test.go
amshinde: fixed

mount_test.go
amshinde: fixed

sboeuf:
LGTM

PR #289, opened 2017-06-23
container: Factorize container creation

sameo:
We were duplicating a bit of code, so we factorize it now.
Also, the single container creation case was wrong as it was
missing the container mount points.

### DIFF
--- a/container.go
+++ b/container.go
@@ -267,6 +267,7 @@ func (c *Container) createContainersDirs() error {
 	return nil
 }
 
+// newContainer creates a Container structure from a pod and a container configuration.
 func newContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 	if contConfig.valid() == false {
 		return &Container{}, fmt.Errorf("Invalid container configuration")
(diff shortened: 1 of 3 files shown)


container.go
```diff
@@ -351,6 +301,38 @@ func createContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 		c.mounts = mounts
 	}
 
+	return c, nil
+}
+
+func createContainers(pod *Pod, contConfigs []ContainerConfig) ([]*Container, error) {
```
jodh-intel: I wonder if this function should be renamed `createEmptyContainers()` as it is calling `newContainer()`, rather than `createContainer()`? Alternatively, maybe a comment on these functions would be useful to explain the differences.

sboeuf: why empty containers ? it calls into newContainer and it creates Container internal structures based on containerConfig, right ?

jodh-intel: Bad choice of words. The question boils down to: "why doesn't `createContainers()` call `createContainer()`?

jodh-intel:
lgtm

container.go
```diff
@@ -351,6 +302,45 @@ func createContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
...
+			return containers, err
+		}
+
+		containers = append(containers, c)
+	}
+
+	return containers, nil
+}
+
+// createContainer creates and start a container inside a Pod.
```
sboeuf: It does not start a container, it only creates it, right ?

sboeuf:
One comment about a comment, other than that:
LGTM

PR #291, opened 2017-06-26
pkg: oci: Rely on sandbox ID instead of sandbox name

sboeuf:
Due to recent changes in CRI-O database, we have to rely on the sandbox ID and not the sandbox name anymore.


sameo:
LGTM

PR #293, opened 2017-06-27
container: Add StateReady to StateStopped as valid transition

sboeuf:
This patch allows any container in StateReady state to be switched
to a StateStopped state, without doing further action than updating
the container state itself.


sameo:
LGTM

PR #296, opened 2017-06-28
pkg: hyperstart: Fix MTU type from string to integer

sboeuf:
The MTU parameter was working with hyperstart but it is only because
the cast worked. This patch makes sure that we are passing the MTU
as an integer as it is expected by hyperstart.


jodh-intel:
wow.

lgtm

sboeuf:
@sameo could you take a look please ?

sameo:
LGTM

PR #298, opened 2017-06-29
Remove Clear Containers specific kernel parameters

sameo:
Offload that to the runtime, who's now supposed to call `AddKernelParam`.


sboeuf:
LGTM

PR #299, opened 2017-07-05
hypervisor: Add Default{VCPUs, MemSz} to HypervisorConfig

dvoytik:
A runtime implementation can use these fields to setup the default
vCPU number and default memory size for new VMs.
VMConfig can overwrite these settings.

Signed-off-by: Dmitry Voytik <dmitry.voytik@huawei.com>

### DIFF
--- a/hypervisor.go
+++ b/hypervisor.go
@@ -31,6 +31,12 @@ const (
 	MockHypervisor HypervisorType = "mock"
 )
 
+const (
+	defaultVCPUs = 1
+	// 2 GiB
+	defaultMemSzMiB = 2048
+)
+
 // deviceType describes a virtualized device type.
 type deviceType int
 
@@ -128,6 +134,14 @@ type HypervisorConfig struct {
 	// Debug changes the default hypervisor and kernel parameters to
 	// enable debug output where available.
 	Debug bool
+
+	// DefaultVCPUs specifies default number of vCPUs for the VM.
+	// Pod configuration VMConfig.VCPUs overwrites this.
+	DefaultVCPUs uint32
+
+	// DefaultMem specifies default memory size in MiB for the VM.
+	// Pod configuration VMConfig.Memory overwrites this.
+	DefaultMemSz uint32
 }
 
 func (conf *HypervisorConfig) valid() (bool, error) {
@@ -137,6 +151,14 @@ func (conf *HypervisorConfig) valid() (bool, error) {
 
 	if conf.ImagePath == "" {
 		return false, fmt.Errorf("Missing image path")
+	}
+
+	if conf.DefaultVCPUs == 0 {
+		conf.DefaultVCPUs = defaultVCPUs
+	}
+
+	if conf.DefaultMemSz == 0 {
+		conf.DefaultMemSz = defaultMemSzMiB
 	}
 
 	return true, nil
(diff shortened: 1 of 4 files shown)


dvoytik:
Okay, I see the problem. I'm on it.

dvoytik:
I see. I have the same problem on my host. It seems like QEMU has an issue with less than 1 GiB guest physical memory size.

jodh-intel:
Hi @dvoytik - thanks for raising.

lgtm.

@dlespiau - could you take a look?

dvoytik:
Hi @dlespiau , are you OK with this patch?

dlespiau:
You have two things in this PR:

- Introducing the memory and vcpu configuration knobs (great)
- Changing the default from 2G to 512MB (debatable :)

I don't remember we have a great reason to default to 2G but I suspect quite a few images won't start with only 512MB of RAM. If you only introduce the first point in this PR (configuration), it can have my lgtm. Changing the default will need more work (we have a script to test the top 100 containers for instance, it probably has to go through it).

So maybe a good idea is to only have the configurability in this PR and merge it, with another PR to change the default if you so wish.

Of course, we can also specify 2G in the clear containers runtime, that should result in no effective change either.

dvoytik:
@dlespiau , I'm OK to keep it 2G. I'll update the PR in a minute.

hypervisor.go
```diff
@@ -139,6 +147,14 @@ func (conf *HypervisorConfig) valid() (bool, error) {
 		return false, fmt.Errorf("Missing image path")
 	}
 
+	if conf.DefaultVCPUs == 0 {
+		conf.DefaultVCPUs = 1
+	}
+
+	if conf.DefaultMemSz == 0 {
+		conf.DefaultMemSz = 2048 // MiB
```
jodh-intel: Nit: I'd be tempted to do something like:

```
// 2G
const defaultMemSzMiB = 2048

    :
conf.DefaultMemSz = defaultMemSzMiB
```

hypervisor_test.go
```diff
@@ -162,6 +162,26 @@ func TestHypervisorConfigIsValid(t *testing.T) {
...
+		HypervisorPath: "",
+	}
+	testHypervisorConfigValid(t, hypervisorConfig, true)
+
+	hypervisorConfigDefaultsExpected := &HypervisorConfig{
+		KernelPath:     fmt.Sprintf("%s/%s", testDir, testKernel),
+		ImagePath:      fmt.Sprintf("%s/%s", testDir, testImage),
+		HypervisorPath: "",
+		DefaultVCPUs:   1,
+		DefaultMemSz:   2048,
```
jodh-intel: Again, since this value is used here and in `qemu_test.go`, I'd define the following at the top of `virtcontainers_test.go` since that is where we place "shared" test variables:

```
const testDefaultMemSzMiB = 2048
```

hypervisor.go
dvoytik: Ok. Also did for conf.DefaultVCPUs

dlespiau:
lgtm

jodh-intel:
lgtm

dlespiau:
Alright, the owners are all on holidays, so merging this with two review. Note we've adding one more owner in https://github.com/containers/virtcontainers/pull/301 to deal with that situation. Could be a good idea to add james as well, potentially.

For now, and until #301 is merged, using superpowers to bypass pullapprove.

PR #304, opened 2017-07-12
shim: Stop forwarding output in detach mode

### LINKED ISSUE #303: Prevent the shim from forwarding outputs in case of "detach" mode
When we run a container with `cc-runtime run -d` or `cc-runtime exec -d`, we don't want to see the output posted on the current shell we are using. Instead, we want this silent.

sboeuf:
When a process is started as "detached" from one of the following
commands:
	- cc-runtime run -d <containerID>
	- cc-runtime exec -d <containerID>

we don't want to see the output posted on the current shell used,
instead we expect this to be silent.

Fixes #303

### DIFF
--- a/pkg/oci/utils_test.go
+++ b/pkg/oci/utils_test.go
@@ -147,7 +147,7 @@ func TestMinimalPodConfig(t *testing.T) {
 		t.Fatalf("Could not parse config.json: %v", err)
 	}
 
-	podConfig, err := PodConfig(ociSpec, runtimeConfig, tempBundlePath, containerID, consolePath)
+	podConfig, err := PodConfig(ociSpec, runtimeConfig, tempBundlePath, containerID, consolePath, false)
 	if err != nil {
 		t.Fatalf("Could not create Pod configuration %v", err)
 	}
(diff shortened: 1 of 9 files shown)


amshinde:
lgtm

amshinde:
lgtm

sboeuf:
@dlespiau could you review this one please.

pkg/oci/utils_test.go
```diff
@@ -147,7 +147,7 @@ func TestMinimalPodConfig(t *testing.T) {
 		t.Fatalf("Could not parse config.json: %v", err)
 	}
 
-	podConfig, err := PodConfig(ociSpec, runtimeConfig, tempBundlePath, containerID, consolePath)
+	podConfig, err := PodConfig(ociSpec, runtimeConfig, tempBundlePath, containerID, consolePath, false)
```
grahamwhaley: Should there be a test that checks that the for console redirection is working as defined by the detach parameter?

sboeuf:
@grahamwhaley I have added some tests, let me know if it sounds better to you.

grahamwhaley:
@sboeuf yep, I'm happy we now have tests.
I'll note that I'm not set up to actually run them at the moment (it looks like there needs to be some pre-test setup done that I'm missing), but still:
lgtm

amshinde:
lgtm

PR #310, opened 2017-07-17
create: Split hyperstart init function to perform setup only at create

amshinde:
The hyperstart agent implementation for init() interface was
performing setup such as adding devices with the hypervisor.
Since init is called during create, start and all other container
lifecycle stages, this setup was being performed repeatedly.

Introduce a new interface for agent called "createPod" that is
supposed to perform all the initial setup while creating a Pod.
Refactor hyperstart init() to exclude the device creation and move
that to createPod.

### DIFF
--- a/noop_agent.go
+++ b/noop_agent.go
@@ -30,6 +30,11 @@ func (n *noopAgent) init(pod *Pod, config interface{}) error {
 	return nil
 }
 
+// createPod is the Noop agent pod creation implementation. It does nothing.
+func (n *noopAgent) createPod(pod *Pod) error {
+	return nil
+}
+
 // capabilities returns empty capabilities, i.e no capabilties are supported.
 func (n *noopAgent) capabilities() capabilities {
 	return capabilities{}
--- a/pod.go
+++ b/pod.go
@@ -492,6 +492,40 @@ func (p *Pod) createSetStates() error {
 // to physically create that pod i.e. starts a VM for that pod to eventually
 // be started.
 func createPod(podConfig PodConfig) (*Pod, error) {
+	p, err := doFetchPod(podConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	state, err := p.storage.fetchPodState(p.id)
+	if err == nil && state.State != "" {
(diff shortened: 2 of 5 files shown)


jodh-intel:
The change looks ok, but I'm confused by the description: your new `agent.createPod()` is only called once, in the same function as `agent.init()` so I'm unclear how this is stopping `pod.hypervisor.addDevice()` being called multiple times?

amshinde:
Because of the somewhat not so evident  code here:
https://github.com/containers/virtcontainers/blob/master/pod.go#L561

If the state is not empty, the function returns. The state would be empty only at create time.
We do need to add some comments there.

pod.go
```diff
@@ -562,6 +562,10 @@ func createPod(podConfig PodConfig) (*Pod, error) {
 		return p, nil
 	}
```
sboeuf: Please add a comment here to specify this part is only accessed when `createPod(podConfig PodConfig)` is called by `CreatePod()` from api.go, which basically means this is only called when we create the pod.

pod.go
```diff
@@ -562,6 +562,10 @@ func createPod(podConfig PodConfig) (*Pod, error) {
...
 
+	if err := p.agent.createPod(p); err != nil {
```
sboeuf: This new statement makes the function reaching the `gocyclo=15` limit. Please split `createPod(podConfig PodConfig)` so that you fix this issue.

amshinde:
@sboeuf I have addressed your review comments, please have a look.

noop_agent.go
```diff
@@ -30,6 +30,11 @@ func (n *noopAgent) init(pod *Pod, config interface{}) error {
 	return nil
 }
 
+// createPod is the Noop agent command execution implementation. It does nothing.
```
sboeuf: The comment does not fit the function.

sboeuf:
@amshinde Please fix the comment, I'll merge after that.

amshinde:
@sboeuf fixed

sboeuf:
LGTM

PR #312, opened 2017-07-19
Enable properly gometalinter and fix found issues

sboeuf:
This PR takes care of fixing all the issues found by the gometalinter after we properly enabled it. It was enabled before this PR, but it was not taking into account the warnings that we want to care about.

### DIFF
--- a/.ci/go-lint.sh
+++ b/.ci/go-lint.sh
@@ -13,10 +13,7 @@
 # limitations under the License.
 #
 
-#!/usr/bin/env bash
-
-set -o errexit
-set -o nounset
+#!/bin/bash
 
 if [ ! $(command -v gometalinter) ]
 then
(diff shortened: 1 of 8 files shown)


.ci/go-lint.sh
```diff
@@ -25,16 +25,15 @@ then
...
-	--disable=aligncheck \
-	--disable=gotype \
-	--disable=gas \
-	--disable=vetshadow \
-	--cyclo-over=15 \
 	--tests \
-	--deadline=600s \
 	--vendor \
-	--errors \
+	--disable-all \
```
jodh-intel: This worries me slightly since as `gometalinter` grows more functionality over time, we won't be automatically taking advantage of it due to this flag. However, looking at the project docs it seems this is the preferred way to handle running it under a CI.

What we could do is something like:

```
    :

# When running the linters in a CI environment we need to disable them all by default
# and then explicitly enable the ones we are care about. This is necessary since *iff* gometalinter
# adds a new linter, that linter may cause the CI build to fail when it really shouldn't. However, when
# this script is run locally, all linters should be run to allow the developer to review any failures
# (and potentially decide whether we need to explicitly enable a new linter in the CI).
if [ "$CI" = true ]; then
    linter_args+=" --disable-all"
fi

linter_args+=" --enable=misspell"
linter_args+=" --enable=vet"

    :
```

That would DTRT under CI, but would run all linters I think(?) for devs calling `.ci/go-lint.sh` manually as we should all be doing before raising a PR. Yes, if new linters get added they might error locally, but that is a signal the code needs improving (and that we need to potentially explicitly enable that new linter in `.ci/go-lint.sh`).

grahamwhaley:
For other related repos we recently decided to use `the go-static-checks.sh` such as https://github.com/clearcontainers/runtime/blob/master/.ci/go-static-checks.sh as I believe the go-lint.sh was showing some timeouts on some of the CIs as it takes a long time to do the download/build of the tools.
Thus, I'm prone to say let's go with `go-static-checks.sh` for this repo unless we know otherwise that `go-lint.sh` is stable. /cc @dlespiau for input.

dlespiau:
That's right, gometalinter is timing out on the proxy sometimes. Not sure what I can do about this.

.ci/go-lint.sh
sboeuf: @jodh-intel What's wrong with `--disable-all` ? That means that we have to enable every linter on purpose. I think this is good to know which linter we enabled and what we could expect. Otherwise, they enable some fancy linters and we get a ton of outputs...

jodh-intel: Yes, but if we disable all linters and then only enable what we want (and crucially know about), how will we know when new linters (that we may wish to use) become available? My suggested change will allow developers locally to become aware of new linters. Yes, the lint may fail but that's what we want as it will trigger the question in the developers mind, "do we need to enable this new linter for the CI?". And if they frankly don't care, they can simply do:

```
(CI=true .ci/go-lint.sh) 
```

sboeuf: Ok sounds good I will do that.

jodh-intel:
lgtm

sboeuf:
@dlespiau @amshinde could you please take a look

amshinde:
lgtm

PR #317, opened 2017-07-20
Increase hook test coverage

jodh-intel:


hook/mock/hook.go
```diff
@@ -85,12 +85,17 @@ func main() {
...
-	if len(os.Args) == 4 {
+	if len(os.Args) >= 4 {
 		timeout, err := strconv.Atoi(os.Args[3])
 		if err != nil {
 			fmt.Fprintf(f, "Could not retrieve timeout %s from args[3]\n", os.Args[3])
 			os.Exit(1)
 		}
 		time.Sleep(time.Duration(timeout) * time.Second)
+		if len(os.Args) == 5 {
+			msg := "panicing at user request"
```
sboeuf: Should not this be "panicking" instead of "panicing" ?

hook_test.go
```diff
@@ -108,6 +120,47 @@ func TestRunHookTimeoutFailure(t *testing.T) {
...
+
+func testRunHookInvalidCommand(t *testing.T, timeout int) {
+	cleanUp()
+
+	dir, err := ioutil.TempDir("", "")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	defer os.RemoveAll(dir)
```
sboeuf: This won't be called because if we exit on a t.Fatal(), this will cause to exit directly, without going through the defer stack.

hook_test.go
```diff
@@ -108,6 +120,47 @@ func TestRunHookTimeoutFailure(t *testing.T) {
...
+
+	cmd := filepath.Join(dir, "does-not-exist")
+
+	savedDefaultMockHookBinPath := DefaultMockHookBinPath
+	DefaultMockHookBinPath = cmd
+
+	defer func() {
```
sboeuf: Ditto.

hook/mock/hook.go
jodh-intel: Good catch - fixing...

hook_test.go
jodh-intel: Correct, and we don't want it to as it might contain useful info to debug the failure.

sboeuf:
LGTM

PR #319, opened 2017-07-20
stop: Wait for end of container process before to remove

### LINKED ISSUE #318: Make sure the container process is really stopped before trying to remove the container
In case a container or a pod needs to be stopped, virtcontainers library first kill the process and then remove the container. But we should not try to remove the container until we are sure the process inside is not running anymore.
We should check if the shim is still running, and until we can see it running, we should not call into removeContainer.

sboeuf:
In case we are stopping a pod or a container, for every container,
we send a SIGKILL signal and then we try to remove the container.
Unfortunately, if the process inside the container takes some time
to be killed, the removal of the container can sometimes fail.

This patch ensures that we are waiting for the end of the shim
process after we sent the SIGKILL signal. In case we hit a timeout,
the stop() will fail.

Fixes #318

### DIFF
--- a/container.go
+++ b/container.go
@@ -508,6 +508,11 @@ func (c *Container) stop() error {
 		return err
 	}
 
+	// Wait for the end of container
+	if err := waitForShim(c.process.Pid); err != nil {
+		return err
+	}
+
 	err = c.pod.agent.stopContainer(*(c.pod), *c)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


sboeuf:
@amshinde please could you review this one.

amshinde:
lgtm

PR #321, opened 2017-07-20
qemu: Add pc machine type as supported

### LINKED ISSUE #320: Add QemuPC as a new machine type supported by our qemu hypervisor implementation
We will probably need to switch to the `pc` machine type in the future. Let's add this new type as a supported one in virtcontainers.

sboeuf:
We might need to switch to the pc machine type in the future. This
commits takes care of adding this new type as supported.

Fixes #320

### DIFF
--- a/qemu.go
+++ b/qemu.go
@@ -60,6 +60,9 @@ const (
 	// QemuPCLite is the QEMU pc-lite machine type
 	QemuPCLite = defaultQemuMachineType
 
+	// QemuPC is the QEMU pc machine type
+	QemuPC = "pc"
+
 	// QemuQ35 is the QEMU Q35 machine type
 	QemuQ35 = "q35"
 )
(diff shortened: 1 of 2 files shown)


amshinde:
lgtm

qemu.go
```diff
@@ -60,13 +60,17 @@ const (
...
+
 	// QemuQ35 is the QEMU Q35 machine type
 	QemuQ35 = "q35"
 )
 
 // Mapping between machine types and QEMU binary paths.
 var qemuPaths = map[string]string{
 	QemuPCLite: "/usr/bin/qemu-lite-system-x86_64",
+	QemuPC:     defaultQemuPath,
 	QemuQ35:    defaultQemuPath,
```
mcastelino: The q35 path will also be a custom path till all out patches get upstream

sboeuf: @mcastelino is that better ?
```go
QemuPC: "/usr/bin/qemu-q35-system-x86_64",
```

mcastelino: @sboeuf that works.

PR #322, opened 2017-07-25
qemu: fix breakage of old binaries

wcwxyz:
clear container kernel 4.5-50 breaks old binaries, for example centos6.
Fix this by using vsyscall=emulate

### DIFF
--- a/qemu.go
+++ b/qemu.go
@@ -148,6 +148,7 @@ var kernelDefaultParams = []Param{
 	{"iommu", "off"},
 	{"cryptomgr.notests", ""},
 	{"net.ifnames", "0"},
+	{"vsyscall", "emulate"},
 }
 
 // kernelDefaultParamsNonDebug is a list of the default kernel
--- a/qemu_test.go
+++ b/qemu_test.go
@@ -57,7 +57,7 @@ func testQemuBuildKernelParams(t *testing.T, kernelParams []Param, expected stri
 	}
 }
 
-var testQemuKernelParamsBase = "root=/dev/pmem0p1 rootflags=dax,data=ordered,errors=remount-ro rw rootfstype=ext4 tsc=reliable no_timer_check rcupdate.rcu_expedited=1 i8042.direct=1 i8042.dumbkbd=1 i8042.nopnp=1 i8042.noaux=1 noreplace-smp reboot=k panic=1 console=hvc0 console=hvc1 initcall_debug iommu=off cryptomgr.notests net.ifnames=0"
+var testQemuKernelParamsBase = "root=/dev/pmem0p1 rootflags=dax,data=ordered,errors=remount-ro rw rootfstype=ext4 tsc=reliable no_timer_check rcupdate.rcu_expedited=1 i8042.direct=1 i8042.dumbkbd=1 i8042.nopnp=1 i8042.noaux=1 noreplace-smp reboot=k panic=1 console=hvc0 console=hvc1 initcall_debug iommu=off cryptomgr.notests net.ifnames=0 vsyscall=emulate"
 var testQemuKernelParamsNonDebug = "quiet systemd.show_status=false"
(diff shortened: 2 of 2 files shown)


grahamwhaley:
For reference for folks the info from the kernel is here:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/admin-guide/kernel-parameters.txt#n4387

The theory looks OK to me. I see that `emulate` is the practical and safer method of enabling this.
Would you happen to have an easy example to demonstrate the failure, and maybe some example output of what you see when it fails - if that is appropriate?
I've added @sboeuf to review. @sameo is not here to review at the minute, so asking @dlespiau for his thoughts as well.

wcwxyz:
Running centos:6 /bin/bash would a segment fault:

[    3.940412] bash[222] vsyscall attempted with vsyscall=none ip:ffffffffff600400 cs:33 sp:7fff828c0a88 ax:ffffffffff600400 si:7fff828c0f70 di:0
[    3.940422] bash[222]: segfault at ffffffffff600400 ip ffffffffff600400 sp 00007fff828c0a88 error 15

kernel: /usr/share/clear-containers/vmlinux-4.5-50.container
image: /usr/share/clear-containers/clear-16410-containers.img

sboeuf:
Reading from the kernel info, I have two questions:
- I don't understand why we have to enable vsyscall=emulate since emulate is designated as the default one.
- Docs mentions that `emulate`will make some exploits possible (even if hard) while no exploits are possible with `none` (but breakage possible, and I think that's our case, right ?).

The big question is, do we want to support old binaries like the ones we can find inside centos6 ? I don't think we want to support centos 6 anyway. Our centos target is centos 7.
@grahamwhaley can you give some feedback here ?

sboeuf:
@wcwxyz we don't want to add such a parameter hardcoded into virtcontainers, but what you could do is add extra kernel parameters through the runtime.
Right now, the way you can do that is to add one more entry here: https://github.com/clearcontainers/runtime/blob/3.0.0-alpha.4/create.go#L149
```go
ccKernelParams := []vc.Param{
		{
			Key:   "vsyscall",
			Value: "emulate",
		},
                ...
}
```
And then you will need to rebuild the runtime.

Another way would be for you to submit a PR to clearcontainers/runtime, by adding a configuration field to our config file `configuration.toml` and parse/add the kernel parameters you would have passed:
```bash
[hypervisor.qemu]
extra_kernel_params = "vsyscall=emulate foo1=bar1 foo2=bar2"
...
```

sboeuf:
-1

grahamwhaley:
So, with some more info and retrospect, I agree with @sboeuf - the right way to do this is not to hard wire it here, but to have a mechanism where such things can optionally be added in. That is just going to be a better more scale-able way forwards, and it avoids us adding the extra bit of attack surface to the default runtime when there is generally not a need.
For more reference, it looks like our setting for the vsyscalls in our default kernel config is indeed `NONE`:
https://github.com/clearcontainers/linux/blob/v4.9.34-clear-containers/arch/x86/configs/clear_containers_defconfig#L500
and hence the need to enable this on the command line if needed.
I'll stick a 'do not merge' on this to avoid any accidental button presses.

wcwxyz:
By centos6, I mean the docker image centos:6. You can't change how distros build their binaries. 

I think a generic runtime should be able to run every docker image **by default**. For those who want to reduce attack surface they can add "vsyscall=none" by method @sboeuf mentioned above.

Anyway I'm happy to submit a pr to support passing extra kernel args from configuration file. But believe me, centos6 is old, yet still very popular.

What do you think?

wcwxyz:
If you guys are looking for improving kernel security, here's a good start:
https://kernsec.org/wiki/index.php/Kernel_Self_Protection_Project/Recommended_Settings

wcwxyz:
This issue is address via

https://github.com/containers/virtcontainers/pull/326
https://github.com/clearcontainers/runtime/pull/369

PR #324, opened 2017-07-25
container: Kill the shim when trying to stop a "ready" container

### LINKED ISSUE #323: Kill the shim when trying to stop/kill a container in "ready" state
A container in "ready" state means that nothing is running inside the VM. The only thing to do in this case is to make sure the shim is properly stopped.

sboeuf:
A container in "ready" state means that nothing is running inside
the VM. The only thing to do in this case is to make sure the shim
is properly stopped.

Fixes #323


sboeuf:
@amshinde please take a look.

container.go
```diff
@@ -481,41 +481,40 @@ func (c *Container) stop() error {
 	// In case our container is "ready", there is no point in trying to
 	// stop it because nothing has been started. However, this is a valid
 	// case and we handle this by updating the container state only.
```
amshinde: This comment no longer holds true. Can you update it.

container.go
```diff
@@ -481,41 +481,40 @@ func (c *Container) stop() error {
...
-		if err := state.validTransition(StateReady, StateStopped); err != nil {
-			return err
-		}
-
-		return c.setContainerState(StateStopped)
-	}
 
-	if state.State != StateRunning {
+	if state.State != StateRunning && state.State != StateReady {
 		return fmt.Errorf("Container not running, impossible to stop")
```
amshinde: This error message needs to be updated as well.

container.go
```diff
@@ -481,41 +481,40 @@ func (c *Container) stop() error {
...
+		defer c.pod.proxy.disconnect()
 
-	err = c.pod.agent.killContainer(*(c.pod), *c, syscall.SIGKILL, true)
-	if err != nil {
-		return err
+		if err := c.pod.agent.killContainer(*(c.pod), *c, syscall.SIGKILL, true); err != nil {
+			return err
+		}
+	} else if state.State == StateReady {
+		if err := stopShim(c.process.Pid); err != nil {
```
amshinde: Can you add some comments regarding the ready case explaining the flow from shim to proxy, and proxy sending the return code here. Will be easier to follow in the future.

container.go
sboeuf: done

container.go
sboeuf: done

container.go
sboeuf: done

jodh-intel:
This change seems fine, but can you point me at the tests that assert the new `cc-shim`-killing behaviour?

sboeuf:
@jodh-intel what do you mean ?

sboeuf:
Here is the issue to create the corresponding test: https://github.com/clearcontainers/tests/issues/281.

jodh-intel:
@sboeuf - I mean, have you written any tests to assert this new behaviour (in a different repo maybe?)? If not, please can you add some.

PR #325, opened 2017-07-29
Drive hotplug support

amshinde:


capabilities.go
```diff
@@ -34,3 +35,14 @@ func (caps *capabilities) isBlockDeviceSupported() bool {
...
+
+func (caps *capabilities) isBlockDeviceHotplugSupported() bool {
+	if caps.flags&blockDeviceHotplugSupport != 0 {
+		return true
+	}
+	return false
+}
+
+func (caps *capabilities) setBlockDeviceHotplugSupport() {
+	caps.flags = caps.flags | blockDeviceHotplugSupport
```
jodh-intel: I guess it's in keeping with the existing code, but I personally find `|=` more intuitive (plus it's less error prone as you only specify the variable once).

container.go
```diff
@@ -606,3 +620,67 @@ func newProcess(token string, pid int) Process {
...
+	dev, err := getDeviceForPath(c.rootFs)
+	if err != nil && err != errMountPointNotFound {
+		return err
+	}
+
+	if err == errMountPointNotFound {
+		return nil
+	}
+
+	virtLog.Infof("Device details for container %s: Major:%d, Minor:%d, MountPoint:%s\n", c.id, dev.major, dev.minor, dev.mountPoint)
```
jodh-intel: I don't think the `\n` is required for these log calls.

qemu.go
```diff
@@ -662,6 +675,64 @@ func (q *qemu) togglePausePod(pause bool) error {
...
+	return nil
+}
+
+func (q *qemu) hotplugDevice(devInfo interface{}, devType deviceType) error {
+	switch devType {
+	case blockDev:
+		drive := devInfo.(Drive)
+		return q.hotplugBlockDevice(drive)
+	default:
+		return fmt.Errorf("Only hotplug for block devices supported for now")
```
jodh-intel: Can you log the `devType` here.

amshinde:
@jodh-intel I have pushed fixes addressing your comments.

qemu.go
```diff
@@ -662,6 +675,64 @@ func (q *qemu) togglePausePod(pause bool) error {
...
+	q.qmpMonitorCh.qmp = qmp
+
+	if err := q.qmpMonitorCh.qmp.ExecuteBlockdevAdd(q.qmpMonitorCh.ctx, drive.File, drive.ID); err != nil {
+		return err
+	}
+
+	devID := "virtio-" + drive.ID
+	driver := "virtio-blk-pci"
+	if err := q.qmpMonitorCh.qmp.ExecuteDeviceAdd(q.qmpMonitorCh.ctx, drive.ID, devID, driver, ""); err != nil {
+		return err
```
jodh-intel: Should the error path "undo" the block device add?

container.go
```diff
@@ -456,6 +456,20 @@ func (c *Container) start() error {
 	}
 	defer c.pod.proxy.disconnect()
 
+	// Try to check for block storage if Fstype is empty
+	// If Fstype has been already assigned, then block device has been checked and assigned earlier
```
jodh-intel: another missing period.

container.go
```diff
@@ -456,6 +456,20 @@ func (c *Container) start() error {
 	}
 	defer c.pod.proxy.disconnect()
 
+	// Try to check for block storage if Fstype is empty
```
jodh-intel: Missing period at end of sentence.

hypervisor.go
```diff
@@ -211,5 +211,7 @@ type hypervisor interface {
 	pausePod() error
 	resumePod() error
 	addDevice(devInfo interface{}, devType deviceType) error
+	hotplugDevice(devInfo interface{}, devType deviceType) error
```
jodh-intel: The obvious question is: "where is the corresponding `hotplugRemoveDevice`?". ciao supports it, so shouldn't we? If so, this should be named something like `hotplugAddDevice` instead.

capabilities.go
```diff
@@ -34,3 +35,14 @@ func (caps *capabilities) isBlockDeviceSupported() bool {
 func (caps *capabilities) setBlockDeviceSupport() {
 	caps.flags = caps.flags | blockDeviceSupport
 }
+
+func (caps *capabilities) isBlockDeviceHotplugSupported() bool {
```
jodh-intel: This (and the `qemu.go` changes as a minimum) need corresponding unit tests.

qemu.go
```diff
@@ -662,6 +675,64 @@ func (q *qemu) togglePausePod(pause bool) error {
...
+	}
+
+	return nil
+}
+
+func (q *qemu) hotplugDevice(devInfo interface{}, devType deviceType) error {
+	switch devType {
+	case blockDev:
+		drive := devInfo.(Drive)
```
jodh-intel: This needs to use the 2 arg type assertion syntax to avoid a potential runtime panic.

amshinde: if the interface does hold a drive pointer, then this would indicate an error in the code path.
I would prefer to have the program panic, rather than handle this.

egernst:
For clarity, without this PR you cannot run K8S when using devicemapper.

container.go
amshinde: Fixed

container.go
amshinde: Fixed.

hypervisor.go
amshinde: Addresses this one as well. hotplugRemoveDevice is now called while stopping a container.

capabilities.go
amshinde: I have tests for capabilities and qemu capabilities. I am planning to add tests for the actual qmp commands by starting a an actual qemu instance. I'll add those tests in a separate PR.

container.go
```diff
@@ -456,6 +456,20 @@ func (c *Container) start() error {
 	}
 	defer c.pod.proxy.disconnect()
 
+	// Try to check for block storage if Fstype is empty.
+	// If Fstype has been already assigned, then block device has been checked and assigned earlier.
+	// Worst case, we perform the check twice for non-block storage.
+	if c.state.Fstype == "" {
```
sboeuf: I get that you use this `string` as a flag, but it might be better to try to improve this for more readability. Maybe introducing a `bool` in order to indicate this container has already been checked/processed.

container.go
```diff
@@ -715,3 +719,21 @@ func (c *Container) addDrive(create bool) (err error) {
 
 	return nil
 }
+
+func (c *Container) removeDrive() (err error) {
+	if c.state.Fstype != "" {
+		virtLog.Infof("Unplugging block device for container %s\n", c.id)
+
+		// Add drive with id as container id
```
sboeuf: Comment is not appropriate since you're removing the drive here.

container.go
```diff
@@ -715,3 +719,21 @@ func (c *Container) addDrive(create bool) (err error) {
 
 	return nil
 }
+
+func (c *Container) removeDrive() (err error) {
+	if c.state.Fstype != "" {
```
sboeuf: Once again, a boolean with a different name (something like `pluggedDrive` maybe) would be much clearer to understand the code here.

pod.go
```diff
@@ -64,6 +64,9 @@ type State struct {
 
 	// File system of the rootfs incase it is block device
 	Fstype string `json:"fstype"`
+
+	// Bool to indicate if the drive for a container was hotplugged.
+	IsDriveHotplugged bool `json:"isDriveHotplugged"`
```
sboeuf: This being a variable and not a function, I would call it `HotpluggedDrive` instead.

qemu.go
```diff
@@ -698,6 +705,83 @@ func (q *qemu) togglePausePod(pause bool) error {
 	return nil
 }
 
+func (q *qemu) qmpSetup() (*ciaoQemu.QMP, error) {
```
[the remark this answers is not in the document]
sboeuf: Sounds good !

container.go
amshinde: A value in Fstype means that drive is plugged, I was trying to avoid adding an extra state variable. Probably a container function isDrivePugged() which checks for fstype and using that function instead would make the code clearer. What do you think?

sboeuf: Yes good idea. This function would hide the "complexity" and would help making things clearer.

qemu.go
amshinde: This is a programming error rather than a runtime error. Better have the code panic with the type assertion failure which will be quite evident from the backtrace.

capabilities.go
```diff
@@ -18,6 +18,7 @@ package virtcontainers
```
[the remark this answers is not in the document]
amshinde: fixed

pod.go
amshinde: Fixed

qemu.go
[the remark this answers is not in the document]
amshinde: done

qemu.go
```diff
@@ -197,6 +197,19 @@ func (q *qemu) buildKernelParams(config HypervisorConfig) error {
 	return nil
 }
 
+func (q *qemu) capabilities() capabilities {
+	var caps capabilities
+
+	// add all capabilities supported by hypervisor
```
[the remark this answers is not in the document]
amshinde: done

container.go
amshinde: fixed

sboeuf:
LGTM

amshinde:
@jodh-intel Can you take a look at this?

jodh-intel:
"approved", but I'd really like to see more tests added for this feature as this PR is lowering the coverage.

sameo:
LGTM

PR #326, opened 2017-08-01
hypervisor: export SerializeParams and DeserializeParams

wcwxyz:
Add an exported function DeserializeParams(), and export
SerializeParams().

This patch is laying the ground work for [https://github.com/clearcontainers/runtime/pull/369](url)

### DIFF
--- a/hypervisor.go
+++ b/hypervisor.go
@@ -18,6 +18,7 @@ package virtcontainers
 
 import (
 	"fmt"
+	"strings"
 )
 
 // HypervisorType describes an hypervisor type.
@@ -180,7 +181,8 @@ func appendParam(params []Param, parameter string, value string) []Param {
 	return append(params, Param{parameter, value})
 }
 
-func serializeParams(params []Param, delim string) []string {
+// SerializeParams converts []Param to []string
+func SerializeParams(params []Param, delim string) []string {
 	var parameters []string
 
 	for _, p := range params {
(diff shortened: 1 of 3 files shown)


hypervisor.go
```diff
@@ -201,6 +203,25 @@ func serializeParams(params []Param, delim string) []string {
...
+	for _, param := range parameters {
+		if param == "" {
+			continue
+		}
+		p := strings.SplitN(param, "=", 2)
+		if len(p) == 2 {
+			params = append(params, Param{Key: p[0], Value: p[1]})
+		} else if len(p) == 1 {
+			params = append(params, Param{Key: p[0], Value: ""})
+		}
```
sboeuf: You should put an else statement giving us some logs:
```go
virtLog.Warnf("Invalid parameter %s", param)
```

wcwxyz: Thanks for review.

After strings.SplitN(param, "=", 2), len(p) is either 1 or 2. So another else statement seems like unnecessary. Correct me if I'm wrong.

sboeuf: Yes you're right, then in that case, use:
```go
else
```
for the second case instead of
```go
else if len(p) == 1
```

wcwxyz: ok, I'll update.

sboeuf:
@wcwxyz Nice thanks for the changes. The PR is ready to be merged after we got confirmation from the CI.
LGTM

PR #333, opened 2017-08-01
Globally improves the handling of container/pod states

### LINKED ISSUE #327: Update the container status if the container's process has terminated
We don't want to provide a wrong info about our containers states when a consumer of virtcontainers retrieve info related to the container. We want virtcontainers to properly stop the container if the process has been terminated.

### LINKED ISSUE #328: Send a signal to a non-running container is valid
We consider sending a signal to a `ready` container as a valid case. In case the signal is a non-killing signal (different from SIGKILL and SIGTERM), this results in a no-op. Otherwise we kill the shim because we know there is no container started inside the VM, and we set the container state to stopped.
The container `stop()` function should be free from this specific case because everything is handled in `kill()`.

### LINKED ISSUE #329: Don't issue an error if trying to stop a container already stopped
Because we now have virtcontainers updating the container state when someone is checking for the container state, a container can be stopped because the process inside the container has terminated. If someone tries to stop a container that has been previously terminated should not generate an error in this case.

### LINKED ISSUE #330: Factorize pod stop() to rely on container stop()
We want to avoid any code duplication by calling
```go
func (c *Container) stop()
```
from
```go
func (p *Pod) stop()
```

### LINKED ISSUE #331: Ignore stopped or ready containers when stopping the pod
When the pod is stopped, we expect to see all containers ending up to `stopped` or `ready`/`created` state. For those reasons, we don't need to stop the containers already in those states.

### LINKED ISSUE #332: Allow to stop a non-running pod
Because we allowed to kill a container process even if the container is not running, we have to allow the same thing at the pod level, otherwise we run into some issues saying that we cannot stop a pod which is not running.

sboeuf:
This PR fixes several small issues but all related as they globally fix/improve the way we handle a container (or pod) stop/kill and also the updates of the container state. This is needed for reworking properly our runtime implementation.

Fixes #327 
Fixes #328 
Fixes #329 
Fixes #330 
Fixes #331 
Fixes #332

### DIFF
--- a/api.go
+++ b/api.go
@@ -578,6 +578,24 @@ func StatusContainer(podID, containerID string) (ContainerStatus, error) {
 func statusContainer(pod *Pod, containerID string) (ContainerStatus, error) {
 	for _, container := range pod.containers {
 		if container.id == containerID {
+			// We have to check for the process state to make sure
+			// we update the status in case the process is supposed
+			// to be running but has been killed or terminated.
+			if (container.state.State == StateRunning ||
+				container.state.State == StatePaused) &&
+				container.process.Pid > 0 {
+				running, err := isShimRunning(container.process.Pid)
+				if err != nil {
+					return ContainerStatus{}, err
+				}
+
+				if !running {
+					if err := container.stop(); err != nil {
+						return ContainerStatus{}, err
+					}
+				}
+			}
+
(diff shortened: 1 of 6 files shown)


amshinde:
lgtm

PR #336, opened 2017-08-09
[RFC] qemu: set maxmem to host memory size

wcwxyz:
Setting maxmem to host memory size wouldn't change memory footprint. It
only defines the upper bounds for memory hotplug.

Currently we set maxmem to mem*1.5, which is not enough for running a
small VM (eg. 128M) considering nvdimm device consumes memory address
space.

This will also allow us to do memory hotplug (maybe in the future).

### DIFF
--- a/qemu.go
+++ b/qemu.go
@@ -106,6 +106,12 @@ const (
 	maxDevIDSize = 31
 )
 
+const (
+	// NVDIMM device needs memory space 1024MB
+	// See https://github.com/clearcontainers/runtime/issues/380
+	maxMemoryOffset = 1024
+)
+
 type qmpLogger struct{}
 
 func (l qmpLogger) V(level int32) bool {
@@ -468,13 +474,20 @@ func (q *qemu) setCPUResources(podConfig PodConfig) ciaoQemu.SMP {
 	return smp
 }
 
-func (q *qemu) setMemoryResources(podConfig PodConfig) ciaoQemu.Memory {
+func (q *qemu) setMemoryResources(podConfig PodConfig) (ciaoQemu.Memory, error) {
+	hostMemKb, err := getHostMemorySizeKb(procMemInfo)
+	if err != nil {
+		return ciaoQemu.Memory{}, fmt.Errorf("Unable to read memory info: %s", err)
+	}
+	if hostMemKb == 0 {
+		return ciaoQemu.Memory{}, fmt.Errorf("Error host memory size 0")
+	}
+
+	// add 1G memory space for nvdimm device (vm guest image)
+	memMax := fmt.Sprintf("%dM", int(float64(hostMemKb)/1024)+maxMemoryOffset)
 	mem := fmt.Sprintf("%dM", q.config.DefaultMemSz)
-	memMax := fmt.Sprintf("%dM", int(float64(q.config.DefaultMemSz)*1.5))
 	if podConfig.VMConfig.Memory > 0 {
 		mem = fmt.Sprintf("%dM", podConfig.VMConfig.Memory)
-		intMemMax := int(float64(podConfig.VMConfig.Memory) * 1.5)
-		memMax = fmt.Sprintf("%dM", intMemMax)
 	}
 
 	memory := ciaoQemu.Memory{
@@ -483,7 +496,7 @@ func (q *qemu) setMemoryResources(podConfig PodConfig) ciaoQemu.Memory {
 		MaxMem: memMax,
 	}
 
-	return memory
+	return memory, nil
 }
 
 // createPod is the Hypervisor pod creation implementation for ciaoQemu.
@@ -502,7 +515,10 @@ func (q *qemu) createPod(podConfig PodConfig) error {
 
 	smp := q.setCPUResources(podConfig)
 
-	memory := q.setMemoryResources(podConfig)
+	memory, err := q.setMemoryResources(podConfig)
+	if err != nil {
+		return err
+	}
 
 	knobs := ciaoQemu.Knobs{
 		NoUserConfig: true,
--- a/qemu_test.go
+++ b/qemu_test.go
@@ -395,10 +395,16 @@ func TestQemuSetMemoryResources(t *testing.T) {
 
 	q := &qemu{}
 
+	hostMemKb, err := getHostMemorySizeKb(procMemInfo)
+	if err != nil {
+		t.Fatal(err)
+	}
+	memMax := fmt.Sprintf("%dM", int(float64(hostMemKb)/1024)+maxMemoryOffset)
+
 	expectedOut := ciaoQemu.Memory{
 		Size:   "1000M",
 		Slots:  uint8(2),
-		MaxMem: "1500M",
+		MaxMem: memMax,
 	}
 
 	vmConfig := Resources{
@@ -409,7 +415,10 @@ func TestQemuSetMemoryResources(t *testing.T) {
 		VMConfig: vmConfig,
 	}
 
-	memory := q.setMemoryResources(podConfig)
+	memory, err := q.setMemoryResources(podConfig)
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	if reflect.DeepEqual(memory, expectedOut) == false {
 		t.Fatalf("Got %v\nExpecting %v", memory, expectedOut)
--- a/hypervisor.go
+++ b/hypervisor.go
@@ -17,7 +17,10 @@
 package virtcontainers
 
 import (
+	"bufio"
 	"fmt"
+	"os"
+	"strconv"
 	"strings"
 )
 
@@ -31,6 +34,8 @@ const (
 	// MockHypervisor is a mock hypervisor for testing purposes
 	MockHypervisor HypervisorType = "mock"
 )
+
+const procMemInfo = "/proc/meminfo"
 
 const (
 	defaultVCPUs = 1
@@ -233,4 +238,37 @@ type hypervisor interface {
 	resumePod() error
 	addDevice(devInfo interface{}, devType deviceType) error
 	getPodConsole(podID string) string
+}
+
+func getHostMemorySizeKb(memInfoPath string) (uint64, error) {
+	f, err := os.Open(memInfoPath)
+	if err != nil {
+		return 0, err
+	}
+	defer f.Close()
+
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		// Expected format: ["MemTotal:", "1234", "kB"]
+		parts := strings.Fields(scanner.Text())
+
+		// Sanity checks: Skip malformed entries.
+		if len(parts) < 3 || parts[0] != "MemTotal:" || parts[2] != "kB" {
+			continue
+		}
+
+		sizeKb, err := strconv.ParseUint(parts[1], 0, 64)
+		if err != nil {
+			continue
+		}
+
+		return sizeKb, nil
+	}
+
+	// Handle errors that may have occurred during the reading of the file.
+	if err := scanner.Err(); err != nil {
+		return 0, err
+	}
+
+	return 0, fmt.Errorf("unable get MemTotal from %s", memInfoPath)
 }
--- a/hypervisor_test.go
+++ b/hypervisor_test.go
@@ -18,6 +18,9 @@ package virtcontainers
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"reflect"
 	"testing"
 )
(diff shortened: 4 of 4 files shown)


wcwxyz:
CI failed because I didn't vendor `github.com/moby/moby/pkg/system`.

Can you guys please show me how to vendor a new go project using `dep`?

sboeuf:
Hi @wcwxyz, thanks again for this new patch :)
I have just looked at it, and I would prefer that you avoid to import this moby package since what we need is very simple. Moreover, we are not trying to support Windows here, so you can directly create our own function to read from `/proc/meminfo`. This way, your problems are solved ;)

wcwxyz:
I've updated the patch. I noticed coverage drop down a bit. That's because some error case is not tested (but should be ok).

qemu.go
```diff
@@ -468,13 +470,16 @@ func (q *qemu) setCPUResources(podConfig PodConfig) ciaoQemu.SMP {
...
 }
 
-func (q *qemu) setMemoryResources(podConfig PodConfig) ciaoQemu.Memory {
+func (q *qemu) setMemoryResources(podConfig PodConfig) (ciaoQemu.Memory, error) {
+	hostMem, err := getHostMemorySize()
+	if err != nil {
+		return ciaoQemu.Memory{}, fmt.Errorf("Unable to read memory info: %s", err)
+	}
+	// add 1G memory space for nvdimm device (vm guest image)
+	memMax := fmt.Sprintf("%dM", int(float64(hostMem)/1024/1024)+1024)
```
grahamwhaley: in the spirit of @jodh-intel , it might be nice if the variable names reflected the units they were holding, maybe like `memMaxMegs` or similar. not a blocker :-)

grahamwhaley:
lgtm
let's add a:
Fixes #337

qemu.go
```diff
@@ -483,7 +488,42 @@ func (q *qemu) setMemoryResources(podConfig PodConfig) ciaoQemu.Memory {
 		MaxMem: memMax,
 	}
 
-	return memory
+	return memory, nil
+}
+
+func getHostMemorySize() (uint64, error) {
```
jodh-intel: I'd prefer this was called `getHostMemorySizeBytes()` to avoid any confusion.

qemu.go
```diff
@@ -483,7 +488,42 @@ func (q *qemu) setMemoryResources(podConfig PodConfig) ciaoQemu.Memory {
...
+	f, err := os.Open("/proc/meminfo")
```
jodh-intel: This function really need a test so I'd either pass this filename as an argument, or use a global that the tests can modify. For an almost identical example (feel free to copy):

- https://github.com/clearcontainers/runtime/blob/master/cc-check.go#L85..L99
- https://github.com/clearcontainers/runtime/blob/master/cc-check_test.go#L45..L90

qemu.go
```diff
@@ -483,7 +488,42 @@ func (q *qemu) setMemoryResources(podConfig PodConfig) ciaoQemu.Memory {
...
+
+		return bytes, nil
+	}
+
+	// Handle errors that may have occurred during the reading of the file.
+	if err := scanner.Err(); err != nil {
+		return 0, err
+	}
+
+	return 0, fmt.Errorf("unable to get MemTotal in /proc/meminfo")
```
jodh-intel: Again, this message can reference the variable for `/proc/meminfo`.

qemu_test.go
```diff
@@ -395,10 +395,16 @@ func TestQemuSetMemoryResources(t *testing.T) {
 
 	q := &qemu{}
 
+	hostMem, err := getHostMemorySize()
```
jodh-intel: `s/hostMem/hostMemBytes/`

qemu_test.go
```diff
@@ -395,10 +395,16 @@ func TestQemuSetMemoryResources(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	memMax := fmt.Sprintf("%dM", int(float64(hostMem)/1024/1024)+1024)
```
jodh-intel: I'd add a paranoid check to ensure that `hostMemBytes > 0` before the division. It would also be clearer potentially to define variables for the division + addition with comments explaining why those values were chosen.

qemu.go
wcwxyz: emm... the code itself is clear enough. Since `memMax` is a string, I think I'm gonna change `hostMem` to `hostMemBytes` to be clearer.

qemu.go
wcwxyz: Will do.

qemu.go
wcwxyz: Will do.

qemu_test.go
wcwxyz: Sure. Thanks.

qemu_test.go
wcwxyz: yes, you're right. Will do that.

qemu_test.go
```diff
@@ -409,13 +416,76 @@ func TestQemuSetMemoryResources(t *testing.T) {
...
+	dir, err := ioutil.TempDir("", "")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(dir)
+
+	file := filepath.Join(dir, "meminfo")
+	// file doesn't exist
+	if _, err := getHostMemorySizeBytes(file); err == nil {
+		t.Fatal()
```
jodh-intel: Nit: You could remove the comment above and make this:

```
t.Fatalf("expected failure as file %q does not exist", file)
```

qemu.go
```diff
@@ -106,6 +108,11 @@ const (
 	maxDevIDSize = 31
 )
 
+const (
+	// NVDIMM device needs memory space (1024MB is suggested by @grahamwhaley)
```
jodh-intel: +1 for the comment but please can you add a link to an issue/PR rather than mentioning individuals.

wcwxyz:
@jodh-intel thanks for review. I'll update the code after @grahamwhaley's review.

qemu.go
wcwxyz: something like this? https://github.com/clearcontainers/runtime/issues/380#issuecomment-321216520

qemu_test.go
wcwxyz: yeah, sure.

qemu.go
wcwxyz: I mean the full url, github seems automatically format that.

grahamwhaley: looks correct to me :-) (yes, github nicely hides the full URL where it can :-) )

wcwxyz:
I take @grahamwhaley is ok with the rest of the patch. 
Re-pushed. Please take a look.

grahamwhaley:
thanks @wcwxyz that:
lgtm
@sboeuf can you review, and if happy, merge?

sboeuf:
I'll review tomorrow, please don't merge until I have reviewed this one.

grahamwhaley:
adding DNM label until @sboeuf gets a chance to review etc.

qemu.go
sboeuf: I agree with @jodh-intel, you should not mention @grahamwhaley directly into the source code. And please refer only the isssue, not the very specific comment.

qemu.go
```diff
@@ -106,6 +108,12 @@ const (
...
+	// See https://github.com/clearcontainers/runtime/issues/380#issuecomment-321216520
+	defaultNvdimmMemory = 1024
```
sboeuf: I would call this constant `maxMemoryOffset` because that's basically what it is, and we don't need to specify the reason in the variable name. As long as we have your comment, a generic name sounds better to me.

qemu.go
```diff
@@ -468,13 +476,20 @@ func (q *qemu) setCPUResources(podConfig PodConfig) ciaoQemu.SMP {
 	return smp
 }
 
-func (q *qemu) setMemoryResources(podConfig PodConfig) ciaoQemu.Memory {
+func (q *qemu) setMemoryResources(podConfig PodConfig) (ciaoQemu.Memory, error) {
+	hostMemBytes, err := getHostMemorySizeBytes("/proc/meminfo")
```
sboeuf: `/proc/meminfo` is a very well known path for Linux. Please define this as a const and don't provide it through this call, instead use it directly from the function `getHostMemorySizeBytes()`.

qemu.go
```diff
@@ -468,13 +476,20 @@ func (q *qemu) setCPUResources(podConfig PodConfig) ciaoQemu.SMP {
...
+	if err != nil {
+		return ciaoQemu.Memory{}, fmt.Errorf("Unable to read memory info: %s", err)
+	}
+	if hostMemBytes == 0 {
+		return ciaoQemu.Memory{}, fmt.Errorf("Error host memory size 0")
+	}
+
+	// add 1G memory space for nvdimm device (vm guest image)
+	memMax := fmt.Sprintf("%dM", int(float64(hostMemBytes)/1024/1024)+defaultNvdimmMemory)
```
sboeuf: This is wrong to do that, isn't it ? I mean you're clearly asking to run a VM with a memory max greater than what the system has to offer ? Or am I reading the code wrong ?
Instead you should check that the system can allow at least the expected memory to be set and this would be a nice error to throw. IMO, we don't want to continue if our host cannot provide the amount of memory requested, because the user could end up in not being able to run the workload he expects.

And I think that you should validate that the expected amount of memory + the offset can fit in the host memory. But please don't consider this case as throwing an error, just log a warning, because the offset might be too much in some cases that could end up functioning very well. Indeed, in case you have the NVDIMM using 200MiB and you ask 1.5GiB on a 2GiB host, you really need only 1.7 and this will work great. But if we check with our offset of 1GiB, we re gonna fail, that's why we don't want to generate an error in that case.

qemu.go
```diff
@@ -483,7 +498,42 @@ func (q *qemu) setMemoryResources(podConfig PodConfig) ciaoQemu.Memory {
...
+		if len(parts) < 3 || parts[0] != "MemTotal:" || parts[2] != "kB" {
+			continue
+		}
+
+		// Convert kB to bytes.
+		size, err := strconv.Atoi(parts[1])
+		if err != nil {
+			continue
+		}
+		bytes := uint64(size) * 1024
```
sboeuf: Why are we converting to bytes ? I mean we're basically multiplying by 1024 here to divide twice from the caller. I would not bother doing this and instead name the function `getHostMemorySizeKb`.

qemu.go
wcwxyz: No. maxmem is used to define upper bounds for memory hotplug. You request memory via `qemu -m MEMORY`.

Say run `qemu -m 2048M,slots=2,maxmem=30000M`, you got 2048M for on your VM and you can hot plug in 2 rom device (pc-dimm) into VM, the total memory can not exceed 30000M. In case your host memory is less than 30000M, hot plug will fail when your ask for memory when host doesn't have available.

qemu.go
wcwxyz: Will do.

qemu.go
wcwxyz: Will do.

qemu.go
wcwxyz: I'll change to KB.

qemu.go
sboeuf: That's exactly what you're saying, in case the host memory is less than 30G, the hotplug will fail. That's why I think we should try to set a maxmem higher than the total RAM available on the host.

wcwxyz: Aha, I understand what you're trying to say.

You're wrong about "offset". It is not consuming real host memory. Nvdimm needs "offset" to work on qemu. It takes memory space, not the real memory. (I'm not familiar with nvdimm device, but a quick test suggests that).

For maxmem, this is just the memory upper bounds for qemu. Setting maxmem to `host_total_memory+offset` is not asking qemu to run a VM with that much memory. Qemu will accept or reject to proceed any memory hotplug requests based on the maxmem. Whether hot plug works or not, depends kernel memory management.

grahamwhaley: For 'offset', allocating that extra space and the extra slot does not take any host memory at startup. Then later we plug the NVDIMM into that space. I think we are probably safe not calculating that into the 'total host memory' equation.

For 'mem_request > host_memory' - whether the host allows or denies a VM (CC-container) to be created if it requests more memory than the host has either available or system total I believe actually comes down to how the host kernel is configured for over-commit. If the system owner has knowingly configured the system to allow VM overcommit then we should probably not error. I think we could be more informative though with one or both of the following options maybe:

- warn if the requested memory is > the current system free memory (the default kernel non-overcommit failure case I believe)
- warn if the requested memory is > the system memory total (because you are likely going to run into performance issues if your container uses all that memory)

For more information on kernel overcommit settings etc. see:
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/tree/Documentation/sysctl/vm.txt
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/tree/Documentation/vm/overcommit-accounting

wcwxyz: Thanks @grahamwhaley for elaborating on memory overcommit.

I don't think we should do anything about user requested memory.  Runtime should not log error/warn solely based on overcommit or free memory. When QEMU fails to allocate memory, it errors out. There's enough information for user to see.

Kernel memory management is tricky. I really don't think we should make assumptions about memory is not enough for VM under any condition. Because we can only know memory actually runs out when there's allocation failure at the moment.

@grahamwhaley @sboeuf What do you think?

grahamwhaley: I'm happy with that theory - we cannot tell when a VM launch will fail or not, as we don't have the intimate knowledge of the system.
Later if we find that the qemu failure is non-obvious from the logs, then we can maybe work on improving how the error/logs are gathered and pushed up through the runtime.

wcwxyz: Thanks. I believe this part of code is ok, if @sboeuf doesn't have any objections.

sboeuf: Ok you convinced me :)

wcwxyz: Thanks for review @sboeuf @grahamwhaley. I'll post v2.

qemu.go
```diff
@@ -106,6 +108,14 @@ const (
 	maxDevIDSize = 31
 )
 
+const (
+	// NVDIMM device needs memory space 1024MB
+	// See https://github.com/clearcontainers/runtime/issues/380
+	maxMemoryOffset = 1024
+)
+
+const procMemInfo = "/proc/meminfo"
```
sboeuf: Please move this to `hypervisor.go` since it can be reused across different hypervisor implementations.

qemu.go
```diff
@@ -483,7 +500,40 @@ func (q *qemu) setMemoryResources(podConfig PodConfig) ciaoQemu.Memory {
 		MaxMem: memMax,
 	}
 
-	return memory
+	return memory, nil
+}
+
+func getHostMemorySizeKb(memInfoPath string) (uint64, error) {
```
sboeuf: Please move this to `hypervisor.go` since it can be reused across different hypervisor implementations.

wcwxyz:
Here comes v3. Thanks.

PR #338, opened 2017-08-10
hyperstart: move hyper and tty socket to per-pod runtime storage path

### LINKED ISSUE #270: Put the hyper ctl/io sockets in the per-pod directory
I see this going through the proxy (foo is my container ID):

```
RegisterVM(containerId=foo,ctlSerial=/tmp/hyper-pod-foo.sock,ioSerial=/tmp/tty-podfoo.sock,console=/run/virtcontainers/pods/foo/console.sock)
```

Two things:

- We're missing a '-' in the io file name
- I think those sockets would be better in the pod directory, next to console.sock

wcwxyz:
Currently we have hyper/tty socket in /tmp/{hyper,tty}-pod-id.sock. We should
move them to /run/virtcontainers/pods/id/{hyper,tty}.sock to be unified.

Fixes #270

### DIFF
--- a/hyperstart.go
+++ b/hyperstart.go
@@ -27,7 +27,7 @@ import (
 	"github.com/containers/virtcontainers/pkg/hyperstart"
 )
 
-var defaultSockPathTemplates = []string{"/tmp/hyper-pod-%s.sock", "/tmp/tty-pod%s.sock"}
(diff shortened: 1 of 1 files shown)


jodh-intel:
good catch!

lgtm

sboeuf:
@wcwxyz This is a simple patch but it really helps to improve and get virtcontainers more consistent. Thanks :)

LGTM

PR #340, opened 2017-08-11
pause: Rewrite pause binary in go.

### LINKED ISSUE #339: Rewrite pause binary in go.
Requiring the pause binary be written in C means we need to build it statically. However, this is problematic on RHEL 7, so let's rewrite it in go for now.

See https://github.com/clearcontainers/packaging/issues/38.

jodh-intel:
Requiring a statically-linked C program causes problems on some
platforms (such as RHEL).

Fixes #339.


pause/pause.go
```diff
@@ -14,22 +13,37 @@
...
+)
+
+func sigHandler(sig os.Signal) {
+	var ret int
 
-static void sigdown(int signo) {
-	psignal(signo, "shutting down, got signal");
-	exit(0);
+	switch sig {
+	case os.Interrupt:
```
sboeuf: You could maybe try to be consistent with the following statement by using `syscall.SIGINT`.

pause/pause.go
```diff
@@ -14,22 +13,37 @@
...
-	if (signal(SIGTERM, sigdown) == SIG_ERR)
-		return 2;
-	for (;;) pause();
-	fprintf(stderr, "error: infinite loop terminated\n");
-	return 42;
+func main() {
+	ch := make(chan os.Signal, 1)
+	signal.Notify(ch)
+
+	for {
```
sboeuf: I think you could avoid to buffer the channel `ch` and that way, you don't need the loop `for`. The call to switch in the function signal handler should block until it gets a signal.

pause/pause.go
jodh-intel: Yep - fixing...

pause/pause.go
sboeuf: @jodh-intel what do you think about this one ?

pause/pause.go
sboeuf: Thanks!

jodh-intel:
This code matches the implementation on https://github.com/clearcontainers/runtime/pull/434 (the runtime has its own copy of the pause binary sources).

PR #341, opened 2017-08-15
pkg: oci: support OCI spec memory limit

wcwxyz:
We should run VM(container) regarding OCI spec config.json memory limit.

-- 

This pull request resolves https://github.com/clearcontainers/runtime/issues/381
The original patch is at https://github.com/clearcontainers/runtime/pull/390

### DIFF
--- a/pkg/oci/utils.go
+++ b/pkg/oci/utils.go
@@ -286,6 +286,30 @@ func (spec *CompatOCISpec) PodID() (string, error) {
 	return "", fmt.Errorf("Could not find pod ID")
 }
 
+// TODO: calculate number of VCPUs from OCI spec
+func vmConfig(ocispec CompatOCISpec, config RuntimeConfig) (vc.Resources, error) {
+	if ocispec.Linux == nil ||
+		ocispec.Linux.Resources == nil ||
+		ocispec.Linux.Resources.Memory == nil ||
+		ocispec.Linux.Resources.Memory.Limit == nil {
+		return config.VMConfig, nil
+	}
+
+	memBytes := *ocispec.Linux.Resources.Memory.Limit
+
+	if memBytes <= 0 {
+		return vc.Resources{}, fmt.Errorf("Invalid OCI memory limit %d", memBytes)
+	}
+
+	// round up memory to 1MB
+	mem := uint((memBytes + (1024*1024 - 1)) / (1024 * 1024))
+
+	return vc.Resources{
+		VCPUs:  config.VMConfig.VCPUs,
+		Memory: mem,
+	}, nil
+}
+
 // PodConfig converts an OCI compatible runtime configuration file
 // to a virtcontainers pod configuration structure.
 func PodConfig(ocispec CompatOCISpec, runtime RuntimeConfig, bundlePath, cid, console string, detach bool) (vc.PodConfig, error) {
@@ -301,12 +325,17 @@ func PodConfig(ocispec CompatOCISpec, runtime RuntimeConfig, bundlePath, cid, co
 		return vc.PodConfig{}, err
 	}
 
+	resources, err := vmConfig(ocispec, runtime)
+	if err != nil {
+		return vc.PodConfig{}, err
+	}
+
 	podConfig := vc.PodConfig{
 		ID: cid,
 
 		Hooks: containerHooks(ocispec),
 
-		VMConfig: runtime.VMConfig,
+		VMConfig: resources,
 
 		HypervisorType:   runtime.HypervisorType,
 		HypervisorConfig: runtime.HypervisorConfig,
--- a/pkg/oci/utils_test.go
+++ b/pkg/oci/utils_test.go
@@ -161,6 +161,49 @@ func TestMinimalPodConfig(t *testing.T) {
 	}
 }
 
+func TestVmConfig(t *testing.T) {
+	var limitBytes int64 = 128 * 1024 * 1024
+
+	config := RuntimeConfig{
+		VMConfig: vc.Resources{
+			Memory: 2048,
+		},
+	}
+
+	expectedResources := vc.Resources{
+		Memory: 128,
+	}
+
+	ocispec := CompatOCISpec{
+		Spec: specs.Spec{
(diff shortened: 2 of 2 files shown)


pkg/oci/utils.go
```diff
@@ -286,6 +287,35 @@ func (spec *CompatOCISpec) PodID() (string, error) {
 	return "", fmt.Errorf("Could not find pod ID")
 }
 
+func ociResourceUpdateRuntimeConfig(ocispec CompatOCISpec, config *RuntimeConfig) error {
```
sboeuf: Please create instead a function:
```go
func vmConfig(ocispec CompatOCISpec, runtime RuntimeConfig) (Resources, error) {
    ...
}
```

In case there is no memory limit defined by the config.json file, we fallback to what is provided by `RuntimeConfig`. Otherwise, we take the value provided by the config.json.

pkg/oci/utils.go
```diff
@@ -286,6 +287,35 @@ func (spec *CompatOCISpec) PodID() (string, error) {
...
+	if ocispec.Linux == nil {
```
sboeuf: Make sure you gather all the statements together as they end up with the same result:
```go
if ocispec.Linux == nil ||
   ocispec.Linux.Resources == nil ||
   ocispec.Linux.Resources.Memory == nil ||
   ocispec.Linux.Resources.Memory.Limit == nil {
	return nil
}
```

pkg/oci/utils.go
```diff
@@ -286,6 +287,35 @@ func (spec *CompatOCISpec) PodID() (string, error) {
...
+	}
+
+	if ocispec.Linux.Resources.Memory.Limit == nil {
+		return nil
+	}
+
+	memBytes := *ocispec.Linux.Resources.Memory.Limit
+
+	// round up memory bytes to MB
+	mem := int(math.Ceil(float64(memBytes) / 1024 / 1024))
```
sboeuf: We don't need that much precision, I think this should be enough
```go
mem := uint(memBytes / (1024 * 1024))
```

wcwxyz: I'm ok with less precision, because this is trivial.

But @grahamwhaley suggested that we should round up at https://github.com/clearcontainers/runtime/pull/390#discussion_r132192014. I just quoted here:

> maybe we should round this up. Docker allows fine granularity (down to bytes). I think it is probably better we allocate more than requested rather than too little

@grahamwhaley what do you think now?

pkg/oci/utils.go
wcwxyz: Will do.

pkg/oci/utils.go
wcwxyz: Will do.

sboeuf:
@wcwxyz Global question here, why do you use [RFC] prefix ? What does that mean ?

wcwxyz:
@sboeuf It's Request For Comment, meaning the idea or design needs to be discussed or commented.

sboeuf:
@wcwxyz oh okay, I didn't know. I think you can remove it since we consider every PR as something we want to review thoroughly :)

wcwxyz:
@sboeuf Sure. Thanks for heads-up.

grahamwhaley:
:-) RFC sort of comes from here: https://www.ietf.org/rfc.html, but funnily enough most of those RFCs pretty much become 'standards', but are still called RFCs :-)
Adding RFC here I think is fine, it is a common idiom - but, you should probably also add our 'do-not-merge' github label at the same time, to indicate this is something to discuss and work on and not to immediately merge, until agreement has been found.

pkg/oci/utils.go
grahamwhaley: I still think rounding up is the technically right thing to do. I'd rather hand a container slightly more than requested, rather than artificially starve it. tbh though, the amount we are likely to drop as a % from not rounding up is tiny - given the smallest memory we can probably support might be 128m or so.
If we don't like the float math, we could do something integer like this to round up:
```
mem = (memBytes + ((1024*1024)-1))) / (1024*1024)
```

sboeuf: @grahamwhaley yes I like this one, you add 1MiB to make sure we round this up. My main concern was to avoid playing with `float64` numbers whilst we don't __really__ need that much precision.

wcwxyz: Thanks @grahamwhaley @sboeuf . I'll update the patch.

pkg/oci/utils.go
```diff
@@ -286,6 +286,31 @@ func (spec *CompatOCISpec) PodID() (string, error) {
...
+		ocispec.Linux.Resources.Memory == nil ||
+		ocispec.Linux.Resources.Memory.Limit == nil {
+		return config.VMConfig, nil
+	}
+
+	memBytes := *ocispec.Linux.Resources.Memory.Limit
+
+	// round up memory to 1MB
+	mem := uint((memBytes + (1024*1024 - 1)) / (1024 * 1024))
+	if mem <= 0 {
```
sboeuf: We should check `memBytes` instead of `mem`, that's the one which could be negative. You should make this check right after you retrieved `memBytes` value.

pkg/oci/utils.go
```diff
@@ -286,6 +286,31 @@ func (spec *CompatOCISpec) PodID() (string, error) {
...
+		return vc.Resources{}, fmt.Errorf("Invalid OCI memory limit %d", memBytes)
+	}
+
+	cpu := config.VMConfig.VCPUs
+
+	return vc.Resources{
+		VCPUs:  cpu,
```
sboeuf: Assign directly `VCPUs` to `config.VMConfig.VCPUs` here. If we need to use a temporary variable for later, let's do that in a follow up patch.

pkg/oci/utils.go
wcwxyz: You're right. I missed this one while updating my code last time.

pkg/oci/utils.go
wcwxyz: Sure. Will update.

PR #343, opened 2017-08-16
qemu: Detect VMM environment to apply specific flags to our VM

### LINKED ISSUE #342: qemu: Detect VMM environment to apply specific flags to our VM
When Clear Containers runs on Azure, it actually runs inside an existing VM (nested VM) managed by Hyper-V hypervisor. And because KVM has some issues when running on a VMM, the only way to get Clear Containers working is to fallback from virtio 1.0 to a previous version, and to disable pmu.
We need to implement a way to detect VMM environment and enable/disable specific flags accordingly.

sboeuf:
When Clear Containers runs on Azure, it actually runs inside an existing VM (nested VM) managed by Hyper-V hypervisor. And because KVM has some issues when running on a VMM, the only way to get Clear Containers working is to fallback from virtio 1.0 to a previous version, and to disable pmu.
We need to implement a way to detect VMM environment and enable/disable specific flags accordingly.

Fixes #342

### DIFF
--- a/hypervisor.go
+++ b/hypervisor.go
@@ -35,7 +35,10 @@ const (
 	MockHypervisor HypervisorType = "mock"
 )
 
-const procMemInfo = "/proc/meminfo"
+const (
+	procMemInfo = "/proc/meminfo"
+	procCPUInfo = "/proc/cpuinfo"
+)
 
 const (
 	defaultVCPUs = 1
@@ -227,19 +230,6 @@ func DeserializeParams(parameters []string) []Param {
 	return params
 }
 
-// hypervisor is the virtcontainers hypervisor interface.
-// The default hypervisor implementation is Qemu.
-type hypervisor interface {
-	init(config HypervisorConfig) error
-	createPod(podConfig PodConfig) error
-	startPod(startCh, stopCh chan struct{}) error
-	stopPod() error
-	pausePod() error
-	resumePod() error
-	addDevice(devInfo interface{}, devType deviceType) error
-	getPodConsole(podID string) string
-}
-
 func getHostMemorySizeKb(memInfoPath string) (uint64, error) {
 	f, err := os.Open(memInfoPath)
 	if err != nil {
@@ -271,4 +261,59 @@ func getHostMemorySizeKb(memInfoPath string) (uint64, error) {
 	}
 
 	return 0, fmt.Errorf("unable get MemTotal from %s", memInfoPath)
+}
+
+// RunningOnVMM checks if the system is running inside a VM.
+func RunningOnVMM(cpuInfoPath string) (bool, error) {
+	flagsField := "flags"
+
+	f, err := os.Open(cpuInfoPath)
+	if err != nil {
+		return false, err
+	}
+	defer f.Close()
+
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		// Expected format: ["flags", ":", ...] or ["flags:", ...]
+		fields := strings.Fields(scanner.Text())
+		if len(fields) < 2 {
+			continue
+		}
+
+		if !strings.HasPrefix(fields[0], flagsField) {
+			continue
+		}
+
+		for _, field := range fields[1:] {
+			if field == "hypervisor" {
+				return true, nil
+			}
+		}
+
+		// As long as we have been able to analyze the fields from
+		// "flags", there is no reason to check what comes next from
+		// /proc/cpuinfo, because we already know we are not running
+		// on a VMM.
+		return false, nil
+	}
+
+	if err := scanner.Err(); err != nil {
+		return false, err
+	}
+
+	return false, fmt.Errorf("Couldn't find %q from %q output", flagsField, cpuInfoPath)
+}
+
+// hypervisor is the virtcontainers hypervisor interface.
+// The default hypervisor implementation is Qemu.
+type hypervisor interface {
+	init(config HypervisorConfig) error
+	createPod(podConfig PodConfig) error
+	startPod(startCh, stopCh chan struct{}) error
+	stopPod() error
+	pausePod() error
+	resumePod() error
+	addDevice(devInfo interface{}, devType deviceType) error
+	getPodConsole(podID string) string
 }
--- a/qemu.go
+++ b/qemu.go
@@ -50,6 +50,8 @@ type qemu struct {
 	qmpControlCh qmpChannel
 
 	qemuConfig ciaoQemu.Config
+
+	nestedRun bool
 }
 
 const defaultQemuPath = "/usr/bin/qemu-system-x86_64"
@@ -206,6 +208,7 @@ func (q *qemu) appendVolume(devices []ciaoQemu.Device, volume Volume) []ciaoQemu
 			Path:          volume.HostPath,
 			MountTag:      volume.MountTag,
 			SecurityModel: ciaoQemu.None,
+			DisableModern: q.nestedRun,
 		},
 	)
 
@@ -223,12 +226,13 @@ func (q *qemu) appendBlockDevice(devices []ciaoQemu.Device, drive Drive) []ciaoQ
 
 	devices = append(devices,
 		ciaoQemu.BlockDevice{
-			Driver:    ciaoQemu.VirtioBlock,
-			ID:        drive.ID,
-			File:      drive.File,
-			AIO:       ciaoQemu.Threads,
-			Format:    ciaoQemu.BlockDeviceFormat(drive.Format),
-			Interface: "none",
+			Driver:        ciaoQemu.VirtioBlock,
+			ID:            drive.ID,
+			File:          drive.File,
+			AIO:           ciaoQemu.Threads,
+			Format:        ciaoQemu.BlockDeviceFormat(drive.Format),
+			Interface:     "none",
+			DisableModern: q.nestedRun,
 		},
 	)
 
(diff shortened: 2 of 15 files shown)


sboeuf:
I have assigned `do-not-merge` label to this PR since we need changes from ciao/qemu https://github.com/01org/ciao/pull/1396. We need to add the new ciao/qemu vendoring to this PR.

mcastelino:
@sboeuf we can detect if we are running on a VMM via CPUID.

From Hyper-V documentation
```
Detecting the Presence of a Hypervisor
Software determines the presence of a hypervisor through the CPUID instruction. Processors conforming to the Intel® 64 architecture have reserved a feature flag in CPUID Function 0x01 - Feature Information for this purpose.  Bit 31 returned in ECX is defined as Not Used, and will always return 0 from the physical CPU.  A hypervisor conformant with the Microsoft hypervisor interface will set CPUID.1:ECX [bit 31] = 1 to indicate its presence to software.
Determining Hypervisor Capabilities
When the hypervisor present bit is set, additional CPUID leaves are provided by the hypervisor which will return more information about the hypervisor and its capabilities.
The Intel® 64 architecture reserves CPUID leaves 0x40000000-0x400000FF for use by system software. A Microsoft compliant hypervisor guarantees leaves 0x40000000 and 0x40000001 are always available
Microsoft Hypervisor Interface Identification
The standard hypervisor CPUID leaf is provided at 0x40000000.  When queried, this leaf will return the maximum hypervisor CPUID leaf number, and the vendor ID signature.
Hypervisor Vendor Neutral Interface Identification
The hypervisor interface identification is provided at CPUID leaf 0x40000001.  Hypervisors conforming to the Microsoft hypervisor interface will return the hypervisor interface identification signature ‘Hv#1’ (0x31237648) in CPUID.40000001:EAX.
```

jodh-intel:
Yes, if we can avoid calling an external binary, that's got to be a good thing. Also, I have seen systems hang/crash running `dmidecode` before (although admittedly, it was a loooong time ago now ;)

sboeuf:
@mcastelino I have updated this PR based on what we have discussed (rely on `/proc/cpuinfo` to detect VMM, and added `pmu=off` to the `-cpu host` option).
Could you please merge ciao's PRs https://github.com/01org/ciao/pull/1396 and https://github.com/01org/ciao/pull/1397 so that I can rebase/revendor this PR on top ?

sboeuf:
@jodh-intel yes I agree I prefer the solution reading `/proc/cpuinfo` rather than invoking an external binary !

hypervisor.go
```diff
@@ -272,3 +262,57 @@ func getHostMemorySizeKb(memInfoPath string) (uint64, error) {
...
+
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		// Expected format: ["flags", ":", ...] or ["flags:", ...]
+		fields := strings.Fields(scanner.Text())
+		if len(fields) < 2 {
+			continue
+		}
+
+		if !strings.Contains(fields[0], flagsField) {
```
jodh-intel: `strings.HasPrefix()` would be safer.

qemu.go
```diff
@@ -425,6 +433,13 @@ func (q *qemu) init(config HypervisorConfig) error {
 		return err
 	}
 
+	nested, err := runningOnVMM(procCPUInfo)
+	if err != nil {
+		return err
+	}
+
+	q.nestedRun = nested
```
jodh-intel: Might be a good idea to log whether we think we're running nested here?

hypervisor.go
```diff
@@ -272,3 +262,57 @@ func getHostMemorySizeKb(memInfoPath string) (uint64, error) {
 
 	return 0, fmt.Errorf("unable get MemTotal from %s", memInfoPath)
 }
+
+func runningOnVMM(cpuInfoPath string) (bool, error) {
```
jodh-intel: This is a useful function. How about making it public to allow consumers of virtcontainers to also make use of it?

hypervisor.go
sboeuf: Will fix.

qemu.go
sboeuf: Will do.

hypervisor.go
sboeuf: Yes I agree, I'll do that.

sboeuf:
@mcastelino @jodh-intel I have revendored ciao and rebased this PR + applied the changes you asked for @jodh-intel. Please let me know if this sounds good now.

jodh-intel:
Thanks!

lgtm

sboeuf:
@amshinde please take a look, I need your approval for this one.

amshinde:
lgtm

HowardHao1:
I followed the steps at https://github.com/clearcontainers/runtime/wiki/Installing-Clear-Containers-3.0-on-Ubuntu to install Clear Container on a HYPER-V VM running Ubuntu. I received the following error: kvm[4459]:vcpu0 unhandled rdmsr: 0x34.

Any updated instructions to avoid this issue?

grahamwhaley:
Hi @HowardHao1 
The `kvm[4459]:vcpu0 unhandled rdmsr: 0x34` I believe is actually benign (I also see those when running on a bare-metal system for instance), so I suspect there is something else we need to diagnose.
Can you supply us the output of `cc-collect-data.sh` please so we can check if anything leaps out there?
Also, when you try to run the container, what happens - do you end up just back at the host prompt or does the container run or??
Thanks (and thanks for reporting!). Depending on how this goes maybe we should take this to an issue over on https://github.com/clearcontainers/runtime
/cc @jodh-intel @mcastelino

HowardHao1:
Hi @grahamwhaley,  I didn't realize that clear container uses a special version Docker daemon. After updated the Docker daemon with the right one, the issues went away.
I measured the Clear Container Ubuntu start up time. I am really surprised that the Clear Container start up time was really close to regular Linux Ubuntu container startup time.  I wonder how I can  tell if I am really running Clear Containers.  For each Clear Container that I created, is there a separated process to back the Clear Container?
thanks,
Howard Hao.

sboeuf:
Hi @HowardHao1, I am glad you made it work :)
For each Clear Containers container that you start, you should be able to see a `cc-shim` process running. And every time you add more processes to this container, you will see new `cc-shim` processes created.
Also, you will see one qemu process per pod (k8s case)

grahamwhaley:
Hi @HowardHao1 - great. Yeah, the docker interfaces change a lot and do not always seem to be backwards compatible :-(, so sometimes we have to stick to certain versions...
You can also tell if you are in a Clear Container by checking which kernel you are running in the container. A `uname -a` inside a Clear Container will show a different kernel version that `uname -a` on the host - whereas inside a normal docker container they would be the same (as the normal docker container uses the host kernel directly).

HowardHao1:
Since cc is KVM backed, can I run containers with windows images? Is there a special command to bring down the windows images?  I tried the following command and received a platform error:
root@xenial:/# docker run -it microsoft/windowsservercore cmd.exe
Unable to find image 'microsoft/windowsservercore:latest' locally
latest: Pulling from microsoft/windowsservercore
3889bb8d808b: Pulling fs layer
3725c17d990a: Downloading
docker: image operating system "windows" cannot be used on this platform.

sboeuf:
@HowardHao1 well first the error you're getting here is a Docker error indicating that you cannot do that. This means the command is not even going through Clear Containers runtime here, but Docker directly identifies this cannot work. Which is valid in case of baremetal containers since you are on a Linux kernel and you expect to run Windows kernel.

Now, in case of Clear Containers, if we can bypass the Docker limitation, we could in theory support windows containers, but we are missing a very important part here, that is we don't have a windows image and a windows agent. We would need to implement what we are doing with our agent inside the Linux VM, and duplicate that into a Windows agent.

In short, this could be done (with a certain amount of work), but there is no support at the moment.

HowardHao1:
I have modified clear-containers.conf as following to enable docker remote:
[Service]
ExecStart=
ExecStart=/usr/bin/dockerd -H tcp://0.0.0.0:2375 -D --add-runtime cc-runtime=/usr/bin/cc-runtime --default-runtime=cc-runtime

I ran into two issues:

1. Running docker command locally hung. If I login to the Linux container host VM with putty as root, all the docker commands hung, but on a remote machine, command "docker -H hostVMIP" worked fine.

2. The issue that I mentioned before still showed: vcpu0 unhandled rdmsr: 0x34. It showed up in the TTY console as the following:

image

I have dumped the process tree to make sure that clear containers were working as expected:
        ├─dockerd─┬─docker-containe─┬─4*[docker-containe─┬─cc-proxy───5*[{cc-proxy}]]
        │         │                 │                    ├─cc-shim───cc-shim]
        │         │                 │                    ├─qemu-lite-syste───9*[{qemu-lite-syste}]]
        │         │                 │                    └─9*[{docker-containe}]]
        │         │                 ├─2*[docker-containe─┬─cc-shim───cc-shim]
        │         │                 │                    └─10*[{docker-containe}]]
        │         │                 ├─2*[docker-containe─┬─cc-proxy───5*[{cc-proxy}]]
        │         │                 │                    ├─cc-shim───cc-shim]
        │         │                 │                    ├─qemu-lite-syste───9*[{qemu-lite-syste}]]
        │         │                 │                    └─10*[{docker-containe}]]
        │         │                 ├─3*[docker-containe─┬─cc-shim───cc-shim]
        │         │                 │                    └─9*[{docker-containe}]]
        │         │                 ├─docker-containe─┬─cc-shim───cc-shim
        │         │                 │                 └─8*[{docker-containe}]
        │         │                 └─33*[{docker-containe}]
        │         ├─4*[docker-proxy───5*[{docker-proxy}]]
        │         ├─docker-proxy───6*[{docker-proxy}]
        │         ├─docker-proxy───4*[{docker-proxy}]
        │         └─45*[{dockerd}]

Please let me know if I missed anything here.
Regards,
Howard Hao

jodh-intel:
Hi @HowardHao1 - please can you raise new issues for the various questions you have - commenting on an already-closed PR is very confusing.

Can you also ensure that you include the output of `cc-collect-data.sh` as previously requested (https://github.com/containers/virtcontainers/pull/343#issuecomment-352382570). When you open a new issue (https://github.com/clearcontainers/runtime/issues/new), the template will prompt you for this too.

grahamwhaley:
Hi @HowardHao1 - as @jodh-intel says, for issue (1) (remote docker), can you open a new Issue please?
For issue (2) - as noted before, the `unhandled rdmsr: 0x34` is a benign warning from the host side KVM. It is not something specific to Clear Containers I'm afraid, this is a generic thing around qemu/KVM. If you search you will find many instances of the same question on the internet.
 I'd love for us to 'get rid' of that, but there is nothing (nothing that is not very invasive or non-trivial anyhow) we can do. Thus, that is not your problem here - pls ignore it :-)
(I will see if I can find a place for us to document that though on this repo, to make it easier for folks to search for and find the answer easier in the future).

grahamwhaley:
And a final(?) update - it appears there are patches in flight for the kernel to both allow suppression of the warnings, and to implement some of the features to stop the warnings in the first place:
https://patchwork.kernel.org/patch/10048427/

Of course, they will take time to land and trickle through to the main distros kernels - but, at least there is a progress here :-)

PR #344, opened 2017-08-17
pkg: oci: support OCI spec cpu quota/period

wcwxyz:
We should run VM(container) regarding OCI spec config.json cpu quota and
period.

The number of VM VCPUs is defined by quota/period (round up to 1).

-- 
Original issues is https://github.com/clearcontainers/runtime/issues/393

### DIFF
--- a/pkg/oci/utils.go
+++ b/pkg/oci/utils.go
@@ -286,13 +286,14 @@ func (spec *CompatOCISpec) PodID() (string, error) {
 	return "", fmt.Errorf("Could not find pod ID")
 }
 
-// TODO: calculate number of VCPUs from OCI spec
 func vmConfig(ocispec CompatOCISpec, config RuntimeConfig) (vc.Resources, error) {
+	resources := config.VMConfig
+
 	if ocispec.Linux == nil ||
 		ocispec.Linux.Resources == nil ||
 		ocispec.Linux.Resources.Memory == nil ||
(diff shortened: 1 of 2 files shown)


pkg/oci/utils.go
```diff
@@ -302,12 +303,25 @@ func vmConfig(ocispec CompatOCISpec, config RuntimeConfig) (vc.Resources, error)
...
+	if ocispec.Linux.Resources.CPU == nil ||
+		ocispec.Linux.Resources.CPU.Quota == nil ||
+		ocispec.Linux.Resources.CPU.Period == nil {
+		return resources, nil
+	}
+
+	quota := *ocispec.Linux.Resources.CPU.Quota
+	period := *ocispec.Linux.Resources.CPU.Period
+
+	if quota < 0 {
```
sboeuf: We want to test:
```go
if quota <= 0 {
```
as we did for the memory. Otherwise, in case `quota == 0`, we are gonna end up with `resources.VCPUs = 0` because `(period -1) / period` will return 0.

wcwxyz: Right. I also will check period is not zero.

sboeuf: Oh yes good point !

sboeuf:
LGTM

sboeuf:
@jodh-intel @grahamwhaley please give us your opinion before I merge this.

jodh-intel:
lgtm

PR #348, opened 2017-08-18
api/tests: Introduce a test mock

### LINKED ISSUE #346: Add mock virtcontainers implementation for testing
`cc-runtime`, a consumer of virtcontainers, needs to have a high level of unit-test coverage. However, the level of coverage is limited by the runtimes ability to test virtcontainers error paths.

Since it is undesirable to duplicate test code in the runtime that already exists in virtcontainers it makes sense for the runtime to use a mock for the virtcontainers package.

The complication is that certain virtcontainers objects (namely `Pod` and `Container`) contain only private fields. Since the virtcontainers API includes calls that return such objects, it is not possible for the runtime to mock virtcontainers in isolation: it needs support from the virtcontainers package.

Introduce a mock framework to virtcontainers to allow consumers to force arbitrary virtcontainers behaviour.

jodh-intel:
Add a mock implementation of virtcontainers via a new vcMock package
that defines a VCMock type, along with an interface that all
virtcontainers implementations are required to implement (VC).

Since the Pod and Container fields are entirely private, it is also
necessary to introduce interfaces for both those types (VCPod and
VCContainer respectively) to allow the mock implenentation to be fully
functional.

As some of the virtcontainers API calls return Pod and Container types,
the API has also had to be modified slightly to return the interface
type (required by the go language). This change then further required
that any functions that need to manipulate the real types (Pod and
Container for example) have to cast the interface types back to the
underlying types using type assertions.

Note that this is potentially an API-breaking change: if clients have
specified an explicit type, the code will need to be updated to change
the explicit types (Pod, Container) to the interface types (VCPod,
VCContainer).

For example, the following code:

    var p vc.Pod
    p = CreatePod(config)

... will need to be changed to either:

    var p vc.VCPod
    p = CreatePod(config)

... or more simply:

    p := CreatePod(config)

Fixes #346.


jodh-intel:
Tested in the runtime using:

- https://github.com/jodh-intel/runtime/tree/test-use-vcmock

jodh-intel:
Aside from the lint fluff, this PR needs 2 things:

- A concrete implementation of `VC` for virtcontainers itself.
- A test to ensure that both the virtcontainers and `vcMock` implement `VC`.

jodh-intel:
Hi @sboeuf - this is now ready for review.

pkg/vcMock/mock_test.go
```diff
@@ -0,0 +1,99 @@
...
+	// interface
+	mainImplType := reflect.TypeOf(mainImpl)
+	mainImplementsIF := mainImplType.Implements(reflect.TypeOf(&interfaceType).Elem())
+	assert.True(t, mainImplementsIF)
+
+	// check that the test implementation implements the
+	// interface
+	testImplType := reflect.TypeOf(testImpl)
+	testImplementsIF := testImplType.Implements(reflect.TypeOf(&interfaceType).Elem())
+	assert.True(t, testImplementsIF)
```
sboeuf: Those tests are nice :+1:

container.go
```diff
@@ -222,7 +222,7 @@ func (c *Container) fetchMounts() ([]Mount, error) {
 }
 
 // fetchContainer fetches a container config from a pod ID and returns a Container.
-func fetchContainer(pod *Pod, containerID string) (*Container, error) {
+func fetchContainer(pod *Pod, containerID string) (VCContainer, error) {
```
sboeuf: I don't think this function needs to return `VCContainer` type since it is only used from the virtcontainers implementation. Neither the mock implementation, nor any external consumer of the lib can access this function as it is private. Leaving this function unchanged simplifies the changes above (in `api.go`).

container.go
```diff
@@ -346,7 +346,7 @@ func newContainers(pod *Pod, contConfigs []ContainerConfig) ([]*Container, error
 }
 
 // createContainer creates and start a container inside a Pod.
-func createContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
+func createContainer(pod *Pod, contConfig ContainerConfig) (VCContainer, error) {
```
sboeuf: I don't think this function needs to return `VCContainer` type since it is only used from the virtcontainers implementation. Neither the mock implementation, nor any external consumer of the lib can access this function as it is private. Leaving this function unchanged simplifies the changes above (in `api.go`).

api.go
```diff
@@ -378,7 +378,8 @@ func CreateContainer(podID string, containerConfig ContainerConfig) (*Pod, *Cont
 	}
 
 	// Store it.
-	err = c.storeContainer()
+	cImpl := c.(*Container)
```
sboeuf: Not needed if we agree that `createContainer()` should not return `VCContainer` type.

api.go
```diff
@@ -423,7 +424,8 @@ func DeleteContainer(podID, containerID string) (*Container, error) {
 	}
 
 	// Delete it.
-	err = c.delete()
+	cImpl := c.(*Container)
```
sboeuf: Not needed if we agree that `fetchContainer()` should not return `VCContainer` type.

api.go
```diff
@@ -471,19 +473,24 @@ func StartContainer(podID, containerID string) (*Container, error) {
 		return nil, err
 	}
 
+	cImpl, ok := c.(*Container)
```
sboeuf: Not needed if we agree that `fetchContainer()` should not return `VCContainer` type.

api.go
```diff
@@ -509,19 +516,24 @@ func StopContainer(podID, containerID string) (*Container, error) {
...
```
sboeuf: Not needed if we agree that `fetchContainer()` should not return `VCContainer` type.

api.go
```diff
@@ -548,12 +560,13 @@ func EnterContainer(podID, containerID string, cmd Cmd) (*Pod, *Container, *Proc
 	}
 
 	// Enter it.
-	process, err := c.enter(cmd)
+	cImpl := c.(*Container)
```
sboeuf: Not needed if we agree that `fetchContainer()` should not return `VCContainer` type.

api.go
```diff
@@ -641,7 +654,8 @@ func KillContainer(podID, containerID string, signal syscall.Signal, all bool) e
 	}
 
 	// Send a signal to the process.
-	err = c.kill(signal, all)
+	cImpl := c.(*Container)
```
sboeuf: Not needed if we agree that `fetchContainer()` should not return `VCContainer` type.

jodh-intel:
@mcastelino - could you review please?

pkg/vcMock/types.go
```diff
@@ -0,0 +1,72 @@
...
+	MockID          string
+	MockURL         string
+	MockToken       string
+	MockProcess     vc.Process
+	MockPid         int
+	MockPod         *Pod
+	MockAnnotations map[string]string
+}
+
+// VCMock is an implementation of the VC interface used for testing
```
mcastelino: This is also an interface right? Not an implementation?

pkg/vcMock/mock.go
```diff
@@ -0,0 +1,189 @@
...
+func (m *VCMock) SetLogger(logger logrus.FieldLogger) {
+	if m.SetLoggerFunc != nil {
+		m.SetLoggerFunc(logger)
+		return
+	}
+}
+
+// CreatePod implements the VC function of the same name.
+func (m *VCMock) CreatePod(podConfig vc.PodConfig) (vc.VCPod, error) {
+	if m.CreatePodFunc != nil {
```
mcastelino: Maybe I am missing something. Where are these functions (*Func) getting assigned? Or are they supposed to always fail unless they are explicitly assigned in the test?

pkg/vcMock/pod.go
```diff
@@ -0,0 +1,75 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Description: A mock implementation of virtcontainers that can be used
+// for testing.
+//
+// This implementation calls the function set in the object that
```
mcastelino: Is this a copy paste carry over?

pkg/vcMock/types.go
```diff
@@ -0,0 +1,72 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Description: A mock implementation of virtcontainers that can be used
+// for testing.
+//
+// This implementation calls the function set in the object that
```
mcastelino: Again. May be a copy paste carry over?

pkg/vcMock/mock.go
jodh-intel: They aren't assigned here and yes, they are designed to fail in a consistent manner unless you provide an explicit function to modify that behaviour.

jodh-intel:
Hi @mcastelino - thanks for reviewing. I'll update the branch and add some further tests as you suggest...

pkg/vcMock/pod.go
jodh-intel: yep ;)

pkg/vcMock/types.go
jodh-intel: Fixing...

pkg/vcMock/types.go
jodh-intel: Yep.

jodh-intel:
Hi @sboeuf, @mcastelino - branch updated. I've fixed the comment issues @mcastelino raised plus added a test for all the calls to `pkg/vcMock/mock_test.go` asserting the expected behaviour (with and without the `VCMock.*Func` handlers set).

sboeuf:
Thanks for the update @jodh-intel 
LGTM

sboeuf:
Waiting for @mcastelino feedback before merging this !

PR #351, opened 2017-08-23
agent: Make the error message from agent generic.

### LINKED ISSUE #350: Make error message from agent generic
Specifically,
https://github.com/containers/virtcontainers/blob/913f9e723e3a4e1226534df650fe45e6285f074c/pkg/hyperstart/hyperstart.go#L442

As we have a new agent implementing hyperstart protocol for communication with proxy, this error message needs to be generic.

amshinde:
Since we have two different agents now implementing
hyperstart protocol, make the error message in case of
the agent returning an error, a bit more generic to
avoid confusion.

Fixes #350


sboeuf:
LGTM

PR #355, opened 2017-08-25
oci: Remove error return from StatusToOCIState()

### LINKED ISSUE #354: Remove error return from StatusToOCIState()
`StatusToOCIState()` currently returns an `error`, but there are no error scenarios to handle.

jodh-intel:
StatusToOCIState() does not need to return an error as there are no
error scenarios to deal with.

Fixes #354.

### DIFF
--- a/pkg/oci/utils.go
+++ b/pkg/oci/utils.go
@@ -431,8 +431,8 @@ func ContainerConfig(ocispec CompatOCISpec, bundlePath, cid, console string, det
 }
 
 // StatusToOCIState translates a virtcontainers container status into an OCI state.
-func StatusToOCIState(status vc.ContainerStatus) (spec.State, error) {
-	state := spec.State{
+func StatusToOCIState(status vc.ContainerStatus) spec.State {
+	return spec.State{
 		Version:     spec.Version,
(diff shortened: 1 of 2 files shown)


pkg/oci/utils.go
```diff
@@ -431,17 +431,15 @@ func ContainerConfig(ocispec CompatOCISpec, bundlePath, cid, console string, det
 }
 
 // StatusToOCIState translates a virtcontainers container status into an OCI state.
-func StatusToOCIState(status vc.ContainerStatus) (spec.State, error) {
-	state := spec.State{
+func StatusToOCIState(status vc.ContainerStatus) spec.State {
+	return spec.State{
```
grahamwhaley: I'm not even sure you need the return here do you? If you are filling out a return variable, doesn't it just 'fall out of the bottom' eu-natural? Having said that, leaving the return is fine - it makes it very clear what is going on :-)

jodh-intel: You can leave out the name of the variable on `return` if you've specified one in the signature but this function hasn't actually specified a name, just a type.

sboeuf:
Sure fine with me.
LGTM

PR #362, opened 2017-09-08
oci: Set hostname to the one passed in spec file

### LINKED ISSUE #361: Command prompt needs to be  fixed
With Clear containers, prompt shown is :
root@747d4f74b761b624543a4ed6825d1917aeef6dc8cbb64505cd33fab7ec714720:/#

With runc or CC2.1 we get a prompt as :
root@adc70fcce219

This is because the hostname is not correctly set to the shortened version.

amshinde:
Hostname was incorrectly being set to the pod-id.
It should instead be set to the hostname passed in the spec
file. The hostname passed is usually the shortened pod-id.

This should fix the command prompt seen when exec'ed into
a container to be the same as runc.

Fixes #361.

### DIFF
--- a/hyperstart.go
+++ b/hyperstart.go
@@ -428,7 +428,7 @@ func (h *hyper) startPod(pod Pod) error {
 		return err
 	}
 
-	hostname := pod.id
+	hostname := pod.config.Hostname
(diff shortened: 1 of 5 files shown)


sameo:
LGTM

PR #370, opened 2017-09-12
Memory Preallocation: Add support for pre-allocation

mcastelino:
Add support for pre-allocating all the RAM for the VM. This will reserve the RAM at VM startup.
Also add support for realtime and mlock.

Enabling prealloc, realtime and mlock gives the lowest latecy.
Disabling prealloc, enabling realtime and disabling mlock gives highest density as the memory will be swapped.

This PR depends on vendoring the updated qemu package. https://github.com/01org/ciao/pull/1428 and
https://github.com/01org/ciao/pull/1430

### DIFF
--- a/hypervisor.go
+++ b/hypervisor.go
@@ -151,6 +151,9 @@ type HypervisorConfig struct {
 	// DefaultMem specifies default memory size in MiB for the VM.
 	// Pod configuration VMConfig.Memory overwrites this.
 	DefaultMemSz uint32
+
+	// MemPrealloc specifies if the memory should be pre-allocated
+	MemPrealloc bool
 }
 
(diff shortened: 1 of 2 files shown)


egernst:
Great - I'm curious what some of the performance impacts for this will be (ie, start time for qemu pending the size of the backed memory).  This looks good to me, though your commit message could use a space...

mcastelino:
@egernst this is disabled by default. This is just an useful know to have available specially when trying to root cause performance issues. This helps eliminate some of the variability.

egernst:
@mcastelino -- yeah, understood.  Still am curious though!  Only I see is need to s/pre-allocatingall/pre-allocating all/

sameo:
LGTM

PR #371, opened 2017-09-13
vendor: Last re-vendoring before 1.0.0

sameo:
In particular we need the ciao memory allocation changes.


Gopkg.lock
```diff
@@ -68,40 +68,46 @@
...
-  revision = "adbae1b6b6fb4b02448a0fc0dbbc9ba2b95b294d"
+  packages = ["curve25519","ed25519","ed25519/internal/edwards25519","ssh","ssh/terminal"]
+  revision = "faadfbdc035307d901e69eea569f5dda451a3ee3"
 
 [[projects]]
   branch = "master"
   name = "golang.org/x/sys"
-  packages = ["unix"]
-  revision = "f7928cfef4d09d1b080aa2b6fd3ca9ba1567c733"
+  packages = ["unix","windows"]
```
jcvenegas: is this windows package need?

jcvenegas: Or was automatically added by dep?

sameo: Dep kept it even after a prune.

jcvenegas: ok, thanks for clarify

PR #376, opened 2017-09-20
Nesting: Disable nesting checks

mcastelino:
This is used to disable the customizations performed when
we are running on top of another VM. This can be used when
running on top of KVM to obtain the best performance.

This check should not be disabled when running on hypervisors
other than KVM. Doing so will result in failure to launch.

### DIFF
--- a/hypervisor.go
+++ b/hypervisor.go
@@ -162,6 +162,10 @@ type HypervisorConfig struct {
 	// Realtime=true and Mlock=false, allows for swapping out of VM memory
 	// enabling higher density
 	Mlock bool
+
+	// DisableNestingChecks is used to override customizations performed
+	// when running on top of another VMM.
+	DisableNestingChecks bool
 }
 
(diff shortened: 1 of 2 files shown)


jodh-intel:
lgtm

sboeuf:
LGTM

sameo:
@mcastelino Could we be a little more dynamic here and try to detect if we're running on top of KVM or not ? When running as a KVM guest, calling into `CPUID` should give us the `KVMKVMKVM` string iirc. I know this would mean adding arch specific code into virtcontainers or ciao/qemu but it would be worth it imho.

mcastelino:
@sameo this is a force disable. I want to be able to run without checking on any hypervisor. The nesting changes we do are due a bug in KVM (in the guest KVM), and hence it does not matter which hypervisor we are running on. Once the KVM bug is fixed, we should always set this, till we remove the nesting specific code.

PR #380, opened 2017-09-22
Vfio passthrough

amshinde:
This add support for passing in vfio devices to the VM.
Requires ciao PR to be merged first:
https://github.com/ciao-project/ciao/pull/1444

### DIFF
--- a/container.go
+++ b/container.go
@@ -494,6 +494,11 @@ func (c *Container) start() error {
 		}
 	}
 
+	// Attach devices
+	if err := c.attachDevices(); err != nil {
+		return err
+	}
+
 	if err = c.pod.agent.startContainer(*(c.pod), *c); err != nil {
 		virtLog.Error("Failed to start container: ", err)
 
@@ -566,6 +571,10 @@ func (c *Container) stop() error {
 
 	err = c.pod.agent.stopContainer(*(c.pod), *c)
 	if err != nil {
+		return err
+	}
+
+	if err = c.detachDevices(); err != nil {
 		return err
 	}
 
@@ -784,6 +793,26 @@ func (c *Container) removeDrive() (err error) {
 
 		if err := c.pod.hypervisor.hotplugRemoveDevice(drive, blockDev); err != nil {
 			virtLog.Errorf("Error while unplugging block device : %s", err)
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (c *Container) attachDevices() error {
+	for _, device := range c.devices {
+		if err := device.attach(c.pod.hypervisor); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (c *Container) detachDevices() error {
+	for _, device := range c.devices {
+		if err := device.detach(c.pod.hypervisor); err != nil {
 			return err
 		}
 	}
--- a/container.go
+++ b/container.go
@@ -79,6 +79,9 @@ type ContainerConfig struct {
 	Annotations map[string]string
 
 	Mounts []Mount
+
+	// Devices are devices that must be available within the container.
+	Devices []Device
 }
 
 // valid checks that the container configuration is valid.
@@ -115,6 +118,8 @@ type Container struct {
 	process Process
 
 	mounts []Mount
+
+	devices []Device
 }
 
 // ID returns the container identifier string.
@@ -326,6 +331,7 @@ func newContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 		state:         State{},
 		process:       Process{},
 		mounts:        contConfig.Mounts,
+		devices:       contConfig.Devices,
 	}
 
 	state, err := c.pod.storage.fetchContainerState(c.podID, c.id)
--- a/device.go
+++ b/device.go
@@ -154,7 +154,7 @@ func isBlock(hostPath string) bool {
 }
 
 func createDevice(devInfo DeviceInfo) Device {
-	path := devInfo.ContainerPath
+	path := devInfo.HostPath
 
 	if isVFIO(path) {
 		return newVFIODevice(devInfo)
--- a/device.go
+++ b/device.go
@@ -18,6 +18,7 @@ package virtcontainers
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -25,6 +26,10 @@ import (
 
 	"github.com/go-ini/ini"
 )
+
+// Defining this as a variable instead of a const, to allow
+// overriding this in the tests.
+var sysIOMMUPath = "/sys/kernel/iommu_groups"
 
 var sysDevPrefix = "/sys/dev"
 
@@ -88,6 +93,30 @@ func newVFIODevice(devInfo DeviceInfo) *VFIODevice {
 }
 
 func (device *VFIODevice) attach(h hypervisor) error {
+	vfioGroup := filepath.Base(device.HostPath)
+	iommuDevicesPath := filepath.Join(sysIOMMUPath, vfioGroup, "devices")
+
+	deviceFiles, err := ioutil.ReadDir(iommuDevicesPath)
+	if err != nil {
+		return err
+	}
+
+	// Pass all devices in iommu group
+	for _, deviceFile := range deviceFiles {
+
+		//Get bdf of device eg 0000:00:1c.0
+		deviceBDF, err := getBDF(deviceFile.Name())
+		if err != nil {
+			return err
+		}
+
+		device.BDF = deviceBDF
+
+		if err := h.addDevice(device, vfioDev); err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 
@@ -227,4 +256,17 @@ func NewDevice(path, devType string, major, minor int64, fileMode *os.FileMode, 
 
 	device := createDevice(devInfo)
 	return device, nil
+}
+
+// bdf returns the BDF of pci device
+// Expected input strng format is [<domain>]:[<bus>][<slot>].[<func>] eg. 0000:02:10.0
+func getBDF(deviceSysStr string) (string, error) {
+	tokens := strings.Split(deviceSysStr, ":")
+
+	if len(tokens) != 3 {
+		return "", fmt.Errorf("Incorrect number of tokens found while parsing bdf for device : %s", deviceSysStr)
+	}
+
+	tokens = strings.SplitN(deviceSysStr, ":", 2)
+	return tokens[1], nil
 }
--- a/device.go
+++ b/device.go
@@ -1,0 +1,230 @@
+//
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package virtcontainers
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/go-ini/ini"
+)
+
+var sysDevPrefix = "/sys/dev"
+
+var blockPaths = []string{
+	"/dev/sd",   //SCSI block device
+	"/dev/hd",   //IDE block device
+	"/dev/vd",   //Virtual Block device
+	"/dev/ida/", //Compaq Intelligent Drive Array devices
+}
+
+const (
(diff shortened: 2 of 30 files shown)


device.go
```diff
@@ -48,3 +54,27 @@ type VFIODevice struct {
 	// Bus-Device-Function for pci device
 	BDF string
 }
+
+// isVFIODeviceGroup checks if the device provided is a vfio group
+func (d *Device) isVFIODeviceGroup() bool {
+	vfioPath := "/dev/vfio/"
```
mcastelino: Can you also make this a const

pod.go
```diff
@@ -1162,6 +1168,44 @@ func togglePausePod(podID string, pause bool) (*Pod, error) {
 	return p, nil
 }
 
+func (p *Pod) handleDevices() error {
+	for _, device := range p.devices {
+		// Handle VFIO devices
+		// TODO Add support for handling devices besides VFIO
+
+		if device.isVFIODeviceGroup() {
```
mcastelino: This is the generic group assignment. This will be uncommon but good to have. You can assign a single device in a iommu group to a VM and vfio should hold on to the other devices. So when we implement specific device assignment we should support assigning just one of the devices in a group.

device.go
```diff
@@ -0,0 +1,80 @@
...
+const (
+	sysIOMMUPath = "/sys/kernel/iommu_groups"
+)
+
+// Device that must be available in the container.
+type Device struct {
+	// Full path to the device inside container.
+	Path string
+
+	// Type of device: c, b, u or p
```
jodh-intel: Can you expand this a bit to explain that `c` == character, etc?

Also, what happens if, say, `Type = foo`? I think we need validation on this field (we shouldn't blindly trust what we are provided in `config.json` because (a) security and (b) we don't know if we can support every conceivable value) and unit-tests for all scenarios (all valid values and a few invalid ones).

pod.go
```diff
@@ -330,6 +331,13 @@ type PodConfig struct {
 	// Annotations keys must be unique strings and must be name-spaced
 	// with e.g. reverse domain notation (org.clearlinux.key).
 	Annotations map[string]string
+
+	// Devices is a linux specific configuration, these are devices that
+	// must be available within the container.
+	// Appears as a direct child of "linux" object in the oci json spec.
+	// Note this is different from the devices under "resources".
+	//Devices []CompatOCISpec.Linux.LinuxDevice
```
jodh-intel: Is this comment line meant to be here? If so, it would be better to put the at the end of the "oci json spec" sentence I think.

pod.go
```diff
@@ -330,6 +331,13 @@ type PodConfig struct {
 	// Annotations keys must be unique strings and must be name-spaced
 	// with e.g. reverse domain notation (org.clearlinux.key).
 	Annotations map[string]string
+
+	// Devices is a linux specific configuration, these are devices that
+	// must be available within the container.
+	// Appears as a direct child of "linux" object in the oci json spec.
```
jodh-intel: s/oci/OCI/

pod.go
```diff
@@ -1152,6 +1168,44 @@ func togglePausePod(podID string, pause bool) (*Pod, error) {
 	return p, nil
 }
 
+func (p *Pod) handleDevices() error {
+	for _, device := range p.devices {
+		// Handle VFIO devices
+		// TODO Add support for handling devices besides VFIO
```
jodh-intel: Can you open an issue for this (if there isn't already one) and reference it here?

pod.go
```diff
@@ -1152,6 +1168,44 @@ func togglePausePod(podID string, pause bool) (*Pod, error) {
...
+				}
+
+				vfioDevice := VFIODevice{
+					BDF: deviceBDF,
+				}
+
+				if err := p.hypervisor.addDevice(vfioDevice, vfioDev); err != nil {
+					return err
+				}
+
```
jodh-intel: Extra whitespace.

device.go
```diff
@@ -0,0 +1,80 @@
...
+
+	return false
+}
+
+// bdf returns the BDF of pci device
+// Expected input strng format is [<domain>]:[<bus>][<slot>].[<func>] eg. 0000:02:10.0
+func bdf(deviceSysStr string) (string, error) {
+	tokens := strings.Split(deviceSysStr, ":")
+
+	if len(tokens) != 3 {
```
jodh-intel: This scenario needs a test.

device.go
```diff
@@ -0,0 +1,80 @@
...
+
+// isVFIODeviceGroup checks if the device provided is a vfio group
+func (d *Device) isVFIODeviceGroup() bool {
+	vfioPath := "/dev/vfio/"
+
+	if strings.HasPrefix(d.Path, vfioPath) && len(d.Path) > len(vfioPath) {
+		return true
+	}
+
+	return false
```
jodh-intel: This scenario needs a test.

amshinde: Looks like I forgot to commit the tests. Will do so.

pod.go
amshinde: Added a reference to the issue here

pod.go
amshinde: This comment line should have been deleted. I have made the change.

device.go
amshinde: Tests file added

pod.go
```diff
@@ -330,6 +330,12 @@ type PodConfig struct {
 	// Annotations keys must be unique strings and must be name-spaced
 	// with e.g. reverse domain notation (org.clearlinux.key).
 	Annotations map[string]string
+
+	// Devices is a linux specific configuration, these are devices that
+	// must be available within the container.
+	// Appears as a direct child of "linux" object in the OCI json spec.
+	// Note this is different from the devices under "resources".
+	Devices []Device
```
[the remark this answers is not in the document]
mcastelino: @sboeuf VFIO devices get assigned to the VM and the driver will be loaded against the VFIO device. 

Within the VM in the case of network interfaces we can isolate them using network namespaces.  For other use cases we need to if mount namespaces may provide the required isolation.

However until recently kubernetes did not support --device equivalent. We will need to track the k8s device plugin proposal to see where the device will land up.

https://github.com/kubernetes/community/blob/master/contributors/design-proposals/resource-management/device-plugin.md

POD level or container level.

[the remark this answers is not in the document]
sboeuf: @mcastelino Yes I agree the devices are assigned through the VM, but the potential issue if we describe the devices through the Pod structure is that we won't be able to add new devices for further containers. In my mind, it's very close from what we have with the hotplug of block devices.
Doing this at container level allows for more flexibility since it can be handled through the Container structure for every new container, without the need to check which devices have been already created previously (to avoid trying to recreate an existing one).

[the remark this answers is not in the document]
sboeuf: I think we could even create an interface Device that would be held into the Container structure. That way, we would have a list of Device in the container structure, with each Device being a specific one (Block, GenericDevice, ...). Indeed, I think we could process those devices with the same basic rules, the implementation being specific to each structure of this interface.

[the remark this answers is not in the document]
sameo: Devices are a per Container resource. The device plugin k8s proposal added devices to the CRI `ContainerConfig` structure, see:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/cri/v1alpha1/runtime/api.pb.go#L1596-L1615

So we should make it a per `ContainerConfig` field as well, even though technically it will be a per VM resource for us.

device.go
```diff
@@ -0,0 +1,47 @@
...
+//
+
+package virtcontainers
+
+import (
+	"os"
+)
+
+// Device that must be available in the container.
+type Device struct {
```
sameo: I agree with @sboeuf proposal here, I think we should have a `Device` interface and device.go should implement a Device factory based on a minimal set of information it gets about the device we want to attach. We need to know what is the host path of the device, its container path and DAC permissions we want to apply to the container path.
We will get all of this (and more) from an OCI spec.

From device.go, based on the host path, we should infer that we're dealing with a VFIO device, a block device or something else and then create and return the right `Device` interface implementation. From container.go, your `handleDevices()` routine would then only go through all the container devices and call into `Device.attach()`.

sameo: And I think the `Device` interface should be:

```Go
type Device interface {
      attach()
      detach()
}
```

pod.go
```diff
@@ -1161,6 +1167,44 @@ func togglePausePod(podID string, pause bool) (*Pod, error) {
 	return p, nil
 }
 
+func (p *Pod) handleDevices() error {
```
sameo: Can we please call that one `attachDevices()` ? And this one should go through all its containers and call into an `Container.attachDevices()` as well.

amshinde:
@sameo  Can you take a look at this. I have moved devices to containers and added a factory method for creating specific device implementations.

pkg/oci/utils.go
```diff
@@ -193,6 +193,50 @@ func containerMounts(spec CompatOCISpec) []vc.Mount {
...
+	if !contains(allowedDeviceTypes, d.Type) {
+		return nil, fmt.Errorf("Unexpected Device Type %s for device %s", d.Type, d.Path)
+	}
+
+	if d.Path == "" {
+		return nil, fmt.Errorf("Path cannot be empty for device")
+	}
+
+	return vc.NewDevice(d.Path, d.Type, d.Major, d.Minor, d.FileMode, *d.UID, *d.GID)
+	//return vc.NewDevice(d.Path, d.Type, d.Major, d.Minor, *d.UID, *d.GID)
```
devimc: debug line?

amshinde: fixed

container.go
```diff
@@ -79,6 +79,9 @@ type ContainerConfig struct {
 	Annotations map[string]string
 
 	Mounts []Mount
+
+	// Devices are devices that must be available within the container.
+	Devices []Device
```
sboeuf: Store the `Device` structure won't go well since it is an interface. I mean that marshal it into a json will work but if we want to retrieve it, we won't be able to determine what kind of device implementation is behind this. This should be another type here, with all the infos we need to rebuild this structure from scratch. I think this is something we're gonna need in order to detach properly the device.

device.go
```diff
@@ -223,3 +252,16 @@ func NewDevice(path, devType string, major, minor int64, fileMode *os.FileMode,
 	device := createDevice(devInfo)
 	return device, nil
 }
+
+// bdf returns the BDF of pci device
+// Expected input strng format is [<domain>]:[<bus>][<slot>].[<func>] eg. 0000:02:10.0
+func bdf(deviceSysStr string) (string, error) {
```
sboeuf: I would prefer `getDevBDF()` or `getBDF()` here since I think `bdf()` is a bit short and confusing for the reading of the code.

device.go
```diff
@@ -0,0 +1,267 @@
...
+
+		if err := h.addDevice(device, vfioDev); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (device *VFIODevice) detach(h hypervisor) error {
```
sboeuf: Why is this not implemented ?

device.go
```diff
@@ -0,0 +1,267 @@
...
+	DeviceInfo
+}
+
+func newBlockDevice(devInfo DeviceInfo) *BlockDevice {
+	d := &BlockDevice{}
+	d.DeviceInfo = devInfo
+	return d
+}
+
+func (device *BlockDevice) attach(h hypervisor) error {
```
sboeuf: Why is this not implemented ? My idea would be that `addDrive()` should be called from there, right ? I mean we are trying to manage the different type of devices through an interface, meaning that we want to remove the specific calls from `container.start()` dedicated to block devices, right ?

amshinde: I mentioned in the PR that I will be handling just vfio devices here. Handling block devices is a separate feature and should be done separately. It involves quite a few cases to handle and test.

device.go
amshinde: Because nothing needs to be done here for this simple case of having a vfio group already provided. We are simply passing this device to qemu, that will act as a user level driver for the device. In the second stage that I will be implementing next, if we unbind the device from its current driver and assign it to vfio, we will need to implement things in detach.

device.go
amshinde: ok. I can make that change.

container.go
amshinde: Yes, maybe we will need to store a device type constant/string explicitly for the device for unmarshalling. Again, would like to tackle the unmarshalling in my next PR, where it is actually needed to store devices to detach them.

device.go
```diff
@@ -0,0 +1,267 @@
...
+	"github.com/go-ini/ini"
+)
+
+// Defining this as a variable instead of a const, to allow
+// overriding this in the tests.
+var sysIOMMUPath = "/sys/kernel/iommu_groups"
+
+var sysDevPrefix = "/sys/dev"
+
+var blockPaths = []string{"/dev/sd", "/dev/hd", "/dev/vd", "/dev/ida/"}
```
jodh-intel: It might be worth splitting this across multiple lines with a comment showing each type of block device supported (SCSI, IDE, virtual, Compaq Intelligent Drive Array?)

sboeuf:
@jodh-intel go ahead and merge it if you're okay with the latest version of this PR.

Issue #390, opened 2017-10-02
vm: increase timeouts for VM start and stop


grahamwhaley:
The VM start and stop code has timeouts set to 1s presently.
These (at least the stop one) have been seen to fire on 'busy systems', and 1s is a little optimistic under extreme circumstances - bump them up - and consider if we just use a large hard coded number (like if it has not come back in 60s then, well, we really have an issue), or if we scale it somehow against the number of running VMs.

https://github.com/containers/virtcontainers/blob/master/qemu.go#L684
https://github.com/containers/virtcontainers/blob/master/pod.go#L722

PR #392, opened 2017-10-03
pod: startup: Wait longer for the VM to start

grahamwhaley:
VMs can take a little while to get going depending on the
circumstances - such as a very busy or a large (cores and RAM)
VM - 1s is not long enough to wait in such cases. Bump the
startup timeout (it is better to wait 'too long' to realise
something broke, rather than firing false failures)

For now we make the concious decision to leave the VM shutdown
timeout at 1s.

Fixes: #390

Signed-off-by: Graham Whaley <graham.whaley@intel.com>

### DIFF
--- a/pod.go
+++ b/pod.go
@@ -707,6 +707,9 @@ func (p *Pod) startSetState() error {
 func (p *Pod) startVM(netNsPath string) error {
 	vmStartedCh := make(chan struct{})
 	vmStoppedCh := make(chan struct{})
+	const timeout = time.Duration(10) * time.Second
+
+	virtLog.Info("Starting VM")
 
 	go func() {
 		p.network.run(netNsPath, func() error {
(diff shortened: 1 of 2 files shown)


pod.go
```diff
@@ -719,7 +719,7 @@ func (p *Pod) startVM(netNsPath string) error {
 	select {
 	case <-vmStartedCh:
 		break
-	case <-time.After(time.Second):
+	case <-time.After(10 * time.Second):
```
jodh-intel: This looks fine. Could you add another commit to add a `virtLog.Info("Waiting for %d seconds for VM to start)` call? It might also be worth logging the timeout in the returned error below.

sboeuf: Yes I agree that it would be nice to have the timeout error returned somewhere. And also, what do you think about putting an extra debug comment about the time elapsed in case of success, just to make sure we really have some cases taking more time ?
BTW, you should do
```go
case <-time.After(time.Duration(10) * time.Second)
```
since I think `time.After` expect a `time.Duration` type.

grahamwhaley:
@sboeuf @jodh-intel updated with more/better logs and const timeouts - pls re-review.
I moved to using time.Duration(x), even though there is a spread of using Duration or not with After() calls through the code base - seems to work either way.

sboeuf:
LGTM

sameo:
LGTM

PR #393, opened 2017-10-03
Use macvlan to connect the VM to the container network

mcastelino:
Networking: Add support for multi-queue mactap

Provide multiple methods to connect the Virtual machine
to the container network. The current implementation allows
this to be chosen at a node level. In the future we can
enhance this to be dynamic, where the container interface
type is used to determine the optimal interconnection method.

Add support for multi-queue macvtap as an alternate means to
connect the container network interface to the virtual machine.

Depends on: https://github.com/ciao-project/ciao/pull/1510
Depends on: https://github.com/containers/virtcontainers/issues/410

### DIFF
--- a/network.go
+++ b/network.go
@@ -19,9 +19,11 @@ package virtcontainers
 import (
 	"errors"
 	"fmt"
+	"math/rand"
 	"net"
 	"os"
 	"runtime"
+	"time"
 
 	types "github.com/containernetworking/cni/pkg/types/current"
 	"github.com/containernetworking/plugins/pkg/ns"
@@ -30,10 +32,40 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-// Introduces constants related to network routes.
+// NetInterworkingModel defines the network model connecting
+// the network interface to the virtual machine.
+type NetInterworkingModel int
+
+const (
+	// ModelBridged uses a linux bridge to interconnect
+	// the container interface to the VM. This is the
+	// safe default that works for most cases except
+	// macvlan and ipvlan
+	ModelBridged NetInterworkingModel = iota
+
+	// ModelMacVtap can be used when the Container network
+	// interface can be bridged using macvtap
+	ModelMacVtap
+
+	// ModelEnlightened can be used when the Network plugins
+	// are enlightened to create VM native interfaces
+	// when requested by the runtime
+	// This will be used for vethtap, macvtap, ipvtap
+	ModelEnlightened
+)
+
+// DefaultNetInterworkingModel is a package level default
+// that determines how the VM should be connected to the
+// the container network interface
+var DefaultNetInterworkingModel = ModelMacVtap
+
+// Introduces constants related to networking
 const (
 	defaultRouteDest  = "0.0.0.0/0"
 	defaultRouteLabel = "default"
+	defaultFilePerms  = 0600
+	defaultQlen       = 1500
+	defaultQueues     = 8
 )
 
 type netIfaceAddrs struct {
@@ -45,14 +77,17 @@ type netIfaceAddrs struct {
 type NetworkInterface struct {
 	Name     string
 	HardAddr string
+	Addrs    []netlink.Addr
 }
 
-// NetworkInterfacePair defines a pair between TAP and virtual network interfaces.
+// NetworkInterfacePair defines a pair between VM and virtual network interfaces.
 type NetworkInterfacePair struct {
 	ID        string
 	Name      string
 	VirtIface NetworkInterface
 	TAPIface  NetworkInterface
+	NetInterworkingModel
+	VMFds []*os.File
 }
 
 // NetworkConfig is the network configuration related to a network.
@@ -159,7 +194,7 @@ func runNetworkCommon(networkNSPath string, cb func() error) error {
 func addNetworkCommon(pod Pod, networkNS *NetworkNamespace) error {
 	err := doNetNS(networkNS.NetNsPath, func(_ ns.NetNS) error {
 		for idx := range networkNS.Endpoints {
-			if err := bridgeNetworkPair(&(networkNS.Endpoints[idx].NetPair)); err != nil {
+			if err := xconnectVMNetwork(&(networkNS.Endpoints[idx].NetPair), true); err != nil {
 				return err
 			}
 		}
@@ -176,7 +211,7 @@ func addNetworkCommon(pod Pod, networkNS *NetworkNamespace) error {
 func removeNetworkCommon(networkNS NetworkNamespace) error {
 	return doNetNS(networkNS.NetNsPath, func(_ ns.NetNS) error {
 		for _, endpoint := range networkNS.Endpoints {
-			err := unBridgeNetworkPair(endpoint.NetPair)
+			err := xconnectVMNetwork(&(endpoint.NetPair), false)
 			if err != nil {
 				return err
 			}
@@ -199,6 +234,22 @@ func createLink(netHandle *netlink.Handle, name string, expectedLink netlink.Lin
 		newLink = &netlink.Tuntap{
 			LinkAttrs: netlink.LinkAttrs{Name: name},
 			Mode:      netlink.TUNTAP_MODE_TAP,
+		}
+	case (&netlink.Macvtap{}).Type():
+		qlen := expectedLink.Attrs().TxQLen
+		if qlen <= 0 {
+			qlen = defaultQlen
+		}
+		newLink = &netlink.Macvtap{
+			Macvlan: netlink.Macvlan{
+				Mode: netlink.MACVLAN_MODE_BRIDGE,
+				LinkAttrs: netlink.LinkAttrs{
+					Index:       expectedLink.Attrs().Index,
+					Name:        name,
+					TxQLen:      qlen,
+					ParentIndex: expectedLink.Attrs().ParentIndex,
+				},
+			},
 		}
 	default:
 		return nil, fmt.Errorf("Unsupported link type %s", expectedLink.Type())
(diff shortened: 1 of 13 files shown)


network.go
```diff
@@ -200,6 +226,13 @@ func createLink(netHandle *netlink.Handle, name string, expectedLink netlink.Lin
 			LinkAttrs: netlink.LinkAttrs{Name: name},
 			Mode:      netlink.TUNTAP_MODE_TAP,
 		}
+	case (&netlink.Macvtap{}).Type():
+		newLink = &netlink.Macvtap{
+			LinkAttrs:   netlink.LinkAttrs{Name: name},
+			Mode:        netlink.MACVLAN_MODE_BRIDGE,
+			TxQLen:      500,
```
jodh-intel: Why `500`? Would an admin ever need to change this?

network.go
```diff
@@ -237,6 +270,89 @@ func getLinkByName(netHandle *netlink.Handle, name string, expectedLink netlink.
...
 }
 
+func connectVMNetwork(netPair *NetworkInterfacePair) error {
+	switch DefaultNetInterworkingModel {
+	case ModelBridged:
+		return bridgeNetworkPair(netPair)
+	case ModelMacVtap:
+		return tapNetworkPair(netPair)
+	case ModelEnlightened:
+		return fmt.Errorf("Unsupported networking model")
```
jodh-intel: Can you add `DefaultNetInterworkingModel` to the error here and in the error below. Also, due to the similarity of this function and `disconnectVMNetwork`, I wonder if it would be worth making both these functions call a new function that both validates `netPair` and either calls the "connect" or "disconnect" function:

```
func toggleVMNetwork(netPair *NetworkInterfacePair, connect bool) error
```

In fact, with the above function, we wouldn't really need `connectVMNetwork` and `disconnectVMNetwork` as the intent would be clear from the `connect` param.

network.go
```diff
@@ -237,6 +270,89 @@ func getLinkByName(netHandle *netlink.Handle, name string, expectedLink netlink.
...
+func tapNetworkPair(netPair *NetworkInterfacePair) error {
+	netHandle, err := netlink.NewHandle()
+	if err != nil {
+		return err
+	}
+	defer netHandle.Delete()
+
+	vethLink, err := getLinkByName(netHandle, netPair.VirtIface.Name, &netlink.Veth{})
+	if err != nil {
+		return fmt.Errorf("Could not get veth interface: %s", err)
```
jodh-intel: Will the error include details from `netPair` or is it worth adding this to the error (same question to the errors returned below)?

network.go
```diff
@@ -237,6 +270,89 @@ func getLinkByName(netHandle *netlink.Handle, name string, expectedLink netlink.
...
+	if err != nil {
+		return fmt.Errorf("Could not create TAP interface: %s", err)
+	}
+
+	// Save the veth MAC address to the TAP so that it can later be used
+	// to build the hypervisor command line. This MAC address has to be
+	// the one inside the VM in order to avoid any firewall issues. The
+	// bridge created by the network plugin on the host actually expects
+	// to see traffic from this MAC address and not another one.
+	netPair.TAPIface.HardAddr = vethLinkAttrs.HardwareAddr.String()
```
jodh-intel: This sounds pretty significant - does it justify a log call?

network.go
```diff
@@ -237,6 +270,89 @@ func getLinkByName(netHandle *netlink.Handle, name string, expectedLink netlink.
...
+	hardAddr, err := net.ParseMAC(netPair.VirtIface.HardAddr)
+	if err != nil {
+		return err
+	}
+	if err := netHandle.LinkSetHardwareAddr(vethLink, hardAddr); err != nil {
+		return fmt.Errorf("Could not set MAC address %s for veth interface %s: %s",
+			netPair.VirtIface.HardAddr, netPair.VirtIface.Name, err)
+	}
+
+	mcastSnoop := false
```
jodh-intel: I presume this won't ever need to be `true`, but can you add a comment explaining why this value is being passed?

network.go
mcastelino: @jodh-intel this is the default value that has to be set for packets to flow. The C netlink library sets this. The go netlink library does not and hence needs to be set. No the admin will not need to change this.

mcastelino: @jodh-intel changed this to pick the value from the parent. This is the best way to avoid the manifest constant.

network.go
mcastelino: @jodh-intel addressed. 

@amshinde we need to make this even more generic to include SRIOV, VFIO and vhost-user. So having an pair abstraction is not ideal. We need a more generic abstraction that support all models.  I will rework this based on VFIO and vhost-user needs

network.go
mcastelino: This was a workaround for a kernel bug. Let me dig out the bug details and add it here.

network.go
jodh-intel: Great - thanks.

network.go
```diff
@@ -237,6 +279,159 @@ func getLinkByName(netHandle *netlink.Handle, name string, expectedLink netlink.
...
+	if err != nil {
+		return nil, err
+	}
+
+	//mq support
+	for q := 0; q < queues; q++ {
+
+		tapDev := fmt.Sprintf("/dev/tap%d", i)
+
+		f, err := os.OpenFile(tapDev, os.O_RDWR, 0666)
```
jodh-intel: Could you define the perms as a `const` at the top of the file? Also, does it absolutely have to be `0666`?

mcastelino: Will fix this. What perms do we use for all the sockets etc we create in CC runtime?

network.go
mcastelino: I will add more details to both the parent error and here. The message does not provide enough debug information otherwise.

network.go
jodh-intel: The proxy uses `0660`:

- https://github.com/clearcontainers/proxy/blob/master/proxy.go#L655

... but for hypervisor sockets we rely on the defaults provided by qemu fwics (there doesn't appear to be a way to request particular perms from qemu for `chardev` devices). Taking a look at a pod:

```
$ sudo ls -ld /run/virtcontainers                                                                     
d--------- 3 root root 60 Oct  9 09:09 /run/virtcontainers
$ sudo ls -ld /run/virtcontainers/pods
d--------- 4 root root 80 Oct  9 11:15 /run/virtcontainers/pods
$ sudo ls -ld /run/virtcontainers/pods/
total 0
d--------- 3 root root 220 Oct  9 11:15 2128c38aa804ac73045e32c737914a6e6c36e33daa80ba6669d989a72e81652b
d--------- 3 root root 220 Oct  9 09:09 412db0928eccbe5a180a5e4be7577457d25fa845b3504129fec014ab89ae375e
$ sudo ls -l /run/virtcontainers/pods/2128c38aa804ac73045e32c737914a6e6c36e33daa80ba6669d989a72e81652b
total 8
d--------- 2 root root 100 Oct  9 11:15 2128c38aa804ac73045e32c737914a6e6c36e33daa80ba6669d989a72e81652b
srwxr-x--- 1 root root   0 Oct  9 11:15 console.sock
srwxr-x--- 1 root root   0 Oct  9 11:15 ctrl.sock
srwxr-x--- 1 root root   0 Oct  9 11:15 hyper.sock
-rw-r--r-- 1 root root   0 Oct  9 11:15 lock
srwxr-x--- 1 root root   0 Oct  9 11:15 monitor.sock
-rw-r--r-- 1 root root 514 Oct  9 11:15 network.json
-rw-r--r-- 1 root root 149 Oct  9 11:15 state.json
srwxr-x--- 1 root root   0 Oct  9 11:15 tty.sock
$ 
```

Aside: I wonder if we want to drop read perms from those state files (and `lock`).

/cc @grahamwhaley, @sameo.

network.go
```diff
@@ -230,13 +278,191 @@ func getLinkByName(netHandle *netlink.Handle, name string, expectedLink netlink.
...
+		}
 	default:
 		return nil, fmt.Errorf("Unsupported link type %s", expectedLink.Type())
 	}
 
 	return nil, fmt.Errorf("Incorrect link type %s, expecting %s", link.Type(), expectedLink.Type())
 }
 
+func xconnectVMNetwork(netPair *NetworkInterfacePair, connect bool) error {
+	switch DefaultNetInterworkingModel {
```
sboeuf: I think this switch should actually test `netPair.NetInterworkingModel`, and the value assignment of `DefaultNetInterworkingModel` should be done during the `netPair` initialization.

Moreover, I think we should really define a `NetInterConnect` interface with methods `connect` and `disconnect`, with different implementations according to the type of `NetInterworkingModel` in use.

network.go
```diff
@@ -230,13 +278,191 @@ func getLinkByName(netHandle *netlink.Handle, name string, expectedLink netlink.
...
+			return untapNetworkPair(*netPair)
+		}
+	case ModelEnlightened:
+		return fmt.Errorf("Unsupported networking model")
+	default:
+		return fmt.Errorf("Invalid networking model")
+	}
+}
+
+func cleanupFds(fds []*os.File, numFds int) {
```
sboeuf: Because this is very generic, it should go to `utils.go`.

network.go
```diff
@@ -230,13 +278,191 @@ func getLinkByName(netHandle *netlink.Handle, name string, expectedLink netlink.
...
+	}
+
+	if err := netHandle.LinkSetUp(tapLink); err != nil {
+		return fmt.Errorf("Could not enable TAP %s: %s", netPair.TAPIface.Name, err)
+	}
+
+	if err := netHandle.LinkSetUp(vethLink); err != nil {
+		return fmt.Errorf("Could not enable veth %s: %s", netPair.VirtIface.Name, err)
+	}
+
```
sboeuf: Don't you think we could try to factorize most of the code in this function `tapNetworkPair()` since a lot is identical with `bridgeNetworkPair()`

network.go
```diff
@@ -230,13 +278,191 @@ func getLinkByName(netHandle *netlink.Handle, name string, expectedLink netlink.
...
+	// Note: The underlying intefaces need to be up prior to fd creation.
+
+	// Setup the multiqueue fds to be consumed by QEMU as macvtap cannot
+	// be directly connected.
+	// Ideally we want
+	// netdev.FDs, err = createMacvtapFds(netdev.ID, int(config.SMP.CPUs))
```
sboeuf: Why cannot you do that ?

network.go
```diff
@@ -307,6 +533,35 @@ func bridgeNetworkPair(netPair *NetworkInterfacePair) error {
 	return nil
 }
 
+func untapNetworkPair(netPair NetworkInterfacePair) error {
```
sboeuf: Ditto for this function, I think we can factorize a lot with `unBridgeNetworkPair()`.

network.go
```diff
@@ -440,11 +455,17 @@ func tapNetworkPair(netPair *NetworkInterfacePair) error {
 		return fmt.Errorf("Could not enable TAP %s: %s", netPair.TAPIface.Name, err)
 	}
 
+	// Clear the IP addresses from the veth interface to prevent ARP conflict
+	err = clearIPs(vethLink)
```
sboeuf: Please use instead:
```go
if err := clearIPs(vethLink); err != nil {
    ...
```

network.go
```diff
@@ -381,6 +381,21 @@ func createMacVtap(netHandle *netlink.Handle, name string, link netlink.Link) (t
...
 }
 
+func clearIPs(link netlink.Link) error {
+	addrs, err := netlink.AddrList(link, netlink.FAMILY_V4)
+	if err != nil || len(addrs) == 0 {
+		return nil
+	}
+
+	for _, addr := range addrs {
+		err = netlink.AddrDel(link, &addr)
```
sboeuf: Please use instead
```go
if err := netlink.AddrDel(link, &addr); err != nil {
    ...
```

network.go
```diff
@@ -30,10 +32,38 @@ import (
...
-// Introduces constants related to network routes.
+// NetworkInterworkingModel defines the type of container to
+// Virtcontainer network interworking models
+type NetInterworkingModel int
+
+const (
+	// ModelBridged uses a linux bridge to interconnect
+	// the container interface to the VM. This is the
+	// safe default that works for most cases except
+	// macvlan and ipvlan
```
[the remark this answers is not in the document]
sameo: @sboeuf afaik macvtap is a TAP interface for macvlan. So macvlan seems correct here.

network.go
sameo: @sboeuf I agree with both comments here.

network.go
```diff
@@ -230,13 +278,212 @@ func getLinkByName(netHandle *netlink.Handle, name string, expectedLink netlink.
...
+	// Setup the multiqueue fds to be consumed by QEMU as macvtap cannot
+	// be directly connected.
+	// Ideally we want
+	// netdev.FDs, err = createMacvtapFds(netdev.ID, int(config.SMP.CPUs))
+
+	// We do not have global context here, hence a manifest constant
+	// that matches our minimum vCPU configuration
+	// Another option is to defer this to ciao qemu library which does have
+	// global context but cannot handle errors when setting up the network
+	netPair.VmFds, err = createMacvtapFds(tapLink.Attrs().Index, 8)
```
jodh-intel: Can you "surface" that magic number into a `const`?

network.go
```diff
@@ -30,10 +32,38 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-// Introduces constants related to network routes.
+// NetworkInterworkingModel defines the type of container to
+// Virtcontainer network interworking models
+type NetInterworkingModel int
```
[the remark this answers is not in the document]
mcastelino: @sboeuf I would not. The reason is that it is an interworking model. There are cases for macvtap direct and ipvtap and the future veth-tap where we will not be doing any interconnection. We will be directly tapping in. Hence interworking model, instead of interconnection model.

network.go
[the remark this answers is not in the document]
mcastelino: @sboeuf @sameo is right. The CNI/CNM plugins today create macvlan and ipvlan. We need to enhance/extend them the near future to create ipvtap and macvtap. So the comment here is valid. For the ipvtap and macvtap we wil use a different interworking model, where we do not need to create any new interfaces but tap directly into them.

network.go
```diff
@@ -30,10 +32,38 @@ import (
...
+// DefaultNetInterworkingModel is a package level default
+// that determines how the VM should be connected to the
+// the container network interface
+var DefaultNetInterworkingModel NetInterworkingModel = ModelBridged
+
+// Introduces constants related to networking
 const (
 	defaultRouteDest  = "0.0.0.0/0"
 	defaultRouteLabel = "default"
+	defaultFilePerms  = 0600
```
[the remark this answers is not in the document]
mcastelino: @sboeuf that was intentional. Is there a reason to allows someone besides root. If we do end up setting up a group for our use then group would also make sense. We do not seem to have a consistent perms model at this time.

@jodh-intel this is something we need to decide if we will support a model where we create group for virt containers, give permission to kvm etc as part of de-priv the runtime.

network.go
```diff
@@ -47,12 +77,14 @@ type NetworkInterface struct {
...
 }
 
-// NetworkInterfacePair defines a pair between TAP and virtual network interfaces.
+// NetworkInterfacePair defines a pair between VM and virtual network interfaces.
 type NetworkInterfacePair struct {
 	ID        string
 	Name      string
 	VirtIface NetworkInterface
 	TAPIface  NetworkInterface
+	NetInterworkingModel
```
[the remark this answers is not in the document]
mcastelino: @sboeuf there is no need to define a field when there is only one element of the type. It makes the code simpler in some cases as you can directly deference the fields of the embedded type.

network.go
mcastelino: @sameo @sboeuf The netpair as we have defined is not the right abstraction for the future. The netpair is only one manifestation. I was planning to rewrite this code to make it generic so it can handle all types of interfaces. So far I have not decided on the right abstraction for the network interface. 

So I would like to defer this comment to a subsequent PR where I would rewrite this code a bit more so that it can support all the upcoming interfaces we plan to support. I will open an issue about the same.

@sboeuf the same is true for our network scanning and caching logic. I will need to change that too to accommodate interfaces other than veth.  If you notice I had to change the scanning logic to do things differently on the second network scan. The second network scan ideally should not be done as it may see results very different from the first one for some use cases.

network.go
mcastelino: agree

network.go
mcastelino: @sboeuf the ordering is different between the two. If not we get issues like ARP table corruption. I need to see if the bridge+linux code can use the same order or if I can find a consistent order. But for now I would keep them seperate.

network.go
mcastelino: @sboeuf we need more context. I feel we need to pass in global context to all the APIs all the time. This will be needed in the future.

network.go
mcastelino: @jodh-intel ideally this should be a constant at all. This should match the number of vCPUs for that POD/container and capped at a some limit. But for now I will move it up to a constant till we get global context.

network.go
mcastelino: @sameo @sboeuf I added a new issue to track this comment

https://github.com/containers/virtcontainers/issues/420

I need to rewrite a bit of the logic to create a fully qualified endpoint with interfaces vs the current struct.

sameo:
@mcastelino will address the code design comments in follow up PRs. So for now:
LGTM

network.go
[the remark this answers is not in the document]
sboeuf: @mcastelino ok I understand. BTW, a group would be great to de-priv the runtime/qemu but the hard part is about translating this group inside the VM. Some implementation for the agent for sure.

sboeuf:
@mcastelino you have some `go vet` issues here

mcastelino:
@sboeuf this PR works as long as it is rebased on top of what cc-runtime is tracking which is 
e8b0c3a6c9fd7691e37a6aa73ed6af4c42f42ad1

So the question is, how was CI passing on tip of virtcontainers prior to this PR?

However this PR applied on master of virtcontainers will break due to existing 
breaking API changes. So there are changes in virtcontainers that break cc-runtime and this PR rebased on tip is catching it.

To make matters worse we have ciao rename from https://github.com/ciao-project/ciao to https://github.com/ciao-project/ciao

@chavafg how does our CI handle the case where a PR lands on the tip of virtcontainers which has breaking API changes and needs cc-runtime to be changed at the same time to accept it?

We should rethink our vendoring strategy where we are vendoring in two places and there is so much co dependency.

sboeuf:
@mcastelino yeah I understand... The thing is that right now our CI for virtcontainers copies the thing from virtcontainers to the runtime in order to test a PR, but it does not work if this same PR revendor ciao from virtcontainers because those changes are not taken into account...
That's the thing about trying to test upward (I mean testing a consumer of our package), instead of the basic testing of the repo itself. I think in cases like your PR, we should simply make an exception and merge it if we agree on it. It is gonna be properly tested when revendoring virtcontainers into the runtime anyway.

About the vendoring, I don't see the problem. If we don't vendor our virtcontainers library, it's gonna get broken all the time and this is gonna be annoying to maintain. From the runtime perspective, `dep` binary is always gonna be able to make the proper vendoring of virtcontainers even if the runtime is using the same packages than virtcontainers.

BTW, about a case in between, that is when we break compatibility from virtcontainers (and not from a vendored package), I have this PR https://github.com/containers/virtcontainers/pull/454 to solve this.

mcastelino:
@sboeuf  We have few things going wrong

We vendor packages in runtime and virtcontainers and they get vendored at different points in time
And virtcontainers is always ahead of runtime. Which means any significant changes to virtcontainers dependencies will mean our virtcontainers CI which runs runtime tests will not work.

So to fix this we should either
a. Not run runtime CI as part of virtcontainers and wait for the CI of runtime once things are properly vendored
b. Improve our virtcontainers CI so that it does a full virtual revendor prior to run

To be able to do b, it means runtime should _not_ use revision based vendoring for any common dependencies and for virtcontainer dependencies that get pulled into runtime.

However this does not address the case where there are API breaking changes in the virtcontainers (which seems to be the same here). A API breaking change cannot be handled by our PR in any case, as it needs changes to the runtime. To handle API breaking changes we need approach a.

However what I do not get here is, why is the CI failing for my PR, which has not made any API breaking changes and how did the API breaking change get merged into virtcontainers. Maybe I am missing something.

PR #395, opened 2017-10-04
DONOTMERGE api: Do not stop the container when checking for status

sameo:
Stopping an already stopped container or a VM that's shutting down
can lead to blocking connections while we force the proxy to connect
to an existing but unavailable VM.

### DIFF
--- a/api.go
+++ b/api.go
@@ -590,8 +590,12 @@ func statusContainer(pod *Pod, containerID string) (ContainerStatus, error) {
 				}
 
 				if !running {
-					if err := container.stop(); err != nil {
-						return ContainerStatus{}, err
+					err = container.state.validTransition(container.state.State, StateStopped)
+					if err == nil {
+						err = container.setContainerState(StateStopped)
+						if err != nil {
+							return ContainerStatus{}, err
+						}
 					}
 				}
 			}


api.go
```diff
@@ -590,8 +590,12 @@ func statusContainer(pod *Pod, containerID string) (ContainerStatus, error) {
 				}
 
 				if !running {
-					if err := container.stop(); err != nil {
-						return ContainerStatus{}, err
+					err = container.state.validTransition(container.state.State, StateStopped)
+					if err == nil {
+						err = container.setContainerState(StateStopped)
```
jodh-intel: I'm not sure I fully understand this - if the shim is killed/dies but the VM is still running, won't this leave stray processes around? Also, if `Container.stop()` isn't being called, the hot-unplug code won't be called (`c.removeDrive()`)?

sameo: Good point, I'll rework the fix.

sboeuf:
I don't think we could end up in the situation described by this PR. The way the runtime/proxy works is that we have to unregister from the proxy before it closes the connection. The same thing with the VM is that we stop the VM when we delete the pod.
The reason why this patch was solving our issue is because by removing this call to `c.stop()`, it was removing the potential cause of the race condition, that is modifying a container state when other call to the runtime could ask for those informations at the same time.
At least that's my understanding of the issues we got recently :)

sboeuf:
#396 is supposed to fix the root cause. @sameo let me know what you think about it.

PR #396, opened 2017-10-04
api: Fix pod race conditions

sboeuf:
This PR allows to fix the CRI-O testing issues that have been observed recently. The root cause of this is the lack of locking regarding the Pod resource when using `StatusPod()` or `StatusContainer()` functions.

### DIFF
--- a/api.go
+++ b/api.go
@@ -325,6 +325,12 @@ func StatusPod(podID string) (PodStatus, error) {
 		return PodStatus{}, errNeedPodID
 	}
 
+	lockFile, err := lockPod(podID)
+	if err != nil {
+		return PodStatus{}, err
+	}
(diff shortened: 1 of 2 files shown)


sboeuf:
This patch has been tested through this PR: https://github.com/clearcontainers/runtime/pull/677
The CI has been consistently passing (restarted it 3 times), with no failures. And it has been separately tested through an Azure VM where I have run in an infinite loop the CRI-O tests.
This PR solves https://github.com/clearcontainers/runtime/issues/622 and https://github.com/containers/virtcontainers/pull/395 is not needed.

amshinde:
@sboeuf Nice work!

sameo:
Thanks @sboeuf

PR #399, opened 2017-10-05
Implement custom devices marshalling routines and revert RO rootfs support

sameo:
The agent does not fully support RO rootfs

 Unmarshalling device interfaces is not possible with our device structures.
 We need to implement our custom marshalling and unmarshalling routines
 in order to properly create the right devices from a stored container config.

### DIFF
--- a/container.go
+++ b/container.go
@@ -155,8 +155,24 @@ func (c *ContainerConfig) UnmarshalJSON(b []byte) error {
 				}
 
 				switch deviceType {
-				case deviceGeneric:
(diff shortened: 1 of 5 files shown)


amshinde:
@sameo I have pushed a fix for the oci tests here : https://github.com/amshinde/virtcontainers/commit/cb794a6904c0e2ec8eca6749f7994ecc7e09041d

sameo:
@amshinde Pulled, thanks.

sboeuf:
LGTM

PR #403, opened 2017-10-06
qemu: support custom bios

devimc:
In order to support custom BIOS, nofw option
must be removed and the Bios must be set with
the path to it

partially fixes https://github.com/clearcontainers/runtime/pull/687

### DIFF
--- a/hypervisor.go
+++ b/hypervisor.go
@@ -127,6 +127,9 @@ type HypervisorConfig struct {
 	// ImagePath is the guest image host path.
 	ImagePath string
 
+	// FirmwarePath is the bios host path
+	FirmwarePath string
+
 	// HypervisorPath is the hypervisor executable host path.
 	HypervisorPath string
 
--- a/hypervisor.go
+++ b/hypervisor.go
@@ -130,6 +130,9 @@ type HypervisorConfig struct {
 	// FirmwarePath is the bios host path
 	FirmwarePath string
 
+	// MachineAccelerators are machine specific accelerators
+	MachineAccelerators string
+
 	// HypervisorPath is the hypervisor executable host path.
 	HypervisorPath string
 
--- a/qemu.go
+++ b/qemu.go
@@ -87,7 +87,7 @@ var supportedQemuMachines = []ciaoQemu.Machine{
 	},
 	{
 		Type:         QemuQ35,
-		Acceleration: "kvm,kernel_irqchip,nvdimm,nosmm,nosmbus,nosata,nopit,nofw",
+		Acceleration: "kvm,kernel_irqchip,nvdimm,nosmm,nosmbus,nosata,nopit",
 	},
 }
 
@@ -648,6 +648,7 @@ func (q *qemu) createPod(podConfig PodConfig) error {
 		Knobs:       knobs,
 		VGA:         "none",
 		GlobalParam: "kvm-pit.lost_tick_policy=discard",
+		Bios:        podConfig.HypervisorConfig.FirmwarePath,
 	}
 
 	q.qemuConfig = qemuConfig
--- a/qemu.go
+++ b/qemu.go
@@ -567,6 +567,14 @@ func (q *qemu) createPod(podConfig PodConfig) error {
 		return err
 	}
 
+	accelerators := podConfig.HypervisorConfig.MachineAccelerators
+	if accelerators != "" {
+		if !strings.HasPrefix(accelerators, ",") {
+			accelerators = fmt.Sprintf(",%s", accelerators)
+		}
+		machine.Acceleration += accelerators
+	}
+
 	smp := q.setCPUResources(podConfig)
 
 	memory, err := q.setMemoryResources(podConfig)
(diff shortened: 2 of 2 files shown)


devimc:
depends on https://github.com/ciao-project/ciao/pull/1517

sboeuf:
Needs https://github.com/ciao-project/ciao/pull/1517 to land and a revendoring of ciao before we can merge this PR. Also, I bet this will need a fix for our qemu unit tests.

devimc:
@sboeuf the problem with ```nofw``` is that it's not a standard accelerator hence if the user decide to use the version of qemu provided by the distro (fedora/suse/ubuntu etc) he/she will not be able to run clear container with q35.
Also i want to add another option into the configuration file to allow users specify custom accelerators (machine_accelerators="")

mcastelino:
@devimc @sboeuf we need to figure out a generic way for us to handle multiple QEMU variants and the accompanying bios variant and options. However if we do not embed any logic in the code but drive it all from the config file we should be ok. This will mean that certain combinations in the config file may result in the system not being able to boot, but that IMO is an acceptable compromise as changing the qemu and firmware variant is an advanced user configuration anyway.

sameo:
@mcastelino 
> we need to figure out a generic way for us to handle multiple QEMU variants and the accompanying bios variant and options.

As you noted, the simplest way to do so would be to drive it from a configuration file. However, each hypervisor implementation in QEMU should be able to error out very early and loudly if the passed configuration file is invalid. Adding a `valid(config HypervisorConfig)` method to our `hypervisor` interface would be an explicit way to do so.

hypervisor.go
```diff
@@ -127,6 +127,9 @@ type HypervisorConfig struct {
 	// ImagePath is the guest image host path.
 	ImagePath string
 
+	// BiosPath is the bios host path
+	BiosPath string
```
sameo: `FirmwarePath` please.

devimc: Q. why firmware? 

I though bios was right, because the firmware is the binary that allows to the bios to access to different resources present in the host like kernel, cmdline, initrd, etc, through DMA [1] [2]

[1] https://github.com/bonzini/qboot/blob/master/fw_cfg.c#L186-L228
[2] https://github.com/clearcontainers/qemu/blob/master/hw/i386/pc_sysfw.c#L104-L174

devimc:
depends on #406

devimc:
@mcastelino I plan to add a page in the wiki explaining how to specify machine accelerators and custom bios, pros and cons, etc

sboeuf:
Please rebase since #406 has been merged

hypervisor.go
sameo: @devimc They're essentially the same, bios being a term that's generally used for legacy firmware implementations. firmware is more generic.

devimc: ohh ok, make sense

qemu.go
```diff
@@ -87,7 +87,7 @@ var supportedQemuMachines = []ciaoQemu.Machine{
 	},
 	{
 		Type:         QemuQ35,
-		Acceleration: "kvm,kernel_irqchip,nvdimm,nosmm,nosmbus,nosata,nopit,nofw",
+		Acceleration: "kvm,kernel_irqchip,nvdimm",
```
sboeuf: you should factorize this as this is used for all qemu machine types.

qemu.go
```diff
@@ -87,7 +87,7 @@ var supportedQemuMachines = []ciaoQemu.Machine{
 	},
 	{
 		Type:         QemuQ35,
-		Acceleration: "kvm,kernel_irqchip,nvdimm,nosmm,nosmbus,nosata,nopit,nofw",
```
sboeuf: We need to make sure to add the proper things to the runtime when vendoring this change. Indeed, without the extra accelerators params, q35 machine type won't work.

qemu.go
devimc: done

mcastelino:
@sboeuf @sameo @devimc I do not think we can really have a valid function that really works. The reason is allow the hypervisor to be chosen based on path. So we really have no idea what that hypervisor really is and what it is compiled with.  So for a lot of the options like accelerators there is not real way to validate.

We can check of absolutely incorrect stuff like CPU and memory, but for things like hypervisor options adding a valid function will only make things more complicated when an advanced user is setting up the config file to work with a custom hypervisor, which will force a change in the code.

sboeuf:
@mcastelino ok that makes sense. I'd like @sameo opinion before we merge that, but other than that:
LGTM

sameo:
@mcastelino 

> So we really have no idea what that hypervisor really is and what it is compiled with

That's partly true. From the `qemu` implementation we know that we're, well, qemu and we know which machine type we're running on. Based on that we can do some basic validation, assuming that qemu does not break it's cli backward compatibility.

>  an advanced user is setting up the config file to work with a custom hypervisor

For that use case, we can have a `developer` mode turn off by default in the hypervisor section of the runtime.

sameo:
LGTM

mcastelino:
@sameo 

> For that use case, we can have a developer mode turn off by default in the hypervisor section of the runtime.

If we have that option to turn off, then we can add the validation. That will ensure a sufficiently advanced user/deployer of clear containers can customize the VM w/o needing to change code.

PR #405, opened 2017-10-07
Refactor device handling

amshinde:
Refactor the way devices are handled so that we store an invariable config set in the configuration.
This PR moves the Device interfaces from ContainerConfig to Container and create the device implementations at create stage.
Devices are store separately in "devices.json". This simplifies the custom marshalling and unmarshalling to an extent.

### DIFF
--- a/container.go
+++ b/container.go
@@ -17,7 +17,6 @@
 package virtcontainers
 
 import (
-	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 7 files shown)


device.go
```diff
@@ -172,9 +125,12 @@ func (device *VFIODevice) attach(h hypervisor) error {
 
 		device.BDF = deviceBDF
 
-		if err := h.addDevice(device, vfioDev); err != nil {
+		if err := h.addDevice(*device, vfioDev); err != nil {
```
sameo: This is a separate fix, please submit it as a separate patch.

container.go
```diff
@@ -452,6 +356,18 @@ func newContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 		c.mounts = mounts
 	}
 
+	devices, err := c.fetchDevices()
+	if err == nil {
+		c.devices = devices
+	} else {
+		// If devices were not found in storage, create Device implementations
+		// from the configuration.
```
sameo: > If devices were not found in storage

Is that solely to handle the `create` to `start` transition ?
It would look better if we could have the same code logic for devices as we do for mounts, process, etc...i.e. we use the Container config information as the default one and it gets overwritten by what we found on storage.

device.go
```diff
@@ -59,6 +59,7 @@ const (
 type Device interface {
 	attach(hypervisor) error
 	detach(hypervisor) error
+	getDeviceType() string
```
sameo: Just `type() string` please. See https://golang.org/doc/effective_go.html#Getters

amshinde: Changed this to deviceType(). golang isnt happy with type()

device.go
amshinde: Done

container.go
amshinde: done

container.go
```diff
@@ -422,6 +327,11 @@ func newContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 		return &Container{}, fmt.Errorf("Invalid container configuration")
 	}
 
+	devices, err := newDevices(contConfig.DeviceInfos)
```
sboeuf: This is gonna be called all the time, even when we don't need it (when we can actually retrieve devices from the runtime storage). Instead I think we should call this only if we cannot retrieve anything from storage, OR the other solution is to make the difference between the first time this container has been created and all other times when it is only recreated.

amshinde: @sboeuf This is what I was doing earlier, but changed it later.
@sameo Giving it a bit more thought, I like the earlier approach of creating the devices only when fetch from storage was not successful.

filesystem.go
```diff
@@ -48,6 +48,9 @@ const (
 
 	// mountsFileType represents a mount file type
 	mountsFileType
+
+	// devicesFileType represents a mount file type
```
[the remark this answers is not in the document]
amshinde: will fix

filesystem.go
```diff
@@ -175,6 +183,55 @@ func (fs *filesystem) storeFile(file string, data interface{}) error {
...
+			Data: tempJSON,
+		}
+		typedDevices = append(typedDevices, typedDevice)
+	}
+
+	jsonOut, err := json.Marshal(typedDevices)
+	if err != nil {
+		return fmt.Errorf("Could not marshal devices: %s", err)
+	}
+	f.Write(jsonOut)
```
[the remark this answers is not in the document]
amshinde: will fix

container.go
```diff
@@ -81,110 +80,8 @@ type ContainerConfig struct {
...
-
-					c.Devices = append(c.Devices, device)
-				}
-			}
-		}
-	}
-
-	return nil
+	// Device configuration for devices that must be available within the container.
+	DeviceInfos []DeviceInfo
```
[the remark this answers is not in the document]
amshinde: I prefer leaving this as is, we were are storing multiple DeviceInfo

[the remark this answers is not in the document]
sboeuf: fair enough

container.go
sameo: @amshinde It's more efficient, I agree. We can keep it there, but please add some further comments describing why and how we can reach each side of the if switch.

filesystem.go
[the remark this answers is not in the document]
amshinde: Fixed

filesystem.go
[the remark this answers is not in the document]
amshinde: Fixed

container.go
amshinde: Fixed

amshinde:
@sameo @sboeuf I have pushed changes.

sboeuf:
@sameo please merge if you're okay with this PR
LGTM

PR #409, opened 2017-10-10
qemu: improve kernel boot time

devimc:
setting the last pci bus to 0 reduces the kernel
boot time for around 40ms

### DIFF
--- a/qemu.go
+++ b/qemu.go
@@ -186,6 +186,7 @@ var kernelDefaultParams = []Param{
 	{"iommu", "off"},
 	{"cryptomgr.notests", ""},
 	{"net.ifnames", "0"},
+	{"pci", "lastbus=0"},
 }
 
 // kernelDefaultParamsNonDebug is a list of the default kernel
--- a/qemu_test.go
+++ b/qemu_test.go
@@ -58,7 +58,7 @@ func testQemuBuildKernelParams(t *testing.T, kernelParams []Param, expected stri
 	}
 }
 
-var testQemuKernelParamsBase = "root=/dev/pmem0p1 rootflags=dax,data=ordered,errors=remount-ro rw rootfstype=ext4 tsc=reliable no_timer_check rcupdate.rcu_expedited=1 i8042.direct=1 i8042.dumbkbd=1 i8042.nopnp=1 i8042.noaux=1 noreplace-smp reboot=k panic=1 console=hvc0 console=hvc1 initcall_debug iommu=off cryptomgr.notests net.ifnames=0"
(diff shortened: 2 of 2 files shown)


qemu.go
```diff
@@ -165,6 +165,7 @@ var kernelDefaultParams = []Param{
 	{"iommu", "off"},
 	{"cryptomgr.notests", ""},
 	{"net.ifnames", "0"},
+	{"pci", "lastbus=0"},
```
sboeuf: I am wondering if we want that as a default in virtcontainers. You could add this here https://github.com/clearcontainers/runtime/blob/master/Makefile#L83 instead.

grahamwhaley:
Nice. I presume this does not affect our hotplug work - do we plug into bus0 I guess?
You have failing tests on the CI btw - looks like they will also need updating.
lgtm

qemu.go
jodh-intel: @devimc - does this option improve boot speed for all machine types?

jcvenegas:
40ms kernel boot time! that is awesome !

devimc:
@grahamwhaley this patch does not affect native hot plug

qemu.go
devimc: @jodh-intel  for pc just a little, I only can see a big improvement in q35, I guess because of the chipset

mcastelino: @devimc @jodh-intel we used to have this option turned on before in 2.1. Somewhere along the way we removed it. 

But then we did not have hotplug support.

So it should be safe to use it once we confirm it does not cause issues for both VFIO, VFIO hotplug and generic network and storage hotplug.

devimc:
changing state to ```do-not-merge``` till I can verify this change does not break VFIO, VFIO hotplug and generic network and storage hotplug.

devimc:
lol, I can't set ```do-not-merge``` label, so I'll change the tittle

mcastelino:
@devimc can you see if the ACPI block hotplug works. If a k8s POD comes up with this change it means that at least the current hotplug still works.

sboeuf:
@devimc what is the status about this PR ?

devimc:
@sboeuf  @mcastelino  done, vfio cold plug and ACPI hot plug works

mcastelino:
LGTM

devimc:
ready to merge

sboeuf:
@devimc please fix the unit tests.

devimc:
@sboeuf unit tests fixed, thanks

sboeuf:
LGTM

Issue #407, opened 2017-10-09
Move from Semaphore to Jenkins CI


sboeuf:
As we did for all the Clear Containers repos, it would be nice to move from Semaphore CI to Jenkins CI for this repository too. We need to add @clearcontainersbot to the __containers__ organization, and make it as an admin of virtcontainers. That way, it is gonna be able to publish statuses on the different PRs.

PR #419, opened 2017-10-12
ci: Fix call to virtcontainers-setup.sh

chavafg:
Specify to sudo the PATH to use to execute the
virtcontainers-setup.sh script. If we do not specify the PATH,
the system will not find `go`

Fixes: #407

Signed-off-by: Salvador Fuentes <salvador.fuentes@intel.com>


sboeuf:
LGTM

PR #422, opened 2017-10-13
Block device support for --device

amshinde:
Add support for passing block devices to a container with --device. This PR passes block devices to the VM using virtio-block.

### DIFF
--- a/container.go
+++ b/container.go
@@ -857,7 +857,7 @@ func (c *Container) removeDrive() (err error) {
 
 func (c *Container) attachDevices() error {
 	for _, device := range c.devices {
-		if err := device.attach(c.pod.hypervisor); err != nil {
+		if err := device.attach(c.pod.hypervisor, c); err != nil {
 			return err
 		}
 	}
--- a/device.go
+++ b/device.go
@@ -286,11 +286,9 @@ func isVFIO(hostPath string) bool {
 }
 
 // isBlock checks if the device is a block device.
-func isBlock(hostPath string) bool {
-	for _, blockPath := range blockPaths {
-		if strings.HasPrefix(hostPath, blockPath) && len(hostPath) > len(blockPath) {
-			return true
-		}
+func isBlock(devInfo DeviceInfo) bool {
+	if devInfo.DevType == "b" {
+		return true
 	}
 
 	return false
@@ -301,7 +299,7 @@ func createDevice(devInfo DeviceInfo) Device {
 
 	if isVFIO(path) {
 		return newVFIODevice(devInfo)
-	} else if isBlock(path) {
+	} else if isBlock(devInfo) {
 		return newBlockDevice(devInfo)
 	} else {
 		deviceLogger().WithField("device", path).Info("Device has not been passed to the container")
--- a/device.go
+++ b/device.go
@@ -165,6 +165,9 @@ func (device *VFIODevice) deviceType() string {
 type BlockDevice struct {
 	DeviceType string
 	DeviceInfo DeviceInfo
+
+	// Path at which the device appears inside the VM, outside of the container mount namespace.
+	VirtPath string
 }
 
 func newBlockDevice(devInfo DeviceInfo) *BlockDevice {
@@ -200,7 +203,12 @@ func (device *BlockDevice) attach(h hypervisor, c *Container) error {
 	// Increment the block index for the pod. This is used to determine the name
 	// for the block device in the case where the block device is used as container
 	// rootfs and the predicted block device name needs to be provided to the agent.
-	_, err := c.pod.getAndSetPodBlockIndex()
+	index, err := c.pod.getAndSetPodBlockIndex()
+	if err != nil {
+		return err
+	}
+
+	driveName, err := getVirtDriveName(index)
 	if err != nil {
 		return err
 	}
@@ -222,6 +230,7 @@ func (device *BlockDevice) attach(h hypervisor, c *Container) error {
 		device.DeviceInfo.Hotplugged = true
 	}
 
+	device.VirtPath = filepath.Join("/dev", driveName)
 	return nil
 }
 
--- a/device.go
+++ b/device.go
@@ -58,7 +58,7 @@ const (
 
 // Device is the virtcontainers device interface.
 type Device interface {
-	attach(hypervisor) error
+	attach(hypervisor, *Container) error
 	detach(hypervisor) error
 	deviceType() string
 }
@@ -90,6 +90,10 @@ type DeviceInfo struct {
 
 	// id of the device group.
 	GID uint32
+
+	// Hotplugged is used to store device state indicating if the
+	// device was hotplugged.
+	Hotplugged bool
 }
 
 // VFIODevice is a vfio device meant to be passed to the hypervisor
@@ -111,7 +115,7 @@ func newVFIODevice(devInfo DeviceInfo) *VFIODevice {
 	}
 }
 
-func (device *VFIODevice) attach(h hypervisor) error {
+func (device *VFIODevice) attach(h hypervisor, c *Container) error {
 	vfioGroup := filepath.Base(device.DeviceInfo.HostPath)
 	iommuDevicesPath := filepath.Join(sysIOMMUPath, vfioGroup, "devices")
 
@@ -166,7 +170,7 @@ func newBlockDevice(devInfo DeviceInfo) *BlockDevice {
 	}
 }
 
-func (device *BlockDevice) attach(h hypervisor) error {
+func (device *BlockDevice) attach(h hypervisor, c *Container) error {
 	return nil
 }
 
@@ -191,7 +195,7 @@ func newGenericDevice(devInfo DeviceInfo) *GenericDevice {
 	}
 }
 
-func (device *GenericDevice) attach(h hypervisor) error {
+func (device *GenericDevice) attach(h hypervisor, c *Container) error {
 	return nil
 }
 
--- a/device.go
+++ b/device.go
@@ -186,7 +186,7 @@ func makeBlockDevIDForHypervisor(deviceID string) string {
 	return devID
 }
 
-func (device *BlockDevice) attach(h hypervisor, c *Container) error {
+func (device *BlockDevice) attach(h hypervisor, c *Container) (err error) {
 	randBytes, err := generateRandomBytes(8)
 	if err != nil {
 		return err
@@ -204,6 +204,13 @@ func (device *BlockDevice) attach(h hypervisor, c *Container) error {
 	// for the block device in the case where the block device is used as container
 	// rootfs and the predicted block device name needs to be provided to the agent.
 	index, err := c.pod.getAndSetPodBlockIndex()
+
+	defer func() {
+		if err != nil {
+			c.pod.decrementPodBlockIndex()
+		}
+	}()
+
 	if err != nil {
 		return err
 	}
@@ -219,14 +226,16 @@ func (device *BlockDevice) attach(h hypervisor, c *Container) error {
 	// devices are fixed. After that we need to move towards  hotplugging all devices.
 	// See https://github.com/containers/virtcontainers/issues/444
 	if c.state.State == "" {
-		if err := h.addDevice(drive, blockDev); err != nil {
+		if err = h.addDevice(drive, blockDev); err != nil {
 			return err
 		}
+
 		device.DeviceInfo.Hotplugged = false
 	} else {
-		if err := h.hotplugAddDevice(drive, blockDev); err != nil {
+		if err = h.hotplugAddDevice(drive, blockDev); err != nil {
 			return err
 		}
+
 		device.DeviceInfo.Hotplugged = true
 	}
 
--- a/device.go
+++ b/device.go
@@ -17,6 +17,7 @@
 package virtcontainers
 
 import (
+	"encoding/hex"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -94,6 +95,9 @@ type DeviceInfo struct {
 	// Hotplugged is used to store device state indicating if the
 	// device was hotplugged.
 	Hotplugged bool
+
+	// ID for the device that is passed to the hypervisor.
+	ID string
 }
 
 // VFIODevice is a vfio device meant to be passed to the hypervisor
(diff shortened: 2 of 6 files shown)


device.go
```diff
@@ -158,11 +166,55 @@ func newBlockDevice(devInfo DeviceInfo) *BlockDevice {
...
+	// rootfs and the predicted block device name needs to be provided to the agent.
+	_, err := c.pod.getAndSetPodBlockIndex()
+	if err != nil {
+		return err
+	}
+
+	virtLog.Infof("Attaching block device %s", device.DeviceInfo.HostPath)
+	if create {
+		if err := h.addDevice(drive, blockDev); err != nil {
+			return err
```
jodh-intel: Doesn't this error path (and the one below) need to undo the effect of `c.pod.getAndSetPodBlockIndex()`?

device.go
```diff
@@ -89,6 +90,13 @@ type DeviceInfo struct {
 
 	// id of the device group.
 	GID uint32
+
+	// Hotplugged is used to store device state indicating if the
+	// device was hotplugged.
+	Hotplugged bool
+
+	// ID passed on the hypervisor command line.
```
jodh-intel: This comment doesn't seem quite correct given you use this value to construct a new string value for the hypervisor?

device.go
```diff
@@ -158,11 +166,55 @@ func newBlockDevice(devInfo DeviceInfo) *BlockDevice {
 	}
 }
 
-func (device *BlockDevice) attach(h hypervisor) error {
+func (device *BlockDevice) attach(h hypervisor, c *Container, create bool) error {
+	device.DeviceInfo.ID = uuid.Generate().String()
+	devID := fmt.Sprintf("drive-%s", device.DeviceInfo.ID)
```
jodh-intel: I know it's trivial, but to guarantee this doesn't diverge from the format in `detach()`, I'd be tempted to create something like:

```
func makeDevIDForHypervisor(deviceID string) string {
    return fmt.Sprintf("drive-%s", deviceID)
}
```

device.go
```diff
@@ -106,7 +114,7 @@ func newVFIODevice(devInfo DeviceInfo) *VFIODevice {
 	}
 }
 
-func (device *VFIODevice) attach(h hypervisor) error {
+func (device *VFIODevice) attach(h hypervisor, c *Container, create bool) error {
```
jodh-intel: Forgot to say, can you write tests for `attach()` and `detach()`?

device.go
```diff
@@ -158,11 +166,55 @@ func newBlockDevice(devInfo DeviceInfo) *BlockDevice {
...
 func (device BlockDevice) detach(h hypervisor) error {
+	if device.DeviceInfo.Hotplugged {
+		virtLog.Infof("Unplugging block device %s", device.DeviceInfo.HostPath)
+
+		devID := fmt.Sprintf("drive-%s", device.DeviceInfo.ID)
+		drive := Drive{
+			ID: devID,
+		}
+
+		if err := h.hotplugRemoveDevice(drive, blockDev); err != nil {
```
devimc: I think the block device can be cold plugged, why hot plug it?

sboeuf: You need to be able to hotplug it in case you are adding a block device through a new container in case of an existing pod. The VM will already be up and running, and you will need to hotplug this new block device related to this new container. Makes sense ?

device.go
```diff
@@ -24,6 +24,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/01org/ciao/ssntp/uuid"
```
mcastelino: @amshinde @jodh-intel @sameo I think its time we moved the uuid implementation to our own utils. That way once ciao-qemu is made its own repo we do not pull in rest of ciao.

device.go
devimc: @sboeuf yes
one more question, I not able to see the block device inside the container, I guess there is something missing in the agent code (re-mount /dev/?)

device.go
sameo: @mcastelino Yes, agreed.

device.go
```diff
@@ -158,11 +166,55 @@ func newBlockDevice(devInfo DeviceInfo) *BlockDevice {
...
+	}
+
+	virtLog.Infof("Attaching block device %s", device.DeviceInfo.HostPath)
+	if create {
+		if err := h.addDevice(drive, blockDev); err != nil {
+			return err
+		}
+		device.DeviceInfo.Hotplugged = false
+	} else {
+		if err := h.hotplugAddDevice(drive, blockDev); err != nil {
```
devimc: be aware of bus option, if you  do not specify a bus qemu will hot plug the device directly in the PCI, for pc that's ok because it's ACPI based, but for q35 this will fail, you have to specify a pcie root port or a downstream port.
This will become a problem since we want to support both q35 and pc

sboeuf: Thanks for noticing this @devimc !
That's important that we stay compatible with both so that we can switch back if needed, and also for performance comparison. Anyway, I think we all agree we need something more modular to support both machine types.

device.go
amshinde: @mcastelino Yup, I will move over the uuid implementation.

container.go
```diff
@@ -517,7 +529,7 @@ func (c *Container) start() error {
 	}
 
 	// Attach devices
-	if err := c.attachDevices(); err != nil {
+	if err := c.attachDevices(false); err != nil {
```
devimc: all devices including vfio will be hot plugged?

device.go
amshinde: @jodh-intel I have updated the comment.

container.go
amshinde: Yes, although we havent yet added support for vfio hotplug.

device.go
amshinde: Yes, I have updated the tests for block device attach,

device.go
amshinde: @jodh-intel If this errors out, then we return the error and the create or start phase simply fails. Unless we want to ignore this error and continue further, we dont need to recover and undo the effect of `c.pod.getAndSetPodBlockIndex()`

device.go
amshinde: @devimc Lets add support for q35 in a separate PR.

device.go
amshinde: Now that I am passing the mount information to the agent in a following commit, this will be resolved, as the agent will handle mounting the devices passed.

amshinde:
@jodh-intel @mcastelino Can you take a look at this?

amshinde:
Related to https://github.com/clearcontainers/agent/pull/134.
This PR once merged needs to be venored in proxy as well.

container.go
devimc: Have you tried hot-plugging the vfio device in a pci bridge?

device.go
devimc: @amshinde @sboeuf  yep, working on that :)

amshinde:
@sameo @sboeuf Can you take a look at this?

device.go
mcastelino: @devimc @amshinde let us first add support in the current ACPI implementation. Let the q35 land seperately.

container.go
```diff
@@ -824,9 +824,9 @@ func (c *Container) removeDrive() (err error) {
 	return nil
 }
 
-func (c *Container) attachDevices() error {
+func (c *Container) attachDevices(create bool) error {
```
sameo: Isn't `container.state.State` enough to understand if the container is created or if it's running already ?

container.go
```diff
@@ -824,9 +824,9 @@ func (c *Container) removeDrive() (err error) {
...
 	for _, device := range c.devices {
-		if err := device.attach(c.pod.hypervisor); err != nil {
+		if err := device.attach(c.pod.hypervisor, c, create); err != nil {
```
sameo: Here as well, you may only need to pass the container pointer and infer from its state if it's already running (thus hotplug) or not.

pod.go
```diff
@@ -1163,9 +1163,9 @@ func togglePausePod(podID string, pause bool) (*Pod, error) {
 	return p, nil
 }
 
-func (p *Pod) attachDevices() error {
+func (p *Pod) attachDevices(create bool) error {
```
sameo: Same comment here: `pod.state.State` should tell you about the pod being already created or not.

device.go
sameo: @amshinde What happens if start fails but since the pod is created one tries to start it again ? Will the pod block index be properly set ?

container.go
```diff
@@ -97,6 +97,16 @@ func (c *ContainerConfig) valid() bool {
...
 
+// FsmapEntry describes a filesytem map entry related to a container.
+// This needs to be sent to the agent to mount devices/files at the specified location.
+type FsmapEntry struct {
+	Source       string
+	Dest         string
+	ReadOnly     string
+	Permissions  string
+	AbsolutePath bool
+}
```
sameo: This structure is quite hyperstart specific. Can't we re-use the `Mount` structure and infer the `AbsolutePath` value from it ?

sboeuf: @amshinde yes I agree with @sameo. You should keep the `Container` structure unchanged, and fill the `FsmapDescriptor` structure from hyperstart implementation.

container.go
```diff
@@ -120,6 +130,8 @@ type Container struct {
 	mounts []Mount
 
 	devices []Device
+
+	Fsmap []FsmapEntry
```
sboeuf: No need for a new field here.

device.go
```diff
@@ -24,7 +24,8 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/01org/ciao/ssntp/uuid"
+	"github.com/containers/virtcontainers/pkg/uuid"
```
sboeuf: This should have been changed in previous commit

device.go
amshinde: @sameo If start fails, docker calls "kill" followed by "delete" for the container process immediately if I remember correctly. And no, the pod index will not be properly set in that case.

device.go
```diff
@@ -274,7 +274,7 @@ func createDevice(devInfo DeviceInfo) Device {
 
 	if isVFIO(path) {
 		return newVFIODevice(devInfo)
-	} else if isBlock(path) {
+	} else if isBlock(path) || devInfo.DevType == "b" {
```
sboeuf: This should be part of `isBlock()`. Providing `devInfo` through `isBlock` should be enough.

device.go
sboeuf: @amshinde you think about Docker case, but I don't think you can expect the same behavior from a CRIO case.

container.go
amshinde: @sameo  I didnt want to pollute  the `mounts` provided in the configuration with the ones that we need to create for handling `devices`. I would like to keep a distinction between them, there is still some work pending in handling all the mounts besides bind mounts in container `Mount` list. We also need to pass and handle `Permissions` for block devices.
I could rename this struct to DeviceMap to make it less hyperstart specific. What do you think?

devimc:
lgtm

device.go
amshinde: Fixed in a separate commit.

amshinde:
Agent change for enabling this needs to land in our image first.

device.go
```diff
@@ -166,11 +177,64 @@ func newBlockDevice(devInfo DeviceInfo) *BlockDevice {
...
+
+	driveName, err := getVirtDriveName(index)
+	if err != nil {
+		return err
+	}
+
+	deviceLogger().WithField("device", device.DeviceInfo.HostPath).Info("Attaching block device")
+	if c.state.State == "" {
+		if err := h.addDevice(drive, blockDev); err != nil {
+			return err
```
devimc: I guess you should also call to ```decrementPodBlockIndex``` here

device.go
```diff
@@ -166,11 +177,64 @@ func newBlockDevice(devInfo DeviceInfo) *BlockDevice {
...
+		}
+		device.DeviceInfo.Hotplugged = false
+	} else {
+		if err := h.hotplugAddDevice(drive, blockDev); err != nil {
+			return err
```
devimc: and here?decrementPodBlockIndex

pkg/uuid/uuid.go
```diff
@@ -0,0 +1,140 @@
+//
+// Copyright (c) 2016 Intel Corporation
```
devimc: 2017 ?

pkg/uuid/uuid_test.go
```diff
@@ -0,0 +1,95 @@
+//
+// Copyright (c) 2016 Intel Corporation
```
devimc: 2017

devimc:
Hi @amshinde 
I'm testing your code and I'm unable to passthrough a loop device, with runc I can do it

device.go
```diff
@@ -166,11 +177,64 @@ func newBlockDevice(devInfo DeviceInfo) *BlockDevice {
 	}
 }
 
-func (device *BlockDevice) attach(h hypervisor) error {
+func makeBlockDevIDForHypervisor(deviceID string) string {
+	return fmt.Sprintf("drive-%s", deviceID)
```
devimc: qemu 2.9 + q35 ID's must be <=31
would you mind reduce the length of the ID?

amshinde: the size limit was added in 2.9?
I'll change this, thanks

devimc: @amshinde yep, ```node-name``` is truncated to 31 characters

markdryan: Yep, that's correct.  "node-name" is truncated at 31 characters but "id" for some reason is not.  This can lead to a subtle problem where the backend (which uses "node-name") and frontend (which uses "id") devices cannot be associated with each other if you specify a node-name of > 31 characters.

devimc:
@amshinde  thanks for fixes q35 compatibility issues
btw this patch works in qemu 2.9 + q35 and qemu 2.7 + pc

devimc:
LGTM

amshinde:
@sameo Can you take another look at this?

device.go
```diff
@@ -166,11 +177,85 @@ func newBlockDevice(devInfo DeviceInfo) *BlockDevice {
...
+}
+
+func (device *BlockDevice) attach(h hypervisor, c *Container) error {
```
sboeuf: In order not to forget to decrement in case of error, you should use `defer` here:
```go
func (device *BlockDevice) attach(h hypervisor, c *Container) (err error) {

        ...

        index, err := c.pod.getAndSetPodBlockIndex()
        if err != nil {
                return err
        }
        defer func() {
                if err != nil {
                         c.pod.decrementPodBlockIndex()
                }
        }()

        ...

}
```

amshinde: I am using the util function generateRandomBytes with parameter 8 to get the id/node-name passed to qemu, so the length of the encoded string will be twice that + len("drive-").
I am not using uuid anymore, which similarly read from the rand device, but generated a string of more than 32 characters.

sboeuf: Ok I understand that you're solving the problem here, but I would expect a check to ensure the final string won't be more than 31 characters. If someday someone is changing `fmt.Sprintf("drive-%s", deviceID)` to `fmt.Sprintf("super-long-drive-string-%s", deviceID)`, we are gonna hit the 31bytes issue. Actually, I think this should be enforced inside the Qemu implementation since this is Qemu specific. We need a function to truncate IDs so that either if you hotplug or hotunplug, Qemu will always provide a consistent string of less than 32 bytes.

Also, adding a comment about that should help anyone reading the code.

device.go
```diff
@@ -200,6 +200,7 @@ func (device *BlockDevice) attach(h hypervisor, c *Container) error {
 	// rootfs and the predicted block device name needs to be provided to the agent.
 	index, err := c.pod.getAndSetPodBlockIndex()
 	if err != nil {
+		c.pod.decrementPodBlockIndex()
```
sameo: You probably want to defer a function that will check for err, and do the decrement if `err != nil`

device.go
```diff
@@ -170,11 +174,66 @@ func newBlockDevice(devInfo DeviceInfo) *BlockDevice {
 	}
 }
 
+func makeBlockDevIDForHypervisor(deviceID string) string {
+	return fmt.Sprintf("drive-%s", deviceID)
```
sameo: @amshinde Please do the truncating as part of this routine.

device.go
amshinde: done

device.go
amshinde: Done

device.go
```diff
@@ -170,11 +174,71 @@ func newBlockDevice(devInfo DeviceInfo) *BlockDevice {
 	}
 }
 
+func makeBlockDevIDForHypervisor(deviceID string) string {
+	devID := fmt.Sprintf("drive-%s", deviceID)
+	if len(devID) > maxDevIDSize {
+		devID = string(devID[:maxDevIDSize])
```
sboeuf: `maxDevIDSize` is specific to qemu and that's the reason why I wanted this check to be implemented in qemu.go through a new interface function of the hypervisor (something like capabilities() or so). But I think we can do that in a follow up PR since I don't want to block this one forever.

sboeuf:
LGTM

sboeuf:
I am gonna merge this, no need to have Fedora 26 build passing. This is currently fixed by another PR https://github.com/containers/virtcontainers/pull/453

PR #423, opened 2017-10-14
Vendoring: Use revision locked vendoring model

mcastelino:
Switch the vendoring model from a mix of Branch and Semver tags
to one that uses locked Revision.

This will mean that each dependency has to be updated explicitly
by the user in the Gopkg.toml file. This is different from the
current model where a prune and dep ensure by a user will
pick up the current tip of the each branch.

This will mean each dependency will have to be explicitly
vendored and updated using revision number.

However longer term it means that we will no longer have vendoring
mismatches between common dependencies across virtcontainers
and the clearcontainer runtime, enabling easier dependency management.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,7 +2,6 @@
 
 
 [[projects]]
-  branch = "master"
   name = "github.com/01org/ciao"
   packages = ["qemu","ssntp/uuid"]
   revision = "44068003db570d97b20b7de0ed0686a9eee0cc0f"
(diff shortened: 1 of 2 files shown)


jodh-intel:
lgtm

sboeuf:
LGTM

mcastelino:
@grahamwhaley @jodh-intel 

I will create a separate workflow document. The proposal would be along the following lines

- Both virtcontainers and runtime will use revision based vendoring
- We will need to keep the runtime and virtcontainers in sync (if the runtime is on an older revision then vendoring in the runtime will fail), as it will not satisfy the needs of virtcontainers
 - This will also mean that PR's have to be raised simultaneously across both the repos

Open
- If there are other dependencies that also use dep, then we may have issues even with this scheme if we cannot satisfy their constrains (e.g. tip of a branch)

PR #428, opened 2017-10-18
Ignore sysfs missing uevent file error, ignore "/dev/vfio/vfio"

amshinde:
Ability to obtain device names from major-minor number of device
was added in kernel 2.6.27 version:
torvalds/linux@e105b8b

But for certain devices like /dev/cuse, /dev/fuse, this sysfs
interface may not exist. This may also depend on the kernel
version used.

We are primarily using this sysfs interface to obtain device name in
case the renamed device name was passed to the runtime.

To accomodate these devices, which are passed by docker by default
for priveleged/non-priveleged containers, we ignore the missing file
error for /sys/dev files and assume device name passed in the oci spec
file for such devices is not the renamed device name.

### DIFF
--- a/device.go
+++ b/device.go
@@ -259,6 +259,20 @@ func GetHostPath(devInfo DeviceInfo) (string, error) {
 	format := strconv.FormatInt(devInfo.Major, 10) + ":" + strconv.FormatInt(devInfo.Minor, 10)
 	sysDevPath := filepath.Join(sysDevPrefix, pathComp, format, "uevent")
 
+	if _, err := os.Stat(sysDevPath); err != nil {
+		// Some devices(eg. /dev/fuse, /dev/cuse) do not always implement sysfs interface under /sys/dev
+		// These devices are passed by default by docker.
+		//
+		// Simply return the path passed in the device configuration, this does mean that no device renames are
+		// supported for these devices.
+
+		if os.IsNotExist(err) {
+			return devInfo.ContainerPath, nil
+		}
+
+		return "", err
+	}
+
 	content, err := ini.Load(sysDevPath)
 	if err != nil {
 		return "", err
--- a/device.go
+++ b/device.go
@@ -205,6 +205,11 @@ func (device *GenericDevice) deviceType() string {
 
 // isVFIO checks if the device provided is a vfio group.
 func isVFIO(hostPath string) bool {
+	// Ignore /dev/vfio/vfio character device
+	if strings.HasPrefix(hostPath, filepath.Join(vfioPath, "vfio")) {
+		return false
+	}
+
 	if strings.HasPrefix(hostPath, vfioPath) && len(hostPath) > len(vfioPath) {
 		return true
 	}
--- a/device.go
+++ b/device.go
@@ -236,6 +236,7 @@ func createDevice(devInfo DeviceInfo) Device {
 	} else if isBlock(path) {
 		return newBlockDevice(devInfo)
 	} else {
+		deviceLogger().WithField("device", path).Info("Device has not been passed to the container")
 		return newGenericDevice(devInfo)
 	}
 }
--- a/device_test.go
+++ b/device_test.go
@@ -39,6 +39,8 @@ func TestIsVFIO(t *testing.T) {
 		{"/dev/vfio", false},
 		{"/dev/vf", false},
 		{"/dev", false},
+		{"/dev/vfio/vfio", false},
+		{"/dev/vfio/vfio/12", false},
 	}
 
 	for _, d := range data {


device.go
```diff
@@ -253,7 +258,12 @@ func GetHostPath(devInfo DeviceInfo) (string, error) {
 
 	content, err := ini.Load(sysDevPath)
 	if err != nil {
-		return "", err
+		// Some devices(eg. /dev/fuse, /dev/cuse) do not always implement sysfs interface under /sys/dev
+		// These devices are passed by default by docker.
+		//
```
sameo: Not erroring out on `ini.Load()` failure could hide some other issues. Can we simply return `devInfo.ContainerPath, nil` when `sysDevPath` does not exist?

sameo:
@amshinde Could you please copy the first commit message into that PR description?

device.go
```diff
@@ -253,7 +253,12 @@ func GetHostPath(devInfo DeviceInfo) (string, error) {
...
+		// Simply return the path passed in the oci.spec file, this does mean that no device renames are
```
sboeuf: I would prefer to avoid mentioning `oci.spec` since this is OCI specific. I think we can just say that we instead provide the path that has been passed through the device configuration.

device.go
sboeuf: +1

device.go
```diff
@@ -197,6 +197,11 @@ func (device *GenericDevice) deviceType() string {
 
 // isVFIO checks if the device provided is a vfio group.
 func isVFIO(hostPath string) bool {
+	// Ignore /dev/vfio/vfio character device
+	if strings.HasPrefix(hostPath, filepath.Join(vfioPath, "vfio")) {
```
sboeuf: This works but I thought that we were not trying to handle privileged containers with virtualized containers since the whole security is compromised with such containers.

sboeuf:
@amshinde please add a description to the PR.

device.go
amshinde: We can't truly run in privileged mode, but when starting a container with --privileged, Docker does pass in a bunch of devices that may be required to run as a privileged container. We need to handle them.

device.go
```diff
@@ -251,6 +251,15 @@ func GetHostPath(devInfo DeviceInfo) (string, error) {
 	format := strconv.FormatInt(devInfo.Major, 10) + ":" + strconv.FormatInt(devInfo.Minor, 10)
 	sysDevPath := filepath.Join(sysDevPrefix, pathComp, format, "uevent")
 
+	if _, err := os.Stat(sysDevPath); err != nil {
```
sboeuf: Well to be more accurate:
```go
if _, err := os.Stat(sysDevPath); err != nil {
        if os.IsNotExist(err) {
               return devInfo.ContainerPath, nil
        }

        return "", err
}
```

device.go
sboeuf: But don't you think this should maybe handled at the runtime level and not in virtcontainers ?

amshinde: What do you mean? We handle the devices in virtcontainers, we were wrongly classifying "/dev/vfio/vfio" as a device bound  to the vfio driver when in fact it is a special device created by the kernel module. This commit rectifies this bug.

sboeuf: No I understand this is working and what it does, but I was thinking that if we don't expect to see this device/path at all since this is not a valid case, this should not be passed by the runtime as a device. I was only suggesting that our runtime would not allow such a device path to be provided to virtcontainers, but I might be wrong here.

amshinde: The runtime just passes the configuration to virtcontainers, it is upto the library how to handle the device. I don't think the runtime should handle what devices need to be passed to the library.

device.go
amshinde: Sure, will fix this.

device.go
sboeuf: Fair enough

device.go
sboeuf: Thanks

sboeuf:
LGTM

device.go
grahamwhaley: my 2p worth - I agree with the idea that as we cannot fully support `--privileged` mode (from the runtime and/or in virtcontainers either), that I'd much rather the runtime gracefully fails with a clear message if invoked with/in `--privileged` mode. Otherwise we are just misleading the end user who has run with `--privileged`, sees their container come up, but may be tracking down 'why did this not work as expected' :-(
I think I'm agreeing with you both:
- I'm happy for the virtcontainers to drop paths it knows it cannot support (I wonder if we should issue a log warning or message along with that if we don't already)
- I'd really like the runtime to gracefully fail to run in `--privileged` mode - but, I don't know if the runtime can even tell from the json if it is in that mode?

I'd like to hear if @sameo or @mcastelino have any thoughts on this (the runtime gracefully failing). Let me go open a runtime ticket (and tie it to the runtime limitations page) - as I think this is a runtime thing, not really a virtcontainers thing.

grahamwhaley: For reference: https://github.com/clearcontainers/runtime/issues/746

mcastelino: @grahamwhaley @amshinde we should log failure to map specific devices we do not support. But we always ignore the standard set the comes for all containers. To log them would make the log less useful. But we should log anything outside the standard set that we do not suppport.

amshinde: @mcastelino there is no way of knowing programatically what devices are passed on the command line whereas those passed by docker itself, unless we maintain a a list of them of what we think are standard ones.

device.go
```diff
@@ -228,6 +228,7 @@ func createDevice(devInfo DeviceInfo) Device {
 	} else if isBlock(path) {
 		return newBlockDevice(devInfo)
 	} else {
+		virtLog.Infof("Device %s has not been passed to the container", path)
```
jodh-intel: Could you make that "structured", so something like:

```
deviceLogger().WithField("device", path).Info("Device not passed to container")
```

amshinde: done

amshinde:
@sboeuf Can you merge this ?

sboeuf:
All comments have been addressed. Let's merge this and we should continue to track clearcontainers/runtime#746 since we should also implement something in the runtime itself.

PR #429, opened 2017-10-19
lock: Improve locking strategy

### LINKED ISSUE #421: lock: Improve the global locking of virtcontainers
I am gonna try to list what I consider as valid/invalid cases of virtcontainers usage and the appropriate way to handle this through a different locking approach than the very global one we are using right now. Indeed, the global locking is good to easily protect and manage locks between containers, but this is clearly not the most optimized way to do it, and we might have some performance issues related to the bottle necks resulting from this very global approach.

Ok so let's try to list what I consider as valid/invalid usages of virtcontainers:

1. Two calls to `ListPod()` at the same time about 2 different pods/containers (we could imagine a 3rd call asking for the status of a specific pod at the same time).
In that case, we want to read the information, this means we only need a shared/read lock for both of those calls, and specifically we need to add this to the `StatusPod()` function which is called from `ListPod()`.
Now, let's say the subsequent call to `statusContainer()` end up finding a stopped shim, meaning it has to update the status of a container, we will need to switch to an exclusive lock for that. I don't think it's gonna be an issue to release the lock when switching from share to exclusive, and we cannot do differently since this operation is not atomic.
Last thought about this case, we could have the 3rd caller asking for the deletion of a pod (through an exclusive lock), after both `ListPod()` took their shared locks. This means we could have a list of 10 pods from the reading done through `ListPod()`, but really have 9 pods a few seconds later, because of the delete coming after. I don't think this would be an issue since that's the caller's responsibility not to do actions on this deleted pod after it knows it has been deleted, even if `ListPod()` reports something different. Also, I think we already have this potential case right now, and it does not sound like an issue.

2. Several calls to the same pod or container are not allowed since it does not make sense to allow someone to call into `start` or `delete` before `create` for example. And I assume that we have to wait for the return of any command before we can issue a new one, otherwise there is no way to ensure about the integrity of the sequence of actions. According to these explanations, I don't think we need to lock every function of the API based on a PodLock, but instead lock more finely, every time we try to access the storage. This would mean two things, having shared locks when reading the storage through `fetch()` functions, and using exclusive locks when writing through `store()` functions. We have to be careful with the `store()` functions since we would have to fetch again, then update, and finally we could write (all of these being included in the exclusive lock), otherwise we could rely on old data at the moment we would try to store.

3. This 3rd case is more a question than a proposal since I don't know if we should consider this as valid or not. Let's say we have two calls to CreatePod() at the same time for 2 pods with the same ID, we should serialize the creation of pods until we actually store the pod, otherwise we could end up with two identical pods ID, with unknown results there... This is a question since I don't know if we consider this a valid case. I guess it can be valid if we have different orchestrations running on the same host and sharing the same /var/lib/virtcontainers/pods directory.

@grahamwhaley @sameo @jodh-intel let me know what you think about the cases and if you agree about the validity/invalidity of a usage, let me know what you think about the way I plan to solve it !

sboeuf:
This PR introduces shared locks so that it can improve the locking strategy. Basically, it takes a shared lock when a function only needs to read from the system, and it takes an exclusive lock when a function needs to modify what's written on the system.

Fixes #421

### DIFF
--- a/api.go
+++ b/api.go
@@ -103,7 +103,7 @@ func DeletePod(podID string) (VCPod, error) {
 		return nil, errNeedPodID
 	}
 
-	lockFile, err := lockPod(podID)
+	lockFile, err := rwLockPod(podID)
 	if err != nil {
 		return nil, err
 	}
@@ -162,7 +162,7 @@ func StartPod(podID string) (VCPod, error) {
 		return nil, errNeedPodID
 	}
 
-	lockFile, err := lockPod(podID)
+	lockFile, err := rwLockPod(podID)
 	if err != nil {
 		return nil, err
 	}
@@ -195,7 +195,7 @@ func StopPod(podID string) (VCPod, error) {
 		return nil, errNeedPod
 	}
 
-	lockFile, err := lockPod(podID)
+	lockFile, err := rwLockPod(podID)
 	if err != nil {
 		return nil, err
 	}
@@ -232,7 +232,7 @@ func RunPod(podConfig PodConfig) (VCPod, error) {
 		return nil, err
 	}
 
-	lockFile, err := lockPod(p.id)
+	lockFile, err := rwLockPod(p.id)
 	if err != nil {
 		return nil, err
 	}
@@ -331,7 +331,7 @@ func StatusPod(podID string) (PodStatus, error) {
 		return PodStatus{}, errNeedPodID
 	}
 
-	lockFile, err := lockPod(podID)
+	lockFile, err := rwLockPod(podID)
 	if err != nil {
 		return PodStatus{}, err
 	}
@@ -372,7 +372,7 @@ func CreateContainer(podID string, containerConfig ContainerConfig) (VCPod, VCCo
 		return nil, nil, errNeedPodID
 	}
 
-	lockFile, err := lockPod(podID)
+	lockFile, err := rwLockPod(podID)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -417,7 +417,7 @@ func DeleteContainer(podID, containerID string) (VCContainer, error) {
 		return nil, errNeedContainerID
 	}
 
-	lockFile, err := lockPod(podID)
+	lockFile, err := rwLockPod(podID)
 	if err != nil {
 		return nil, err
 	}
@@ -466,7 +466,7 @@ func StartContainer(podID, containerID string) (VCContainer, error) {
 		return nil, errNeedContainerID
 	}
 
-	lockFile, err := lockPod(podID)
+	lockFile, err := rwLockPod(podID)
 	if err != nil {
 		return nil, err
 	}
@@ -504,7 +504,7 @@ func StopContainer(podID, containerID string) (VCContainer, error) {
 		return nil, errNeedContainerID
 	}
 
-	lockFile, err := lockPod(podID)
+	lockFile, err := rwLockPod(podID)
 	if err != nil {
 		return nil, err
 	}
@@ -542,7 +542,7 @@ func EnterContainer(podID, containerID string, cmd Cmd) (VCPod, VCContainer, *Pr
 		return nil, nil, nil, errNeedContainerID
 	}
 
-	lockFile, err := lockPod(podID)
+	lockFile, err := rwLockPod(podID)
 	if err != nil {
 		return nil, nil, nil, err
 	}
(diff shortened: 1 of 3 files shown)


pod.go
```diff
@@ -362,8 +362,12 @@ func lockPod(podID string) (*os.File, error) {
 		return nil, err
 	}
 
-	err = syscall.Flock(int(lockFile.Fd()), syscall.LOCK_EX)
-	if err != nil {
+	lockType := syscall.LOCK_SH
```
grahamwhaley: I'd probably code this as an if/else - seeing something unconditionally set and then immediately conditionally set always makes me scrutinise the code as it looks there is something 'clever' going on to my eyes :-)

api.go
```diff
@@ -97,7 +97,7 @@ func DeletePod(podID string) (VCPod, error) {
 		return nil, errNeedPodID
 	}
 
-	lockFile, err := lockPod(podID)
+	lockFile, err := lockPod(podID, true)
```
grahamwhaley: just a nit: It would be nice if the exclusive arg gave some visual clue to if it was an exclusive or not request, something like:
`lockPod(podID, LOCK_EXCL)`
or similar - just to make it clearer when reading the code what the bool arg is saying. But, I don't think there is any easy existing arg type we can purloin for that. If it's easy to add a type just for this, maybe consider it.

sboeuf: Yes I can introduce `exclusivePodLock` and `sharedPodLock`.

pod.go
sboeuf: done

sboeuf:
@grahamwhaley I have addressed your comments, let me know how it looks now please.
@sameo could you take a look ?

sboeuf:
@sameo could you please take a look ?

pod.go
```diff
@@ -347,8 +347,16 @@ func (podConfig *PodConfig) valid() bool {
...
+	// R/W lock
+	exclusiveLock = syscall.LOCK_EX
+
+	// Read only lock
+	sharedLock = syscall.LOCK_SH
+)
+
 // lock locks any pod to prevent it from being accessed by other processes.
-func lockPod(podID string) (*os.File, error) {
+func lockPod(podID string, lockType int) (*os.File, error) {
```
sameo: I'd prefer to see wrappers defined around a `__lockPod(podID string lockType)`. Something like:

```Go
func rLockPod(podID string) (*os.File, error) {
     return __lockPod(podID, sharedLock)
}

func rwLockPod(podID string) (*os.File, error) {
     return __lockPod(podID, exclusiveLock)
}
```

api.go
```diff
@@ -579,7 +588,15 @@ func StatusContainer(podID, containerID string) (ContainerStatus, error) {
 		return ContainerStatus{}, errNeedContainerID
 	}
 
-	lockFile, err := lockPod(podID)
+	var wg sync.WaitGroup
+	// Deferring the synchronization here is very important since we want
+	// to avoid a deadlock. Indeed, the goroutine started by statusContainer
```
sameo: So the `statusContainer` go routine will have to wait for all read locks to be released before running. That means it will run asynchronously, but it does not mean there will be any deadlock. Am I missing something ?

Running asynchronously is fine, as long as the container status is properly set at some point.

pod.go
sboeuf: Done

api.go
sboeuf: There are two things here. First, we have to wait for the end of the goroutine, otherwise the runtime could exit and kill the goroutine when exiting, meaning the go routine could be interrupted while updating the container status.
Then, because we have to wait, we use here a WaitGroup and the call to `wg.Wait()` has to happen after the lock has been released from this function. Because the unlock is called also through a defer function, we have to make sure the `defer wg.Wait()` is earlier in the code so that it is called as the last thing.

api.go
```diff
@@ -331,7 +332,15 @@ func StatusPod(podID string) (PodStatus, error) {
...
+	// will need to lock an exclusive lock, meaning that all other locks have
+	// to be released to let this happen. This call ensures this will be the
+	// last operation executed by this function.
+	defer wg.Wait()
```
sameo: Can we please make the `wg` part of the Pod structure, so that we don't need to change the `statusContainer` prototype? And then have `pod.wait()`, `pod.add()` and `pod.done()` wrappers around it.

Also, the comment here should describe why we need to synchronize this: We need to **potentially** wait for a separate `container.stop()` routine that needs to be terminated before we return from the `StatusPod` call.

sboeuf: Well we cannot do that since we don't have the pod before to call into `defer wg.Wait()`.

I will add your comment to the code :)

sameo:
LGTM

PR #432, opened 2017-10-20
tests: Add tests to check the changes made to ignore uevent error

amshinde:
If the uevent file in /sys/dev/ is not found, we added a change to
ignore this error. Add test to check this behaviour.

Also, test should have been changed from a helper function.
Fix that as well.

### DIFF
--- a/device_test.go
+++ b/device_test.go
@@ -26,6 +26,8 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+const fileMode0640 = os.FileMode(0640)
+
 func TestIsVFIO(t *testing.T) {
 	type testData struct {
 		path     string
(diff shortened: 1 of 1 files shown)


device_test.go
```diff
@@ -26,6 +26,8 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+const fileMode = os.FileMode(0640)
```
sboeuf: Maybe you could call this `fileMode0640` because `fileMode` could be anything and it does not help when reading the code.

device_test.go
```diff
@@ -128,6 +122,33 @@ func testNewDevices(t *testing.T) {
 	}
 
 	devices, err := newDevices([]DeviceInfo{deviceInfo})
+	assert.NotNil(t, err)
+
+	format := strconv.FormatInt(major, 10) + ":" + strconv.FormatInt(minor, 10)
+	ueventPathPrefix := filepath.Join(sysDevPrefix, "char", format)
+	ueventPath := filepath.Join(ueventPathPrefix, "uevent")
+
+	// Return true for non-existant /sys/dev path.
```
sboeuf: You got a golint error with mispell `non-existant` instead of `non-existent`

amshinde: yes, fixing it

sboeuf:
LGTM

PR #433, opened 2017-10-22
pod: Do not overwrite pod state

sameo:
When setting a pod state, we don't want to overwrite its existing values
by storing an empty pod state with all block and fs states unitialized.

### DIFF
--- a/pod.go
+++ b/pod.go
@@ -491,18 +491,14 @@ func (p *Pod) GetContainer(containerID string) VCContainer {
 }
 
 func (p *Pod) createSetStates() error {
-	podState := State{
-		State: StateReady,
-		// retain existing URL value
-		URL: p.state.URL,
-	}
+	p.state.State = StateReady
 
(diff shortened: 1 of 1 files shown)


sameo:
@amshinde @mcastelino @sboeuf Without this patch, all our block indexes with CRI-O and k8s are off by one.

pod.go
```diff
@@ -491,18 +491,15 @@ func (p *Pod) GetContainer(containerID string) VCContainer {
 }
 
 func (p *Pod) createSetStates() error {
-	podState := State{
-		State: StateReady,
-		// retain existing URL value
-		URL: p.state.URL,
-	}
+	p.state.State = StateReady
+	p.state.URL = p.state.URL
```
amshinde: We are assigning the same variable to itself here.

sameo:
@amshinde Feel free to merge if all CIs pass.

PR #436, opened 2017-10-22
pod: Hotplug all pod container rootfs with devicemapper

### LINKED ISSUE #435: Stopping and removing CRI-O pod fails with devicemapper
When stopping a CRI-O pod, the devicemapper storage backend tries to unmount the pod container rootfs and that fails as the QEMU instance still holds the dm device (we cold plug it).

sameo:
When mixing cold and hot plugged devicemapper drives, orchestration
engines can get confused and removing pods can fail as the assumption
that a stopped pod pause container rootfs can be unmounted fails with
virtual machines.
    
Fixes #435

### DIFF
--- a/api.go
+++ b/api.go
@@ -82,6 +82,12 @@ func CreatePod(podConfig PodConfig) (VCPod, error) {
 		return nil, err
 	}
 
+	// Hotplug all pod containers rootfs if the hypervisor supports it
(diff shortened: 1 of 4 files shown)


sameo:
cc @amshinde @mcastelino @sboeuf PTAL

PR #438, opened 2017-10-23
qemu: Initialize qmp logger properly

sameo:
Initialize and use the qmpLogger logger field instead
of using a global logger.


sameo:
@jodh-intel This fixes the stderr errors with the qemu package.

PR #440, opened 2017-10-23
pause: Remove all things related to pause container

sboeuf:
The pause container was introduced long time ago because our agent
was not able to maintain a pod running when the last container had
exited. But now that our agent has this capability, we don't need
to maintain this workaround. Moreover, this will simplify our code
base.


jodh-intel:
\o/

lgtm

jodh-intel:
@sboeuf - could you handle the removal in the runtime too when you revendor?

jodh-intel:
... aaaand I've just come across https://github.com/clearcontainers/runtime/pull/756 :smile:

devimc:
magically cc-agent now is able to maintain a pod running? how?
commit message is not very clear, would you mind explain a little more how was that possible?

sboeuf:
@devimc well basically our agent was always able to run without this, that's a legacy thing we had introduced when our former hyperstart (the one from Hyper) was not able to maintain a running pod with no containers running.

devimc:
@sboeuf  make sense, thanks

sboeuf:
@jodh-intel @sameo do you mind merging this PR even without the CI passing here. I mean this is not worst than before, when we're going to vendor this from runtime, we will see if something got wrong, and we won't break the runtime. But of course, I have submitted https://github.com/clearcontainers/runtime/pull/756 and the CI is passing, so I think this is covered and prove nothing is gonna be broken.
I am asking to avoid blocking our PRs while we find a proper solution for the case where virtcontainers API or exported structures change.

sameo:
LGTM

PR #450, opened 2017-10-27
qemu: Fix incorrectly formatted log message

### LINKED ISSUE #449: Fix incorrectly formatted log message
The `inside-bool` message is incorrectly formatted (it should be logged as a string value):

```
pid=2819 name="cc-runtime" level="debug" inside-vm=%!q(bool=true) source="virtcontainers" subsystem="qemu" msg="Checking nesting environment"
```

jodh-intel:
The message that shows if running inside a VM was being displayed as a
raw golang variable ("%!q(bool=true)") rather than a "true"/"false"
string.

Fixes #449.

### DIFF
--- a/qemu.go
+++ b/qemu.go
@@ -488,7 +488,7 @@ func (q *qemu) init(config HypervisorConfig) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


sameo:
LGTM

PR #456, opened 2017-10-30
 hypervisor: add hot plugging support for Qemu Q35

devimc:
Qemu Q35 machine type does not support to hot plug devices
directly on pcie.0, hence we have to find a way to allow users
hot plug N devices.
The only way to hot plug devices in Qemu Q35 is through PCI brdiges.
Each PCI bridge is able to support until 32 devices, therefore we have a
limitation in the amount of devices we can hot plug.
This patch adds support for hot plugging devices on PCI bridges in
pc and q35 machine types.

### DIFF
--- a/bridge.go
+++ b/bridge.go
@@ -1,0 +1,104 @@
+//
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package virtcontainers
+
+import "fmt"
+
+type bridgeType string
+
+const (
+	pciBridge  bridgeType = "pci"
+	pcieBridge            = "pcie"
+)
+
+const pciBridgeMaxCapacity = 30
+
+// Bridge is a bridge where devices can be hot plugged
+type Bridge struct {
+	// Address contains information about devices plugged and its address in the bridge
+	Address map[uint32]string
+
+	// Type is the type of the bridge (pci, pcie, etc)
+	Type bridgeType
+
+	//ID is used to identify the bridge in the hypervisor
+	ID string
+}
+
+// NewBridges creates n new pci(e) bridges depending of the machine type
+func NewBridges(n uint32, machine string) []Bridge {
+	var bridges []Bridge
+	var bt bridgeType
+
+	switch machine {
+	case QemuQ35:
+		// currently only pci bridges are supported
+		// qemu-2.10 will introduce pcie bridges
+		fallthrough
+	case QemuPC:
+		bt = pciBridge
+	default:
+		return nil
+	}
+
+	for i := uint32(0); i < n; i++ {
+		bridges = append(bridges, Bridge{
+			Type:    bt,
+			ID:      fmt.Sprintf("%s-bridge-%d", bt, i),
+			Address: make(map[uint32]string),
+		})
+	}
+
+	return bridges
+}
+
+// addDevice on success adds the device ID to the bridge and return the address
+// where the device was added, otherwise an error is returned
+func (b *Bridge) addDevice(ID string) (uint32, error) {
+	var addr uint32
+
+	// looking for the first available address
+	for i := uint32(1); i <= pciBridgeMaxCapacity; i++ {
+		if _, ok := b.Address[i]; !ok {
+			addr = i
+			break
+		}
+	}
+
+	if addr == 0 {
+		return 0, fmt.Errorf("Unable to hot plug device on bridge: there are not empty slots")
+	}
+
+	// save address and device
+	b.Address[addr] = ID
+	return addr, nil
+}
+
+// removeDevice on success removes the device ID from the bridge and return nil,
+// otherwise an error is returned
+func (b *Bridge) removeDevice(ID string) error {
+	// check if the device was hot plugged in the bridge
+	for addr, devID := range b.Address {
+		if devID == ID {
+			// free address to re-use the same slot with other devices
+			delete(b.Address, addr)
+			return nil
+		}
+	}
+
+	return fmt.Errorf("Unable to hot unplug device %s: not present on bridge", ID)
+}
--- a/filesystem.go
+++ b/filesystem.go
@@ -42,6 +42,9 @@ const (
 	// networkFileType represents a network file type (pod only)
 	networkFileType
 
+	// hypervisorFileType represents a hypervisor file type (pod only)
+	hypervisorFileType
+
 	// processFileType represents a process file type
 	processFileType
 
@@ -63,6 +66,9 @@ const stateFile = "state.json"
 
 // networkFile is the file name storing a pod network.
 const networkFile = "network.json"
+
+// hypervisorFile is the file name storing a hypervisor's state.
+const hypervisorFile = "hypervisor.json"
 
 // processFile is the file name storing a container process.
 const processFile = "process.json"
@@ -108,6 +114,10 @@ type resourceStorage interface {
 	fetchPodState(podID string) (State, error)
 	fetchPodNetwork(podID string) (NetworkNamespace, error)
 	storePodNetwork(podID string, networkNS NetworkNamespace) error
+
+	// Hypervisor resources
+	fetchHypervisorState(podID string, state interface{}) error
+	storeHypervisorState(podID string, state interface{}) error
 
 	// Container resources
 	storeContainerResource(podID, containerID string, resource podResource, data interface{}) error
@@ -319,7 +329,7 @@ func (fs *filesystem) fetchDeviceFile(fileData []byte, devices *[]Device) error 
 func resourceNeedsContainerID(podSpecific bool, resource podResource) bool {
 
 	switch resource {
-	case lockFileType, networkFileType:
+	case lockFileType, networkFileType, hypervisorFileType:
 		// pod-specific resources
 		return false
 	default:
@@ -342,7 +352,7 @@ func resourceDir(podSpecific bool, podID, containerID string, resource podResour
 	case configFileType:
 		path = configStoragePath
 		break
-	case stateFileType, networkFileType, processFileType, lockFileType, mountsFileType, devicesFileType:
+	case stateFileType, networkFileType, processFileType, lockFileType, mountsFileType, devicesFileType, hypervisorFileType:
 		path = runStoragePath
 		break
 	default:
@@ -378,6 +388,8 @@ func (fs *filesystem) resourceURI(podSpecific bool, podID, containerID string, r
 		filename = stateFile
 	case networkFileType:
 		filename = networkFile
+	case hypervisorFileType:
+		filename = hypervisorFile
 	case processFileType:
 		filename = processFile
 	case lockFileType:
@@ -429,6 +441,7 @@ func (fs *filesystem) commonResourceChecks(podSpecific bool, podID, containerID 
 	case configFileType:
 	case stateFileType:
 	case networkFileType:
+	case hypervisorFileType:
 	case processFileType:
 	case mountsFileType:
 	case devicesFileType:
@@ -598,8 +611,21 @@ func (fs *filesystem) fetchPodNetwork(podID string) (NetworkNamespace, error) {
 	return networkNS, nil
 }
 
+func (fs *filesystem) fetchHypervisorState(podID string, state interface{}) error {
+	return fs.fetchResource(true, podID, "", hypervisorFileType, state)
+}
+
 func (fs *filesystem) storePodNetwork(podID string, networkNS NetworkNamespace) error {
 	return fs.storePodResource(podID, networkFileType, networkNS)
+}
+
+func (fs *filesystem) storeHypervisorState(podID string, state interface{}) error {
+	hypervisorFile, _, err := fs.resourceURI(true, podID, "", hypervisorFileType)
+	if err != nil {
+		return err
+	}
+
+	return fs.storeFile(hypervisorFile, state)
 }
 
 func (fs *filesystem) deletePodResources(podID string, resources []podResource) error {
--- a/pod.go
+++ b/pod.go
@@ -578,6 +578,10 @@ func createPod(podConfig PodConfig) (*Pod, error) {
 		return nil, err
 	}
 
+	if err := p.storage.storeHypervisorState(p.id, p.hypervisor.getState()); err != nil {
+		return nil, err
+	}
+
 	return p, nil
 }
 
@@ -590,10 +594,6 @@ func doFetchPod(podConfig PodConfig) (*Pod, error) {
 
 	hypervisor, err := newHypervisor(podConfig.HypervisorType)
 	if err != nil {
-		return nil, err
-	}
-
-	if err := hypervisor.init(podConfig.HypervisorConfig); err != nil {
 		return nil, err
 	}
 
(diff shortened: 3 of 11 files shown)


devimc:
@mcastelino @amshinde can you please review it ?

bridge.go
```diff
@@ -0,0 +1,122 @@
...
+		return fmt.Errorf("Unable to hot plug device on bridge: there are not empty slots")
+	}
+
+	// hotplug drive on this bridge
+	drive.Bus = b.ID
+
+	// hotplug drive on this bridge's addres
+	drive.Addr = fmt.Sprintf("0x%x", addr)
+
+	err := h.hotplugAddDevice(drive, devType)
```
sboeuf: I am not sure this is the right way to implement it inside virtcontainers. This should be called from the hypervisor implementation since different hypervisors might handle this in a different way.

bridge.go
```diff
@@ -0,0 +1,122 @@
...
+		if devID == drive.ID {
+			devAddr = addr
+		}
+	}
+
+	if devAddr == 0 {
+		return fmt.Errorf("Unable to hot unplug device %s: not present on bridge", drive.ID)
+	}
+
+	err := h.hotplugRemoveDevice(drive, devType)
```
sboeuf: Ditto

bridge.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	if addr == 0 {
+		return fmt.Errorf("Unable to hot plug device on bridge: there are not empty slots")
+	}
+
+	// hotplug drive on this bridge
+	drive.Bus = b.ID
+
+	// hotplug drive on this bridge's addres
```
jodh-intel: typo: "address".

bridge.go
```diff
@@ -0,0 +1,122 @@
...
+			Type:    bt,
+			ID:      fmt.Sprintf("%s-bridge-%d", bt, i),
+			Address: make(map[uint32]string),
+		})
+	}
+
+	return bridges
+}
+
+func (b *Bridge) hotplugAddPCIDevice(h hypervisor, drive Drive, devType deviceType) error {
```
jodh-intel: Am I missing something? This function and `hotplugRemovePCIDevice()` don't seem to be used?

pod.go
```diff
@@ -283,6 +283,10 @@ type Resources struct {
 
 	// Memory is the amount of available memory in MiB.
 	Memory uint
+
+	// Bridges is the list of pci bridges attached to the pod
+	// pci bridges can be used to hot plug devices
+	Bridges []*Bridge
```
sboeuf: This should be an hypervisor resource IMO.

bridge.go
```diff
@@ -0,0 +1,122 @@
...
+import "fmt"
+
+type bridgeType string
+
+const (
+	pciBridge  bridgeType = "pci"
+	pcieBridge            = "pcie"
+)
+
+const pciBridgeMaxCapacity = 30
```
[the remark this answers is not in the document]
devimc: > Where does this limit come from?

Qemu's documentation is not very good, I found this [1], but I'm not able to hot plug 32 devices,
I guess because of SHPC is enabled. 

> Is there any way to avoid hard-coding this?

I don't think so

[1] https://github.com/qemu/qemu/blob/master/docs/pcie.txt#L76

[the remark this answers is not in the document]
grahamwhaley: Having a peek at: https://en.wikipedia.org/wiki/PCI_configuration_space
looks like the per-bus device space is defined by a 5-bit field.
Maybe 0 and 0x1F are reserved - leaving 30 ? Just a guess. We could add a note saying this is part of the PCI spec I guess etc.

bridge.go
devimc: ouch!, sorry I forgot to push some patches

pod.go
devimc: I think a bridge is also a resource, same as memory and CPU

devimc:
@sboeuf  @jodh-intel  changes applied

bridge.go
devimc: why Ditto [1]? 

[1] https://bulbapedia.bulbagarden.net/wiki/Ditto_(Pok%C3%A9mon)   😄

bridge.go
```diff
@@ -0,0 +1,122 @@
+//
```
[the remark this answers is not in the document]
devimc: fixed, thanks

bridge.go
devimc: PR updated, thanks

bridge.go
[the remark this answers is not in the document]
sboeuf: The title of the commit message is still wrong.

device.go
```diff
@@ -232,7 +232,16 @@ func (device *BlockDevice) attach(h hypervisor, c *Container) (err error) {
 
 		device.DeviceInfo.Hotplugged = false
 	} else {
-		if err = h.hotplugAddDevice(drive, blockDev); err != nil {
+		var err error
+		for _, b := range c.pod.config.VMConfig.Bridges {
+			err = b.hotplugAddPCIDevice(h, drive, blockDev)
```
sboeuf: That's exactly what I am pointing in a previous commit as not the right way to proceed. hotplugAddPCIDevice() should be called from the hypervisor implementation (no need for `h` as parameter in that case), and we should not modify the call here. Hypervisor implementation should be able to detect if it needs to hotplug a PCI device or not I think.

bridge.go
sboeuf: @devimc you have not taken this comment into account.

bridge.go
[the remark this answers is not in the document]
devimc: ahh ok s/Implements/Implement/ 
done, thanks

device.go
devimc: we need to save PCI bridges in the pod state because before trying to hot plug any device we must know that slots are available, is the hypervisor structure saved in the pod state?

sboeuf: @devimc currently, we don't save any data related to hypervisor in a state file (relates to run-time which is different from config files storing the Hypervisor config). But I think that if you need to store some informations so that you can reuse them across different containers, you should maybe create a new state file for the hypervisor.
Before you do that, I'd like to make sure I understand properly what you need.

bridge.go
```diff
@@ -0,0 +1,122 @@
...
+			break
+		}
+	}
+
+	if addr == 0 {
+		return fmt.Errorf("Unable to hot plug device on bridge: there are not empty slots")
+	}
+
+	// hotplug drive on this bridge
+	drive.Bus = b.ID
```
amshinde: You are updating drive here. Wouldn't you require to pass pointer to the drive for the update?

bridge.go
```diff
@@ -0,0 +1,122 @@
...
+		return fmt.Errorf("Unable to hot unplug device %s: not present on bridge", drive.ID)
+	}
+
+	err := h.hotplugRemoveDevice(drive, devType)
+	if err != nil {
+		return err
+	}
+
+	// free address to re-use the same slot with other devices
+	delete(b.Address, devAddr)
```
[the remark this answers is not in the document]
mcastelino: @amshinde that is a good idea. This will ensure we are fully OS independent.

device.go
mcastelino: @sboeuf do we store the POD state somewhere? also where do we store details about the devices attached to the PODs and containers.

devimc:
@mcastelino  @amshinde  @sboeuf  I did some changes in order to handle Sebastien's comments,
now:
* hypervisor is who plug and unplug devices
* hypervisor updates the bridges and its addresses
* All bridges are part of the hypervisor state
* hypervisor state is part of pod state

could you please re-review it?  thanks

sboeuf:
@devimc thanks for the update, I am gonna take a look

qemu.go
```diff
@@ -358,6 +358,33 @@ func (q *qemu) appendFSDevices(devices []ciaoQemu.Device, podConfig PodConfig) [
...
+			t = ciaoQemu.PCIEBridge
+		}
+
+		devices = append(devices,
+			ciaoQemu.BridgeDevice{
+				Type: t,
+				Bus:  bus,
+				ID:   b.ID,
+				// Each bridge is required to be assigned a unique chassis id > 0
+				Chassis: (idx + 1),
```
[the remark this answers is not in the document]
devimc: I think it is not needed, we already have an ID to identify the bridge

pod.go
```diff
@@ -72,6 +72,9 @@ type State struct {
 
 	// Bool to indicate if the drive for a container was hotplugged.
 	HotpluggedDrive bool `json:"hotpluggedDrive"`
+
+	// HypervisorState keeps the state of the hypervisor
+	HypervisorState HypervisorState `json:"hypervisorState"`
```
sboeuf: This is not the way we have implemented the state files on virtcontainers. Let me think about it, I'll try to come up with a clean solution for this storage purpose (now that I am clear about what we expect).
@amshinde please take a look and give some inputs about this too.

sboeuf:
@devimc I have discussed with @amshinde about the options we have to implement the storage of this list of bridges. I know what we want to store is at a Pod level (because it depends on the VM), but I don't want to store this through the pod state since this is very specific to the hypervisor implementation. Indeed, let's take the exemple of Xen hypervisor, and let's say we want to hotplug a new device. In case Xen does not need a bridge because the hotplug works as expected, we don't want to update the Pod state for nothing. What we need is to let every hypervisor implementation decide if they want to store something into `hypervisor.json` and what they want to store.

In our case this means several changes, let me try to summarize this:

__1. Update Qemu structure__
```go
type qemu struct {
        pod *Pod
        ...
        bridges []Bridges
}
```
__2. Update hypervisor interface__
Change this:
```go
type hypervisor interface {
        init(config HypervisorConfig) error
        ...
}
```
to that:
```go
type hypervisor interface {
        init(pod *Pod) error
        ...
}
```
The explanation is that we need the pod to be stored inside qemu structure so that we can access storage anytime we want from the qemu implementation. Xen implementation would do nothing with this pod actually since it would not need to store anything.
This pod handler is also going to be used to retrieve the `HypervisorConfig` through `pod.config.HypervisorConfig`.

Of course, you're gonna have to update the mock implementation, but that's nothing.

__3. Modify doFetchPod() sequence__
Change this:
```go
func doFetchPod(podConfig PodConfig) (*Pod, error) {
	...
	if err := hypervisor.init(podConfig.HypervisorConfig); err != nil {
		return nil, err
	}
	...
	p := &Pod{
		id:              podConfig.ID,
		hypervisor:      hypervisor,
		...
	}
        ...
```
to that:
```go
func doFetchPod(podConfig PodConfig) (*Pod, error) {
	...
	p := &Pod{
		id:              podConfig.ID,
		hypervisor:      hypervisor,
		...
	}

        if err := hypervisor.init(pod); err != nil {
		return nil, err
	}
        ...
```
We need this, otherwise we cannot provide the pod to the `hypervisor.init()`.

__4. Create a new hypervisor.json file__
For this one, follow what we have inside `filesystem.go` and duplicate what is needed for creating a `hypervisorFileType`.

__5. Fetch hypervisor info__
```go
func (q *qemu) init(pod *Pod) error {
        q.pod = pod
        ...
        // Fetch the hypervisor.json if it exists and assign q.bridges with what we read.
}
```
Every time the hypervisor is gonna be created from `doFetchPod()`, the call to hypervisor.init() will fetch the info from the storage in the specific case of Qemu. Also, we save the pod handler to be able to reuse it later for storage purpose.

__6. Store hypervisor info__
Both `createPod()`, `hotplugAddDevice()` and `hotplugRemoveDevice()` are gonna be responsible for updating the storage since they will modify the list of bridges.

@devimc I am trying to think of everything here, but I might have forgotten some details. @amshinde, @sameo please let me know what you think about this proposal and let me know if something is missing.

bridge.go
```diff
@@ -0,0 +1,104 @@
...
+	pciBridge  bridgeType = "pci"
+	pcieBridge            = "pcie"
+)
+
+const pciBridgeMaxCapacity = 30
+
+// Bridge is a bridge where devices can be hot plugged
+type Bridge struct {
+	// Address contains information about devices plugged and its address in the bridge
+	Address map[uint32]string
```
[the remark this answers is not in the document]
devimc: pcie has more capacity, for this reason I'd like to keep ```uint32```

devimc:
@sboeuf your proposal is good and feasible

amshinde:
@sboeuf you have captured almost everything that we discussed. Another point to note, we should consider storing the hypervisor.json at the end of a container lifecycle event vs having it done immediately after a device is added.

sboeuf:
@amshinde oh yes you're right, and this should come as a second step (as an improvement).

sboeuf:
@devimc I am glad that's clear :)

devimc:
@sboeuf  just one question, if hypervisor.json is hypervisor specific then filesystem fetch method should return ```[]byte``` and store should receive ```interface{}```, right?

sboeuf:
@devimc oh yeah let me clarify this. So basically I think you'll have the fetch method returning an `interface{}` and the store method receiving an `interface{}`. And from qemu implementation, you will provide `QemuState` (make sure the struct and all its fields are public otherwise marshalling won't work) through the `interface{}` parameter (that's for the `store()` and it should be straightforward). For the `fetch()`, you will receive an `interface{}` that you're going to cast into the expected structure from every hypervisor implementation that needs to retrieve `hypervisor.json`.
Is that clearer ? Don't hesitate if you still have questions ;)

devimc:
@sboeuf  changes applied

sboeuf:
@devimc thanks, I am going to review this !

qemu.go
```diff
@@ -862,11 +947,27 @@ func (q *qemu) hotplugDevice(devInfo interface{}, devType deviceType, op operati
 }
 
 func (q *qemu) hotplugAddDevice(devInfo interface{}, devType deviceType) error {
-	return q.hotplugDevice(devInfo, devType, addDevice)
+	if err := q.hotplugDevice(devInfo, devType, addDevice); err != nil {
+		return err
+	}
+
+	if err := q.pod.storage.storeHypervisorState(q.pod.id, q.state); err != nil {
```
sboeuf: This can be improved later (moving this to container.go where we actually add the devices), but that's fine for now.

qemu.go
```diff
@@ -862,11 +947,27 @@ func (q *qemu) hotplugDevice(devInfo interface{}, devType deviceType, op operati
...
+	return nil
 }
 
 func (q *qemu) hotplugRemoveDevice(devInfo interface{}, devType deviceType) error {
-	return q.hotplugDevice(devInfo, devType, removeDevice)
+	if err := q.hotplugDevice(devInfo, devType, removeDevice); err != nil {
+		return err
+	}
+
+	if err := q.pod.storage.storeHypervisorState(q.pod.id, q.state); err != nil {
```
sboeuf: Ditto.

filesystem.go
```diff
@@ -675,10 +728,31 @@ func (fs *filesystem) fetchPodNetwork(podID string) (NetworkNamespace, error) {
 	return NetworkNamespace{}, fmt.Errorf("Unknown network type")
 }
 
+func (fs *filesystem) fetchHypervisorState(podID string, state interface{}) error {
```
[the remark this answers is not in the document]
devimc: I'll cast data to []byte, if not, I get this error

```sh
# github.com/containers/virtcontainers
./filesystem.go:735:25: cannot use data (type interface {}) as type []byte in argument to json.Unmarshal: need type assertion
Makefile:27: recipe for target 'build' failed
make: *** [build] Error 2
```

```go
func (fs *filesystem) fetchHypervisorState(podID string, state interface{}) error {
        data, err := fs.fetchResource(true, podID, "", hypervisorFileType)
	if err != nil {
		return err
	}

        if err = json.Unmarshal(data.([]byte), state); err != nil {
                return err
        }

        return nil
}
```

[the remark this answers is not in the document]
sboeuf: @devimc you're right but there is an underlying problem here and it's not related to your PR, but to the way the storage (and the fetch specifically) is implemented in virtcontainers.
I have submitted https://github.com/containers/virtcontainers/pull/472 in order to make this more flexible and it should solve your issues. I am waiting for @sameo to review this PR and hopefully this will get merged soon.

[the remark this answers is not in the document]
sboeuf: #472 and #477 have been merged! You can rebase your PR on top of it, and simplify the storage part :)

filesystem.go
```diff
@@ -598,10 +610,23 @@ func (fs *filesystem) fetchPodNetwork(podID string) (NetworkNamespace, error) {
 	return networkNS, nil
 }
 
+func (fs *filesystem) fetchHypervisorState(podID string, state interface{}) error {
+	return fs.fetchResource(true, podID, "", hypervisorFileType, state)
```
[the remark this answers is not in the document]
devimc: change applied, thanks

sboeuf:
LGTM

PR #457, opened 2017-10-30
virtc: Fix virtc to make it properly usable

sboeuf:
With recent changes brought to the Clear Containers runtime, virtc
was not really in sync with what we were currently testing. This
patch replaces the systemd target file to be run by the guest kernel
and it also enhances the handling of the hypervisor machine type.

By default, we want to rely on "pc" machine type (with the proper
uncompressed kernel associated), and through the CLI, we can choose
"pc-lite" which will be automatically associated to the compressed
kernel.


sameo:
LGTM

jodh-intel:
lgtm

PR #460, opened 2017-10-30
Multi OS support

### LINKED ISSUE #439: Add multi OS support
## Problem statement
One key advantage of Clear Containers over `runc` containers is the ability for each container/pod to run on top of its own kernel and guest image.
For customized workloads like e.g. NFV ones, this is very important as they sometimes need some specific kernel features that can be mutually exclusive with other workload needs.

Unfortunately, the OCI runtime specifications do not allow (yet) for specifying a kernel and guest image and thus `virtcontainers` callers can only use a default path for each of those and pass it through the `HypervisorConfig` structure.

## Proposed solutions

We want to be able to define a kernel and a guest image paths per container/pod. Both of them should be **absolute** paths, and should be optional. In other words, there must be a default value for those paths which will be passed to virtcontainers through the `HypervisorConfig` structure. Clear Containers, for example, gets the default paths for its kernel and guest image through the `configuration.toml` system wide configuration file.

### Short term solution: Annotations
As a short term solution we are going to use virtcontainers namespaced pod annotations for passing a kernel and guest image absolute path:

```Go
// KernelPath is a pod annotation for passing a per container path pointing at the kernel needed to boot the container VM.
KernelPath = "com.github.containers.virtcontainers.KernelPath"

// ImagePath is an pod annotation for passing a per container path pointing at the guest image that will run the container VM.
ImagePath = "com.github.containers.virtcontainers.ImagePath"
```

Both annotations can be set through `virtcontainers` `PodConfig` annotations. Kernel or image path `ContainerConfig` annotations will be ignored.

### Long term solution: OCI specification changes

See https://github.com/opencontainers/runtime-spec/pull/405

### Code path

The virtcontainers code path should be identical for both solutions: The `pod.go` will trap pod creation request and modify its `config.HypervisorConfig` structure if there is an `ImagePath` and/or `KernelPath` annotation as part of its `PodConfig`

The `pkg/oci/utils.go` file will be responsible for adding the right pod annotation, either by parsing the OCI `config.json` annotations or by parsing the future VM specific section in that file. The latter should take precedence over the former.

### Error handling

If any of the kernel or guest image path is invalid (empty files, or inexistent ones), `virtcontainers` will error out and stop creating the pod's VM. Passing an image or/and a guest image into the container configuration file is a very specific requirement from the user and thus switching to the default value when one of those paths is invalid could be misleading and error prone.

### Security

Optionally `virtcontainers` callers can require kernel and/or image binaries integrity checking. For that purpose, `virtcontainers` will handle 2 additional annotations:

```Go
// KernelHash is a pod annotation for passing a container kernel image SHA-512 hash value.
KernelHash = "com.github.containers.virtcontainers.KernelHash"

// ImageHash is an pod annotation for passing a container guest image SHA-512 hash value.
ImageHash = "com.github.containers.virtcontainers.ImageHash"
```

By default `virtcontainers` will not verify the kernel and image binaries integrity unless one or both of the hash annotations are passed in the `PodConfig` annotations. In the latter case, `virtcontainers` will build SHA-512 hashes on the kernel or/and image binaries and compare that against the passed annotations.

sameo:
Based on the pod config annotations, we change the underlying
hypervisor configuration. Pod config annotations can carry a
kernel, image, hypervisor and firmware paths.
Those annotations, if valid, will supersede the hypervisor
configuration paths. If not valid, virtcontainers will error
out.

Optionally, asset SHA-512 hashes can be passed through pod
annotations as well. In that case virtcontainers will verify
that the asset binary matches the hash annotation and error
out when it does not.

More details on the architecture in #439 
    
Fixes #439

TODO:

- [X] Unit tests
- [X] Update `pkg/oci/utils.go` to convert an OCI annotations into a `PodConfig` one for assets

### DIFF
--- a/asset.go
+++ b/asset.go
@@ -1,0 +1,168 @@
+//
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package virtcontainers
+
+import (
+	"crypto/sha512"
+	"encoding/hex"
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+
+	"github.com/containers/virtcontainers/pkg/annotations"
+)
+
+type assetType string
+
+func (t assetType) annotations() (string, string, error) {
+	switch t {
+	case kernelAsset:
+		return annotations.KernelPath, annotations.KernelHash, nil
+	case imageAsset:
+		return annotations.ImagePath, annotations.ImageHash, nil
+	}
+
+	return "", "", fmt.Errorf("Wrong asset type %s", t)
+}
+
+const (
+	kernelAsset assetType = "kernel"
+	imageAsset  assetType = "image"
+)
+
+type asset struct {
+	path         string
+	computedHash string
+	kind         assetType
+}
+
+func (a *asset) valid() bool {
+	if !filepath.IsAbs(a.path) {
+		return false
+	}
+
+	switch a.kind {
+	case kernelAsset:
+		return true
+	case imageAsset:
+		return true
+	}
+
+	return false
+}
+
+// hash returns the hex encoded string for the asset hash
+func (a *asset) hash(hashType string) (string, error) {
+	var hashEncodedLen int
+	var hash string
+
+	// We read the actual asset content
+	bytes, err := ioutil.ReadFile(a.path)
+	if err != nil {
+		return "", err
+	}
+
+	if len(bytes) == 0 {
+		return "", fmt.Errorf("Empty asset file at %s", a.path)
+	}
+
+	// Build the asset hash and convert it to a string.
+	// We only support SHA512 for now.
+	switch hashType {
+	case annotations.SHA512:
+		virtLog.Debugf("Computing %v hash", a.path)
+		hashComputed := sha512.Sum512(bytes)
+		hashEncodedLen = hex.EncodedLen(len(hashComputed))
+		hashEncoded := make([]byte, hashEncodedLen)
+		hex.Encode(hashEncoded, hashComputed[:])
+		hash = string(hashEncoded[:])
+		virtLog.Debugf("%v hash: %s", a.path, hash)
+	default:
+		return "", fmt.Errorf("Invalid hash type %s", hashType)
+	}
+
+	a.computedHash = hash
+
+	return hash, nil
+}
+
+func (a *asset) override(podConfig *PodConfig) error {
+	if a == nil || a.path == "" {
+		virtLog.Debug("Using default asset as no custom value specified")
+		return nil
+	}
+
+	if !a.valid() {
+		return fmt.Errorf("Invalid %s at %s", a.kind, a.path)
+	}
+
+	switch a.kind {
+	case kernelAsset:
+		podConfig.HypervisorConfig.KernelPath = a.path
+	case imageAsset:
+		podConfig.HypervisorConfig.ImagePath = a.path
+	}
+
+	return nil
+}
+
+// newAsset returns a new asset from the pod annotations.
+func newAsset(podConfig *PodConfig, t assetType) (*asset, error) {
+	pathAnnotation, hashAnnotation, err := t.annotations()
+	if err != nil {
+		return nil, err
+	}
+
+	if pathAnnotation == "" || hashAnnotation == "" {
+		return nil, fmt.Errorf("Missing annotation paths for %s", t)
+	}
+
+	path, ok := podConfig.Annotations[pathAnnotation]
+	if !ok || path == "" {
+		return nil, nil
+	}
+
+	if !filepath.IsAbs(path) {
+		return nil, fmt.Errorf("%s is not an absolute path", path)
+	}
+
+	a := &asset{path: path, kind: t}
+
+	hash, ok := podConfig.Annotations[hashAnnotation]
+	if !ok || hash == "" {
+		return a, nil
+	}
+
+	// We have a hash annotation, we need to verify the asset against it.
+	hashType, ok := podConfig.Annotations[annotations.AssetHashType]
+	if !ok {
+		virtLog.Warningf("Unrecognized hash type: %s, switching to %s", hashType, annotations.SHA512)
+		hashType = annotations.SHA512
+	}
+
+	hashComputed, err := a.hash(hashType)
+	if err != nil {
+		return a, err
+	}
+
+	// If our computed asset hash does not match the passed annotation, we must exit.
+	if hashComputed != hash {
+		return nil, fmt.Errorf("Invalid hash for %s: computed %s, expecting %s]", a.path, hashComputed, hash)
+	}
+
+	return a, nil
+}
--- a/asset.go
+++ b/asset.go
@@ -34,14 +34,20 @@ func (t assetType) annotations() (string, string, error) {
 		return annotations.KernelPath, annotations.KernelHash, nil
 	case imageAsset:
 		return annotations.ImagePath, annotations.ImageHash, nil
+	case hypervisorAsset:
+		return annotations.HypervisorPath, annotations.HypervisorHash, nil
+	case firmwareAsset:
+		return annotations.FirmwarePath, annotations.FirmwareHash, nil
 	}
 
 	return "", "", fmt.Errorf("Wrong asset type %s", t)
 }
 
 const (
-	kernelAsset assetType = "kernel"
-	imageAsset  assetType = "image"
+	kernelAsset     assetType = "kernel"
+	imageAsset      assetType = "image"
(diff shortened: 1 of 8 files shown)


sameo:
cc @mcastelino @sboeuf @jodh-intel @grahamwhaley

pkg/annotations/annotations.go
```diff
@@ -0,0 +1,31 @@
...
+// limitations under the License.
+//
+
+package annotations
+
+const (
+	// KernelPath is a pod annotation for passing a per container path pointing at the kernel needed to boot the container VM.
+	KernelPath = "com.github.containers.virtcontainers.KernelPath"
+
+	// ImagePath is an pod annotation for passing a per container path pointing at the guest image that will run the container VM.
```
grahamwhaley: s/run/run in/

pod.go
```diff
@@ -525,12 +528,100 @@ func (p *Pod) createSetStates() error {
...
+		return err
+	}
+
+	imagePath, err := selectAssetFromAnnotation(podConfig, annotations.ImagePath, annotations.ImageHash)
+	if err != nil {
+		return err
+	}
+
+	// Do we want to update the hypervisor config only if both
+	// paths are not empty?
```
grahamwhaley: I think we are good either way:
- require both paths to be filled out, to ensure the end user clearly states exactly what they need
or
- allow the end user to just over-ride one of the parts.

Maybe the former is better long term, as it could save somebody only setting one part, and then later down the line picking up an incompatible 'other part' during an upgrade, and getting hard to find problems then occurring...

pod.go
```diff
@@ -17,14 +17,17 @@
 package virtcontainers
```
grahamwhaley: nit: in your commit message you say:
```
If we find a valid path for a kernel
or an image, we change the hypervisor config to point to them.
```

The `if` bothers me - if half implies that if the paths are not valid then we don't change the hypervisor (and carry on??) - presumably if there is an invalid config/annotation we will actually error out and not continue to launch the workload with the defaults?

pkg/annotations/annotations.go
```diff
@@ -0,0 +1,31 @@
...
+//
+
+package annotations
+
+const (
+	// KernelPath is a pod annotation for passing a per container path pointing at the kernel needed to boot the container VM.
+	KernelPath = "com.github.containers.virtcontainers.KernelPath"
+
+	// ImagePath is an pod annotation for passing a per container path pointing at the guest image that will run in the container VM.
+	ImagePath = "com.github.containers.virtcontainers.ImagePath"
```
jodh-intel: This is very explicit, but might be easier to maintain if we had instead:

```
const (
    vcAnnPrefix = "com.github.containers.virtcontainers."

    KernelPath = vcAnnPrefix + "KernelPath"
    ImagePath = vcAnnPrefix + "ImagePath"
        :
)
```

pod.go
```diff
@@ -525,12 +528,108 @@ func (p *Pod) createSetStates() error {
...
+
+	// TODO Check for asset type
+	switch t {
+	case kernelAsset:
+		return true
+	case imageAsset:
+		return true
+	}
+
+	return true
```
jodh-intel: Shouldn't this be `false`?

pkg/annotations/annotations.go
```diff
@@ -0,0 +1,31 @@
...
+
+	// KernelHash is a pod annotation for passing a container kernel image SHA-512 hash value.
+	KernelHash = "com.github.containers.virtcontainers.KernelHash"
```
jodh-intel: This might need to be `SHA512KernelHash` (ditto for the image) to be future-proof. Alternatively, another annotation for `HashType` (implying that both kernel + image must use the same hash algorithm of course).

pod.go
```diff
@@ -525,12 +528,108 @@ func (p *Pod) createSetStates() error {
...
+}
+
+func selectAssetFromAnnotation(podConfig *PodConfig, pathAnnotation, hashAnnotation string) (string, error) {
+	assetPath, ok := podConfig.Annotations[pathAnnotation]
+	if !ok {
+		return "", nil
+	}
+
+	if !filepath.IsAbs(assetPath) {
```
jodh-intel: This implicitly handles the case where `assetPath == ""`, but I wonder if this should be an explicit check (`if !ok || assetPath == ""` or something?). Same comment for `assetHashString`.

pod.go
```diff
@@ -525,12 +528,108 @@ func (p *Pod) createSetStates() error {
...
+
+	imagePath, err := selectAssetFromAnnotation(podConfig, annotations.ImagePath, annotations.ImageHash)
+	if err != nil {
+		return err
+	}
+
+	// Do we want to update the hypervisor config only if both
+	// paths are not empty?
+	if kernelPath != "" {
+		podConfig.HypervisorConfig.KernelPath = kernelPath
```
jodh-intel: Could you add a log call for both kernel + image when the config is overridden?

pod.go
```diff
@@ -525,12 +528,108 @@ func (p *Pod) createSetStates() error {
...
+	}
+
+	if imagePath != "" {
+		podConfig.HypervisorConfig.ImagePath = imagePath
+	}
+
+	if !validateAsset(kernelPath, kernelAsset) {
```
jodh-intel: Wouldn't it be better if these `validateAsset()` calls came before the change to `podConfig`?

sameo:
@sboeuf 
> Did I miss something explaining that we MUST implement this in virtcontainers ?

No you did not miss anything and this could be done by only modifying `utils/oci.go`. However, I did not want to do something OCI specific and I wanted the code to be generic and specification agnostic. I also realized it would be good to have a generic asset handling API for virtcontainers.
So it makes the code a little complex but I think it's cleaner.

Also, one thing I was considering is adding private fields to `HypervisorConfig` for not overwriting the public paths passed from above. I think it would be cleaner.

sboeuf:
Well, I think the problem is to have two ways to setup the same thing from virtcontainers perspective. We should not give the choice between either the Pod annotations, or the HypervisorConfig structure to the consumer of virtcontainers, otherwise we're just bringing confusion IMO.
Now, if you say we want to modify HypervisorConfig so that all its fields are private, then the annotations make sense since this would be the only way to setup the kernel/image. But there one issue about that, we cannot make those fields private because this structure is stored to our filesystem through a json marshalling and it needs to fields to be public.

sameo:
> Now, if you say we want to modify HypervisorConfig so that all its fields are private, then the annotations make sense since this would be the only way to setup the kernel/image. But there's one issue about that, we cannot make those fields private because this structure is stored to our filesystem through a json marshalling and it needs those fields to be public.

Oh, don't get me wrong: I don't want to have all `HypervisorConfig` fields private, only the ones we'd use to set assets not defined by the caller directly.

So I fully agree things could be done from the oci utils package or from the runtime directly, and it would make sense if the definition of alternative assets would be part of the OCI spec (Which may eventually become reality). In that case, we'd just use the OCI spec and translate that into an `HypervisorConfig` setting. 
But here I'm looking at the problem from a k8s point of view where the only thing we're going to be able to pass in to the runtime or virtcontainers are pod annotations. We can decide to translate those into `HypervisorConfig` fields in several places. In generalization order:
* From the runtime: This becomes CC specific
* From `virtcontainers/pkg/oci`: This becomes OCI specific
* From `virtcontainers/assets.go`: This becomes virtcontainers specific

So by adding a `customAssets []asset` field to `HypervisorConfig` or even to `Pod`, we have a generic implementation (not OCI or CC specific) and it is transparent to the user: There is still only one way to pass a configured set of assets (through `HypervisorConfig`) and Annotations would only allow the higher layers (Docker CLI, k8s) to overwrite the runtime configured assets.

pod.go
```diff
@@ -525,12 +525,48 @@ func (p *Pod) createSetStates() error {
...
+	if err != nil {
+		return err
+	}
+
+	image, err := newAsset(podConfig, imageAsset)
+	if err != nil {
+		return err
+	}
+
+	if kernel != nil && kernel.path != "" {
```
jodh-intel: This test confused me the first time I read it. Might be worth a comment on the function to state that it modifies the podConfig if explicit assets have been specified, but otherwise uses the default values. This would be slightly clearer if the `if` tests had an `else` specifying something like:

```
else {
    virtLog.Debug("using default kernel asset as no custom value specified")
}
```

In fact, if you added a `func (a *asset) override(podConfig *PodConfig) override`, couldn't you compress these 2 blocks into:

```
for _, a := range []asset{image, kernel{
  // this could handle a==nil and also call asset.valid()
  asset.override(podconfig)
}
```

sboeuf:
@sameo so you're saying that you want something generic for those annotations, but as of today, k8s relies on CRI which relies on OCI. That's why I consider this has to be handled at the OCI level and not at the virtcontainers level. But I think I understand that you want to be able to have any orchestrator using those annotations in order to give specific info to virtcontainers. The problem that I see here is that all orchestrations use a runtime (and hence a specification file). This means that we are creating a generic thing, but in practice, we will have to translate those annotations to PodConfig annotations  in every runtime implementation.

sboeuf:
You have some golint errors.
Let's also wait for @jodh-intel review now that you have addressed his comments.
LGTM

asset.go
```diff
@@ -0,0 +1,159 @@
...
+	computedHash string
+	kind         assetType
+}
+
+func (a *asset) valid() bool {
+	if !filepath.IsAbs(a.path) {
+		return false
+	}
+
+	// TODO Check for asset type
```
jodh-intel: Nit: can this comment be dropped now?

sameo:
I'll merge this one once the 17.04 CI is fixed.

PR #465, opened 2017-11-03
vendor: Replace ciao/qemu package

### LINKED ISSUE #431: qemu: Move ciao-qemu implementation to virtcontainers repo
We are constantly adding new things to ciao-qemu in order to support new features for our runtime, but ciao itself is not using them. There is no point in sharing this between both ciao and virtcontainers, and it would make easier not to have to revendor.

sboeuf:
This PR introduces a new package `pkg/qemu` inside the virtcontainers repository. This package is basically a copy of what can be found into github.com/ciao-project/ciao (only changed some names). After that, it changes the virtcontainers code (qemu implementation of the hypervisor interface) by relying onto this new package instead of ciao package.
And as the last part, it completely removes the dependency on ciao.

Fixes #431


jodh-intel:
lgtm.

It would be good to establish and document a strategy to ensure our version of the code stays in sync with the original (to ensure both projects benefit from performance/reliability/security changes).

sameo:
I'm fine with that as long as @markdryan is ok as well

sboeuf:
After discussion with @markdryan and @mcastelino about this, we have decided that qemu package will move to its own repo somewhere else. Let's drop this PR.

PR #469, opened 2017-11-06
.ci: Fix coveralls reports

sboeuf:
Coveralls needs its report to be performed against master branch to
report properly the code coverage to Github.

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -36,9 +36,9 @@ go get ${vc_repo} || true
 cd "${GOPATH}/src/${vc_repo}"
 if [ "${ghprbPullId}" ] && [ "${ghprbTargetBranch}" ]
 then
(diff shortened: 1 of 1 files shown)


jodh-intel:
lgtm

sboeuf:
@grahamwhaley Coveralls is not properly reported because we are using origin/master instead of master. The name of the branch matters apparently and this should help our Coveralls reports to be recognized as the right ones

jodh-intel:
@sboeuf - does that mean we also need to update:

- https://github.com/clearcontainers/tests/blob/master/.ci/jenkins_job_build.sh#L43..L79

... and could this maybe explain https://github.com/clearcontainers/jenkins/issues/9?

sboeuf:
@jodh-intel Yes I think we should do the same for the script in tests repo, but I am not sure this explains the sporadic issue clearcontainers/jenkins#9

PR #470, opened 2017-11-08
network: Replace CNI Result structure

sboeuf:
The goal of this PR is to replace the dependency on CNI structure
types.Result with a new one. The new structure relies on very generic
Go structures such as net.Interface, netlink.Route, and net.IPNet in
order to provide a lot of information about the underlying network.

The choice has been made to translate only the DNS information coming
from the CNI network implementation because other data can be found
through a simple network scan. And because this scan has to be
performed anyway (need MTU which is not provided by CNI result for
instance), it is easier to fill network interfaces, IP addresses and
routes structures from scratch.

Also, notice that CNI vendoring has been updated to rely on latest definition
of what is considered to be a network Result.


cni.go
```diff
@@ -65,6 +107,26 @@ func (n *cni) deleteVirtInterfaces(networkNS NetworkNamespace) error {
 	return nil
 }
 
+func (n *cni) fillEndpointsPropertiesFromScan(networkNS *NetworkNamespace) error {
+	for idx, endpoint := range networkNS.Endpoints {
+		netIface, err := getIfaceByNameFromNetNs(networkNS.NetNsPath, endpoint.NetPair.VirtIface.Name)
```
mcastelino: @sboeuf it would be better to get the information using netlink call in the implementation. I think the netlink call provides all the information and some additional info https://godoc.org/github.com/vishvananda/netlink#Handle.LinkByName

Do you see an advantage using the current method.

network.go
```diff
@@ -68,9 +68,26 @@ const (
 	defaultQueues     = 8
 )
 
-type netIfaceAddrs struct {
-	iface net.Interface
-	addrs []net.Addr
+// NetIfaceAddrs describes a network interface an its associated IP addresses.
+type NetIfaceAddrs struct {
+	Iface net.Interface
```
mcastelino: net.Interface gives you
```
// Interface represents a mapping between network interface name
  // and index. It also represents network interface facility
  // information.
  type Interface struct {
  	Index        int          // positive integer that starts at one, zero is never used
  	MTU          int          // maximum transmission unit
  	Name         string       // e.g., "en0", "lo0", "eth0.100"
  	HardwareAddr HardwareAddr // IEEE MAC-48, EUI-48 and EUI-64 form
  	Flags        Flags        // e.g., FlagUp, FlagLoopback, FlagMulticast
  }
  ```

however  netlink.Link.Attrs() gives you
```
// LinkAttrs represents data shared by most link types
type LinkAttrs struct {
	Index        int
	MTU          int
	TxQLen       int // Transmit Queue Length
	Name         string
	HardwareAddr net.HardwareAddr
	Flags        net.Flags
	RawFlags     uint32
	ParentIndex  int         // index of the parent link device
	MasterIndex  int         // must be the index of a bridge
	Namespace    interface{} // nil | NsPid | NsFd
	Alias        string
	Statistics   *LinkStatistics
	Promisc      int
	Xdp          *LinkXdp
	EncapType    string
	Protinfo     *Protinfo
	OperState    LinkOperState
	NetNsID      int
}
```

which is much more comprehensive. So how about switching to that? That will help longer term.

File: network.go [hunk 3]
```diff
@@ -68,9 +68,26 @@ const (
...
+	Servers  []string
+	Domain   string
+	Searches []string
+	Options  []string
+}
+
+// NetworkInfo gathers all information related to a network interface.
+// It can be used to store the description of the underlying network.
+type NetworkInfo struct {
+	Iface  NetIfaceAddrs
```
mcastelino: Same here. Maybe we can switch the netlink.Link
This has another benefit. This will allow us to make interconnection decisions based on the type of the link, which we cannot do today without an extra scan.

For example if the interface is of type tap, macvtap we can directly connect it.

network.go
```diff
@@ -99,7 +116,7 @@ type NetworkConfig struct {
 // Endpoint gathers a network pair and its properties.
 type Endpoint struct {
```
mcastelino: An unrelated comment. But in lot of functions we pass in NetPair. I feel in all cases we should pass in the Endpoint struct vs the NetPair struct as this has more info which these functions will need.

mcastelino: Also not all endpoints have netpairs. Example is SRIOV today and vhost-user/tap in the future. So using endpoint as the struct that gets passed around will ensure that there is no significant rewrite in the future as we add support for more interface types.

sboeuf:
@mcastelino I like the idea of relying more on the netlink package, it's gonna be consistent with the Route structure that I get from netlink and I agree it will bring more info ;)

sboeuf:
@mcastelino @sameo @amshinde please take a look (again). I have rebased on top of @amshinde PR for SRIOV support. I think I did this right and then I have used netlink library, which really helped me getting the code simplified. Tell me what you think about it.
@amshinde BTW, one of my commit is fixing CNI case for physical device. Indeed, I have realized that we had missed the CNI case since we were not scanning the underlying network at that time.

cni.go
```diff
@@ -118,24 +120,26 @@ func (n *cni) deleteVirtInterfaces(networkNS NetworkNamespace) error {
...
-			return err
-		}
+	netnsHandle, err := netns.GetFromPath(networkNS.NetNsPath)
+	if err != nil {
+		return err
+	}
+	defer netnsHandle.Close()
 
-		prop.Iface = netIface
+	netlinkHandle, err := netlink.NewHandleAt(netnsHandle)
```
mcastelino: @sboeuf I like this change.

network.go
```diff
@@ -74,10 +73,10 @@ const (
 	defaultQueues     = 8
 )
 
-// NetIfaceAddrs describes a network interface an its associated IP addresses.
+// NetIfaceAddrs describes a network interface and its associated IP addresses.
 type NetIfaceAddrs struct {
-	Iface net.Interface
-	Addrs []net.IPNet
+	Iface netlink.LinkAttrs
```
mcastelino: @sboeuf I think it is better to actually have the actual link vs the LinkAttrs. The reason is that the link will help us get the interface type. If you only cache the attrs we lose information that we need.

network.go
```diff
@@ -91,7 +90,8 @@ type DNSInfo struct {
 // NetworkInfo gathers all information related to a network interface.
 // It can be used to store the description of the underlying network.
 type NetworkInfo struct {
-	Iface  NetIfaceAddrs
+	Iface  netlink.LinkAttrs
```
mcastelino: As before, can we just store the link vs LinkAttrs. You can always get LinkAttrs from Link

network.go
```diff
@@ -1032,142 +1032,61 @@ func createNetworkEndpoints(numOfEndpoints int) (endpoints []Endpoint, err error
...
+	return NetworkInfo{
+		Iface:  *attrs,
+		Addrs:  addrs,
+		Routes: routes,
+	}, nil
+}
 
-		return nil
-	})
+func networkInfoFromIfaceName(handle *netlink.Handle, ifaceName string) (NetworkInfo, error) {
```
mcastelino: An observation. If we have the link index cached we should use the link index. The reason is that the LinkByName is highly inefficient and take much longer to complete. So we should see if we can use the index as far as possible when it is available, so that network setup does not contribute to start latency.

network.go
```diff
@@ -1089,6 +1081,66 @@ func networkInfoListFromNetworkScan(handle *netlink.Handle) ([]NetworkInfo, erro
...
+	idx := 0
+	for _, netInfo := range netInfoList {
+		var endpoint Endpoint
+
+		// Skip any loopback interface.
+		if (netInfo.Iface.Flags & net.FlagLoopback) != 0 {
+			continue
+		}
+
+		if err := doNetNS(networkNSPath, func(_ ns.NetNS) error {
```
mcastelino: A thought here. Given that netlink is fully namespace aware though the handle. If is pass the ns handle to the network function do we need this?

cni.go
sboeuf: Done

network.go
sboeuf: Done

network.go
sboeuf: Done

network.go
sboeuf: Nope we cannot since Physical devices needs to be analyzed with ethtool and this has to happen from the netns.

network.go
sboeuf: This function is removed in another commit, no need to worry about it, but thanks for the heads up :)

sameo:
LGTM

sboeuf:
@amshinde I'd like to have your input before to merge this PR since I have rebased on top of your SRIOV work ;)

network.go
```diff
@@ -1209,25 +1203,10 @@ func createPhysicalEndpoint(ifaceName string) (*PhysicalEndpoint, error) {
...
-	mac := link.Attrs().HardwareAddr.String()
-	MTU := link.Attrs().MTU
-
 	physicalEndpoint := &PhysicalEndpoint{
-		IfaceName:      ifaceName,
-		HardAddr:       mac,
-		MTU:            MTU,
+		IfaceName:      netInfo.Iface.Name,
+		HardAddr:       netInfo.Iface.HardwareAddr.String(),
+		MTU:            netInfo.Iface.MTU,
```
amshinde: We should drop the MTU field from the Physical Endpoint now, since we can access it from endpoint.Properties().Iface.MTU now as we don't rely on the second scan for the MTU value.

network.go
```diff
@@ -82,10 +82,16 @@ type DNSInfo struct {
 	Options  []string
```
amshinde: This is regarding your commit message. Physical interfaces were being skipped in the second scan

sboeuf: Yes but this means we could not retrieve the Link type in that case (if needed obviously).

network.go
sboeuf: Yes I can probably do that :)

sboeuf: Done

PR #473, opened 2017-11-09
api: Simplify drive hotplugging

sboeuf:
The way CreatePod() was implemented, it was making an extra call
to hotplug all the drives. But when StartPod() is called, it leads
eventually to Container.start() to be invoked, resulting into a
hotplug of the drives related to this container.

A mechanism based on a flag was in place to detect if the drive had
been previously hotplugged, but there is no need for such thing
anymore. Indeed, we know that we are hotplugging only from the
container implementation.

### DIFF
--- a/api.go
+++ b/api.go
@@ -82,12 +82,6 @@ func CreatePod(podConfig PodConfig) (VCPod, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


jodh-intel:
lgtm

PR #471, opened 2017-11-09
 vfio: add support for hot plugging VFIO devices

devimc:
This patch adds support for hot pluging VFIO devices and
remove the cold plug support

fixes #481

Signed-off-by: Julio Montes <julio.montes@intel.com>

PR #474, opened 2017-11-09
block: Skip cold-plugging block devices.

amshinde:
Since vfio devices cannot be hotplugged currently, we
need to call atachDevices from pod.go.
We can avoid cold plugging block devices with this check for now.

Fixes https://github.com/clearcontainers/runtime/issues/802

### DIFF
--- a/device.go
+++ b/device.go
@@ -187,6 +187,15 @@ func makeBlockDevIDForHypervisor(deviceID string) string {
 }
 
 func (device *BlockDevice) attach(h hypervisor, c *Container) (err error) {
+	// If VM has not been launched yet, return immediately.
+	// This is because we always want to hotplug block devices.
+	// Eventually attachDevices will be called only after VM is launched,
+	// and this check can be taken out.
+	// See https://github.com/containers/virtcontainers/issues/444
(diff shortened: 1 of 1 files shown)


sboeuf:
It's all good !!
LGTM

sameo:
LGTM

device.go
```diff
@@ -187,6 +187,15 @@ func makeBlockDevIDForHypervisor(deviceID string) string {
 }
 
 func (device *BlockDevice) attach(h hypervisor, c *Container) (err error) {
```
mcastelino: @devimc @amshinde are we doing this only for block devices. Can we do this for network devices also?

sboeuf: This PR only intends to fix https://github.com/clearcontainers/runtime/issues/802
We have discussed that @devimc is gonna submit a more global PR for removing all the coldplug codepaths (see my comment on #471). This PR will be blocked until we move to qemu2.9

mcastelino: fair enough.

devimc:
lgtm

jodh-intel:
lgtm

PR #477, opened 2017-11-10
filesystem: Fix fetchFile behaviour for devicesFileType

sboeuf:
In a previous patch, I have been simplifying the way we were fetching
stored information from the filesystem, but I simplified too much,
because I forgot the specific case of devicesFileType.

This commit fixes this lack of support by calling into the specific
function fetchDeviceFile() in case of devicesFileType as a resource.


sameo:
LGTM

jodh-intel:
lgtm

PR #482, opened 2017-11-11
pod: Improve pod stop/delete sequence

sboeuf:
This PR first improves the way the tear down of the pod is performed from the agent. Until now, nothing was done and the VM was stopped without tearing down properly the things previously set up inside the VM.
Then, it modifies the QEMU implementation for the `stopPod()` method, by not waiting for the VM to exit. Instead, it send a SHUTDOWN through QMP but it does not wait for the VM since there is really nothing it can do.

### DIFF
--- a/pod.go
+++ b/pod.go
@@ -581,10 +581,6 @@ func createPod(podConfig PodConfig) (*Pod, error) {
 		return nil, err
 	}
 
-	if err := p.storage.storeHypervisorState(p.id, p.hypervisor.getState()); err != nil {
-		return nil, err
-	}
-
 	return p, nil
 }
 
@@ -629,10 +625,6 @@ func doFetchPod(podConfig PodConfig) (*Pod, error) {
 		wg:              &sync.WaitGroup{},
 	}
 
-	if err := p.hypervisor.init(p); err != nil {
-		return nil, err
-	}
-
 	containers, err := newContainers(p, podConfig.Containers)
 	if err != nil {
 		return nil, err
@@ -641,6 +633,11 @@ func doFetchPod(podConfig PodConfig) (*Pod, error) {
 	p.containers = containers
 
 	if err := p.storage.createAllResources(*p); err != nil {
+		return nil, err
+	}
+
+	if err := p.hypervisor.init(p); err != nil {
+		p.storage.deletePodResources(p.id, nil)
 		return nil, err
 	}
 
--- a/qemu.go
+++ b/qemu.go
@@ -803,7 +803,6 @@ func (q *qemu) waitPod(timeout int) error {
 func (q *qemu) stopPod() error {
 	cfg := ciaoQemu.QMPConfig{Logger: newQMPLogger()}
 	disconnectCh := make(chan struct{})
-	const timeout = time.Duration(10) * time.Second
 
 	q.Logger().Info("Stopping Pod")
 	qmp, _, err := ciaoQemu.QMPStart(q.qmpControlCh.ctx, q.qmpControlCh.path, cfg, disconnectCh)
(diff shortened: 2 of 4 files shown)


sboeuf:
This PR is blocked by https://github.com/clearcontainers/agent/pull/158 and https://github.com/clearcontainers/agent/pull/160 to be merged in order to behave properly.

sameo:
LGTM

sboeuf:
@sameo Now everything is passing! I'd like you to give this a second review since I have added patches to solve the issue introduced by the fact we're not waiting for the end of the VM anymore.

mcastelino:
@amshinde @sameo @sboeuf the only worry I have with this change is disk cache flushes when using virtio-blk. @amshinde do we do a flush in the VM prior to shutdown? It would be prudent to ensure we do flush before we tore down the VM.

mcastelino:
@sboeuf thanks for https://github.com/clearcontainers/agent/pull/169

LGTM

qemu.go
```diff
@@ -542,16 +525,23 @@ func (q *qemu) init(pod *Pod) error {
...
 
-	if err = pod.storage.fetchHypervisorState(pod.id, &q.state); err != nil {
+	if err := pod.storage.fetchHypervisorState(pod.id, &q.state); err != nil {
 		q.Logger().Debug("Creating bridges")
 		q.state.Bridges = NewBridges(q.config.DefaultBridges, q.config.HypervisorMachineType)
+
+		q.Logger().Debug("Creating UUID")
+		q.state.UUID = uuid.Generate().String()
+
+		if err := pod.storage.storeHypervisorState(pod.id, q.state); err != nil {
```
sameo: We should keep the hypervisor state storage either in pod.go , after the `hypervisor.init()` call.

qemu.go
```diff
@@ -613,6 +603,23 @@ func (q *qemu) setMemoryResources(podConfig PodConfig) (ciaoQemu.Memory, error)
 	return memory, nil
 }
 
+func (q *qemu) buildQMPSocketPath(socketName string) (string, error) {
```
sameo: nit: `qmpSocketPath()` would be enough

pod.go
```diff
@@ -600,6 +596,8 @@ func doFetchPod(podConfig PodConfig) (*Pod, error) {
 		return nil, err
 	}
 
+	agentConfig := newAgentConfig(podConfig)
+
```
sameo: Why are we creating the `agentConfig` earlier here ?

sboeuf: I thought I was cleaning up the sequence but I had not realized that was `agentConfig` we were getting from this. Nevermind, I am gonna move this back !

qemu.go
sboeuf: Sure, let's keep things simple.

qemu.go
sboeuf: Well, this was done on purpose since I think it should be handled from the hypervisor implementation itself. Indeed, this state file is specific to every hypervisor implementation so that they can store some specific information. But let's say we have other hypervisor implementation that don't need to store/fetch anything, this means we would force them to store nothing.
I agree we would avoid code duplication if storing from pod.go, but in that case I think we should do everything from container/pod code. Or we go with everything controlled from the hypervisor implementation. The huge benefit with having this controlled from the hypervisor implementation is that we don't need to make assumption about when we should fetch/store from the pod/container.
I think we can take this change as is, and if you think we should move the entire fetch/store out of the qemu implementation because we can find a common pattern for all hypervisors about storage, then this should be done in a follow up PR.
Let me know what you think !

sameo: Fair enough, I'm good with that.

sameo:
LGTM

PR #483, opened 2017-11-13
One proxy per vm

### LINKED ISSUE #478: Use one proxy per VM
Modify code to launch one proxy instance per pod rather than relying on the single host-wide proxy instance.

jodh-intel:
proxy: Launch one proxy instance per pod
    
Create a proxy instance for each pod (virtual machine).
    
Since the proxy is now launched by virtcontainers, it is necessary to specify the path to the proxy binary
in the pod configuration.
    
Fixes #478.

### DIFF
--- a/api.go
+++ b/api.go
@@ -86,6 +86,12 @@ func createPodFromConfig(podConfig PodConfig) (*Pod, error) {
 		return nil, err
 	}
 
+	// Start the proxy
+	err = p.startProxy()
+	if err != nil {
+		return nil, err
+	}
+
 	// Start shims
 	if err := p.startShims(); err != nil {
 		return nil, err
--- a/cc_proxy.go
+++ b/cc_proxy.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"os/exec"
+	"path/filepath"
 	"time"
 
 	"github.com/clearcontainers/proxy/client"
@@ -41,7 +43,8 @@ type ccProxy struct {
 // CCProxyConfig is a structure storing information needed for
 // the Clear Containers proxy initialization.
 type CCProxyConfig struct {
-	URL string
+	Path  string
+	Debug bool
 }
 
 // connectProxyRetry repeatedly tries to connect to the proxy on the specified
@@ -96,12 +99,12 @@ func (p *ccProxy) connectProxyRetry(scheme, address string) (conn net.Conn, err 
 	}
 }
 
-func (p *ccProxy) connectProxy(proxyURL string) (*client.Client, error) {
-	if proxyURL == "" {
-		proxyURL = defaultCCProxyURL
+func (p *ccProxy) connectProxy(uri string) (*client.Client, error) {
+	if uri == "" {
+		return nil, fmt.Errorf("no proxy URI")
 	}
 
-	u, err := url.Parse(proxyURL)
+	u, err := url.Parse(uri)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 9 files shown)


jodh-intel:
Blocked on https://github.com/clearcontainers/proxy/pull/167.

api.go
```diff
@@ -82,6 +82,12 @@ func CreatePod(podConfig PodConfig) (VCPod, error) {
 		return nil, err
 	}
 
+	// Start the proxy
+	err = p.createProxyProcess()
```
sboeuf: You should move that into `p.startVM(netNsPath)` function, so that we don't expand too much `CreatePod()` and `RunPod()`.

api.go
```diff
@@ -126,6 +132,11 @@ func DeletePod(podID string) (VCPod, error) {
 		return nil, err
 	}
 
+	err = p.waitForProxy()
```
sboeuf: Similar to `createProxyProcess()`, I think this should be moved to `stopVM()`.

pod.go
```diff
@@ -774,6 +776,45 @@ func (p *Pod) startVM(netNsPath string) error {
...
+func (p *Pod) waitForProxy() error {
+	pid := p.proxyProcess.Pid
+	if pid == 0 {
+		return nil
+	}
+
+	// cannot fail on Linux
+	proc, _ := os.FindProcess(pid)
+
+	_, err := proc.Wait()
```
sboeuf: We should add a timeout here.

pod.go
```diff
@@ -774,6 +776,45 @@ func (p *Pod) startVM(netNsPath string) error {
...
+	p.proxyProcess = newProcess("", pid)
+
+	p.Logger().WithField("proxy-pid", pid).Debug("proxy started")
+
+	return nil
+}
+
+// waitForProxy waits for the proxy process to finish
+func (p *Pod) waitForProxy() error {
+	pid := p.proxyProcess.Pid
```
sboeuf: I don't see any place where you save this info about the proxy PID. When `DeletePod()` will be called, I think that `p.proxyProcess` is going to be empty.

sboeuf:
https://github.com/clearcontainers/proxy/pull/167 has been merged, there is no blocker left for this PR.

cc_proxy.go
```diff
@@ -127,17 +130,44 @@ func (p *ccProxy) connectProxy(proxyURL string) (*client.Client, error) {
...
+		return -1, "", fmt.Errorf("Wrong proxy config type, should be CCProxyConfig type")
+	}
+
+	if config.Path == "" {
+		return -1, "", fmt.Errorf("Proxy path cannot be empty")
+	}
+
+	// construct the socket path the proxy instance will use
+	socketPath := filepath.Join(runStoragePath, pod.id, "proxy.sock")
+	uri := fmt.Sprintf("unix://%s", socketPath)
```
devimc: I wonder if we have to ensure proxy's socket mode is 600

api.go
jodh-intel: ()Funnily enough, it was there originally, but I moved it out for clarity since `startVM` wasn't just starting the VM so the name was confusing. Also, it matched more closely how we handle the shims (`p.startShims()`).

On the topic of `CreatePod()` and `RunPod()`, I've just raised #486 to remove the duplication so how about we land that first and I can then rebase this PR and there won't be any duplication.

cc_proxy.go
jodh-intel: This is just constructing a path to give to the proxy. The proxy itself creates the socket (with perms `0660`) here:

- https://github.com/clearcontainers/proxy/blob/master/proxy.go#L674

pod.go
jodh-intel: Good catch - fixing...

cc_proxy.go
devimc: ohh I see, thanks @jodh-intel for the link

pod.go
```diff
@@ -774,6 +776,45 @@ func (p *Pod) startVM(netNsPath string) error {
...
+
+	// save state
+	p.state.URL = uri
+	if err := p.setPodState(p.state); err != nil {
+		return err
+	}
+
+	p.proxyProcess = newProcess("", pid)
+
+	p.Logger().WithField("proxy-pid", pid).Debug("proxy started")
```
sameo: I'd add the pod ID as part of the log here.

api.go
```diff
@@ -264,6 +275,12 @@ func RunPod(podConfig PodConfig) (VCPod, error) {
 		return nil, err
 	}
 
+	// Start the proxy
+	err = p.createProxyProcess()
```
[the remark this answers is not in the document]
jodh-intel: Good call - fixing...

pod.go
jodh-intel: Actually, `Logger()` will already add that for us ;)

pod.go
jodh-intel: Good call.

Looking at this again, the code actually won't work as we can't guarantee the calling process (now) owns the `cc-proxy` instance (it's likely to have been reparented to PID 1).

Do we already have BKM code somewhere for a (deep breath) pid or`/proc/$pid`-sniffing loop? I'm pretty sure we used to a long time ago but can't find it atm.

cc_proxy.go
```diff
@@ -127,17 +130,44 @@ func (p *ccProxy) connectProxy(proxyURL string) (*client.Client, error) {
...
+
+	args := []string{config.Path, "-uri", uri}
```
sameo: Maybe we could modify the process comm here by using something like `"proxy-%s", pod.ID`.

jodh-intel: We already have `pod.id` in the path as the socket is placed into the pod-specific directory. We could also add the pod ID to the socket name, but if we do, we'd be getting very close to the 107 byte socket name limit.

jodh-intel:
Branch updated with some simplifications... and more tests! :smile:

jodh-intel:
CI is failing due to the API breakage introduced by this PR (`CCProxyConfig` has changed).

sboeuf:
@jodh-intel then you have to submit the PR on the runtime and point this PR in your commit message.

sboeuf:
@jodh-intel yes sure, let me check that right now

jodh-intel:
@sboeuf - branch updated again.

sboeuf:
@jodh-intel thx, let me take a look !

sboeuf:
LGTM

sboeuf:
Thanks for the PR @jodh-intel, everything looks good now.
@sameo could you please take a look before I merge this ?

sameo:
LGTM

PR #484, opened 2017-11-14
pod: Improve pod create performances

sboeuf:
pod: Improve pod create performances
    
In order to optimize pod creation, this patch does not wait for the
VM after it has been started. Instead, it will wait for it during the
"start" stage. This will allow our VM to be started while the caller
could potentially do other things between create and start. Globally,
this improves the boot time of our containers.

### DIFF
--- a/pod.go
+++ b/pod.go
@@ -23,7 +23,6 @@ import (
 	"strings"
 	"sync"
 	"syscall"
-	"time"
 
 	"github.com/sirupsen/logrus"
 )
@@ -38,6 +37,10 @@ const controlSocket = "ctrl.sock"
 // This is a socket that any monitoring entity will listen to in order
 // to understand if the VM is still alive or not.
 const monitorSocket = "monitor.sock"
+
+// vmStartTimeout represents the time in seconds a pod can wait before
+// to consider the VM starting operation failed.
+const vmStartTimeout = 10
 
 // stateString is a string representing a pod state.
 type stateString string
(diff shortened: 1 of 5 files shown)


pod.go
```diff
@@ -835,6 +814,27 @@ func (p *Pod) start() error {
 		return err
 	}
 
+	l := p.Logger()
+
+	vmStartedCh := make(chan error)
+	const timeout = time.Duration(10) * time.Second
+
+	go p.hypervisor.waitPod(vmStartedCh)
```
sameo: Can we simply make `waitPod` blocking and pass it a timeout instead?

qemu.go
```diff
@@ -775,20 +741,58 @@ func (q *qemu) createPod(podConfig PodConfig) error {
...
+
+	for {
+		qmp, ver, err = ciaoQemu.QMPStart(q.qmpMonitorCh.ctx, q.qmpMonitorCh.path, cfg, disconnectCh)
+		if err == nil {
+			break
+		}
+
+		q.Logger().WithError(err).Warn("Failed to connect to QEMU instance, retrying...")
+		time.Sleep(time.Duration(50) * time.Millisecond)
+	}
```
sameo: This potentially is an infinite loop. By giving `waitPod` a timeout, you could implement a non infinte loop here based on the timeout argument.

qemu.go
```diff
@@ -775,20 +741,58 @@ func (q *qemu) createPod(podConfig PodConfig) error {
...
+	var ver *ciaoQemu.QMPVersion
+	var err error
+
+	for {
+		qmp, ver, err = ciaoQemu.QMPStart(q.qmpMonitorCh.ctx, q.qmpMonitorCh.path, cfg, disconnectCh)
+		if err == nil {
+			break
+		}
+
+		q.Logger().WithError(err).Warn("Failed to connect to QEMU instance, retrying...")
```
sameo: I would not print a new log entry every 50ms, but print one if we timeout instead.

pod.go
sboeuf: Sure I can do that. The reason I have chosen such an implementation is to avoid the duplication of timeout handling for every hypervisor implementation.

qemu.go
sboeuf: I'll change that.

qemu.go
sboeuf: Yes I agree, but I thought that every hypervisor implementation will have to implement the same mechanism which is duplication. I know this could be an infinite loop but when the main process returns, this go routine is gonna be terminated.

sameo:
LGTM

sboeuf:
@sameo comments addressed !

PR #488, opened 2017-11-18
cni: Fix bug while updating endpoints after scanning net namespace

amshinde:
After scanning the network namespace, we need to update the scanned
endpoints with the DNS information, and use the scanned endpoints
thereafter discarding the endpoints structures we had created prior
to calling into the CNI plugin.

There was a bug in the update, which was causing the entire endpoint
structure being assigned back to the scanned endpoints causing us to
lose information from the scan including if a virtual or physical
endpoint was found in the scan.

### DIFF
--- a/cni.go
+++ b/cni.go
@@ -132,6 +132,11 @@ func (n *cni) updateEndpointsFromScan(networkNS *NetworkNamespace) error {
 				prop := endpoint.Properties()
 				prop.DNS = ep.Properties().DNS
 				endpoint.SetProperties(prop)
+
+				switch e := endpoint.(type) {
+				case *VirtualEndpoint:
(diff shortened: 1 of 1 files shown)


sameo:
LGTM
But CNI unit tests are failing.

amshinde:
@sameo Units tests are passing now.
@sboeuf Does this look good to you?

sboeuf:
LGTM
Waiting for @amshinde to run a local test. I will merge right after that

amshinde:
@sboeuf I have run a manual test with cni-bridge plugin as well,able to see the network setup correctly and the pod is deleted successfully.

sboeuf:
@amshinde thanks for that !

PR #490, opened 2017-11-20
Remove redundant "if" checks

WeiZhang555:
Some "if" checks are redundant, we can remove them and return error
directly instead.

Signed-off-by: Zhang Wei <zhangwei555@huawei.com>

### DIFF
--- a/container.go
+++ b/container.go
@@ -225,11 +225,7 @@ func (c *Container) startShim() error {
 
 	c.process = *process
 
-	if err := c.storeProcess(); err != nil {
-		return err
-	}
-
-	return nil
+	return c.storeProcess()
 }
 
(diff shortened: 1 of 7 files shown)


WeiZhang555:
This is detected by a latest golint tools, I can see that you have enabled gometalinter tools in you CI, I guess the reason that gometalinter didn't detected this is that it's using a old golint tool.

As the issue is real, I think it's better to fix this, or we still need to fix this someday when we updated gometalinter. :smile:

sboeuf:
LGTM

sboeuf:
@WeiZhang555 thanks for the PR, I agree it's good to clean those asap!

WeiZhang555:
@sboeuf 🤝 👍

PR #503, opened 2017-12-11
ethtool: Remove .swp file

sameo:
Probably was pulled in with a git add .


devimc:
lgtm

PR #507, opened 2017-12-11
cni: pass pod ID to CNI plugin's add/delete calls

egernst:
We were passing a virtualEndpoint.NetPair.ID previously.

CNI expects to receive a POD ID when being called.  This patch changes
the CNI calls to provide this behavior.

Signed-off-by: Eric Ernst <eric.ernst@intel.com>

### DIFF
--- a/cni.go
+++ b/cni.go
@@ -66,7 +66,7 @@ func convertCNIResult(cniResult cniTypes.Result) (NetworkInfo, error) {
 	}
 }
 
-func (n *cni) addVirtInterfaces(networkNS *NetworkNamespace) error {
+func (n *cni) addVirtInterfaces(pod Pod, networkNS *NetworkNamespace) error {
 	netPlugin, err := cniPlugin.NewNetworkPlugin()
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


egernst:
Need to test locally still - WIP

jodh-intel:
@egernst, @mcastelino - is there any way we could create a minimal `cni_test.go` for this and start to build up a set of unit tests? Or would that require a lot of infrastructure to allow them to run?

sboeuf:
LGTM

egernst:
@jodh-intel  - This would require at least creating a network namespace with a number of endpoints in it.  AFAICT, we wouldn't need a POD and this part could just be "faked".   I'd need to take a look at what exists today in the network common testing...  Can we open an issue for adding this?  I think it'll be a bit more involved, and I'd love to fix this bug in the meantime.

sboeuf:
I agree with @egernst, this can be part of a follow up PR to solve the more global handling of CNI testing. @egernst please open an issue and reference it here. I'll merge this PR in the meantime.

sboeuf:
@egernst could you remove `do-not-merge` label ?

PR #512, opened 2017-12-13
Add support for CPU hot add/remove

### LINKED ISSUE #511: Support CPU hotplugging
Add support for CPU hotplugging

devimc:
For hot-add, we have to know what CPUs are available using
`query-hotpluggable-cpus`.
For hot-remove we only need to know the CPUs ids in order to avoid the
use of 1 query 1 cpu, hotplugAddCPUs and hotRemoveCPUs receive the
total amount of CPUs to hot add or hot remove.

depending of the container's state, pod's resources are updated.
for example, if the container has a CPU constraint of 4 vCPU and it is stopped,
then these 4 vCPUs are removed.
another example, if the container is ready to be started and it has a CPU
constraint of 3 vCPUs, 3 vCPUs are added and the constraint is sent to the
agent to limit CPU usage of the container.
With this patch the fractional part of the CPU constraint is also honoured,
for example, if the container has a CPU constraint of 5.5, then 6 vCPUs are
added to the POD and a constraint of 5.5 CPUs is applied to the container

fixes #511

### DIFF
--- a/container.go
+++ b/container.go
@@ -58,6 +58,16 @@ type ContainerStatus struct {
 	Annotations map[string]string
 }
 
+// ContainerResources describes container resources
+type ContainerResources struct {
+	// CPUQuota specifies the total amount of time in microseconds
+	// The number of microseconds per CPUPeriod that the container is guaranteed CPU access
+	CPUQuota int64
+
+	// CPUPeriod specifies the CPU CFS scheduler period of time in microseconds
+	CPUPeriod uint64
+}
+
 // ContainerConfig describes one container runtime configuration.
 type ContainerConfig struct {
 	ID string
@@ -80,6 +90,9 @@ type ContainerConfig struct {
 
 	// Device configuration for devices that must be available within the container.
 	DeviceInfos []DeviceInfo
+
+	// Resources container resources
+	Resources ContainerResources
 }
 
 // valid checks that the container configuration is valid.
@@ -436,6 +449,10 @@ func createContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 		return nil, err
 	}
 
+	if err := c.addResources(); err != nil {
+		return nil, err
+	}
+
 	// Deduce additional system mount info that should be handled by the agent
 	// inside the VM
 	c.getSystemMountInfo()
@@ -588,6 +605,10 @@ func (c *Container) stop() error {
 	}
 
 	if err := c.pod.agent.stopContainer(*(c.pod), *c); err != nil {
+		return err
+	}
+
+	if err := c.removeResources(); err != nil {
 		return err
 	}
 
@@ -748,6 +769,40 @@ func (c *Container) attachDevices() error {
 func (c *Container) detachDevices() error {
 	for _, device := range c.devices {
 		if err := device.detach(c.pod.hypervisor); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (c *Container) addResources() error {
+	//TODO add support for memory, Issue: https://github.com/containers/virtcontainers/issues/578
+	if c.config == nil {
+		return nil
+	}
+
+	vCPUs := ConstraintsToVCPUs(c.config.Resources.CPUQuota, c.config.Resources.CPUPeriod)
+	if vCPUs != 0 {
+		virtLog.Debugf("hot adding %d vCPUs", vCPUs)
+		if err := c.pod.hypervisor.hotplugAddDevice(uint32(vCPUs), cpuDev); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (c *Container) removeResources() error {
+	//TODO add support for memory, Issue: https://github.com/containers/virtcontainers/issues/578
+	if c.config == nil {
+		return nil
+	}
+
+	vCPUs := ConstraintsToVCPUs(c.config.Resources.CPUQuota, c.config.Resources.CPUPeriod)
+	if vCPUs != 0 {
+		virtLog.Debugf("hot removing %d vCPUs", vCPUs)
+		if err := c.pod.hypervisor.hotplugRemoveDevice(uint32(vCPUs), cpuDev); err != nil {
 			return err
 		}
 	}
--- a/hyperstart_agent.go
+++ b/hyperstart_agent.go
@@ -424,6 +424,13 @@ func (h *hyper) startOneContainer(pod Pod, c *Container) error {
 		Process: process,
 	}
 
+	if c.config.Resources.CPUQuota != 0 && c.config.Resources.CPUPeriod != 0 {
+		container.Constraints = hyperstart.Constraints{
+			CPUQuota:  c.config.Resources.CPUQuota,
+			CPUPeriod: c.config.Resources.CPUPeriod,
+		}
+	}
+
 	container.SystemMountsInfo.BindMountDev = c.systemMountsInfo.BindMountDev
 
 	if c.state.Fstype != "" {
--- a/hypervisor.go
+++ b/hypervisor.go
@@ -51,6 +51,9 @@ const (
 	defaultBlockDriver = VirtioSCSI
 )
 
+// In some architectures the maximum number of vCPUs depends on the number of physical cores.
+var defaultMaxQemuVCPUs = maxQemuVCPUs()
+
 // deviceType describes a virtualized device type.
 type deviceType int
 
@@ -81,6 +84,9 @@ const (
 
 	// vhostuserDev is a Vhost-user device type
 	vhostuserDev
+
+	// CPUDevice is CPU device type
+	cpuDev
 )
 
 // Set sets an hypervisor type based on the input string.
@@ -179,6 +185,9 @@ type HypervisorConfig struct {
 	// Pod configuration VMConfig.VCPUs overwrites this.
 	DefaultVCPUs uint32
 
+	//DefaultMaxVCPUs specifies the maximum number of vCPUs for the VM.
+	DefaultMaxVCPUs uint32
+
 	// DefaultMem specifies default memory size in MiB for the VM.
 	// Pod configuration VMConfig.Memory overwrites this.
 	DefaultMemSz uint32
@@ -235,6 +244,10 @@ func (conf *HypervisorConfig) valid() (bool, error) {
 
 	if conf.BlockDeviceDriver == "" {
 		conf.BlockDeviceDriver = defaultBlockDriver
+	}
+
+	if conf.DefaultMaxVCPUs == 0 {
+		conf.DefaultMaxVCPUs = defaultMaxQemuVCPUs
 	}
 
 	return true, nil
--- a/pkg/hyperstart/types.go
+++ b/pkg/hyperstart/types.go
@@ -202,6 +202,16 @@ type SystemMountsInfo struct {
 	DevShmSize int `json:"devShmSize"`
 }
 
+// Constraints describes the constrains for a container
+type Constraints struct {
+	// CPUQuota specifies the total amount of time in microseconds
+	// The number of microseconds per CPUPeriod that the container is guaranteed CPU access
+	CPUQuota int64
+
+	// CPUPeriod specifies the CPU CFS scheduler period of time in microseconds
+	CPUPeriod uint64
+}
+
 // Container describes a container running on a pod.
 type Container struct {
 	ID               string              `json:"id"`
@@ -216,6 +226,7 @@ type Container struct {
 	RestartPolicy    string              `json:"restartPolicy"`
 	Initialize       bool                `json:"initialize"`
 	SystemMountsInfo SystemMountsInfo    `json:"systemMountsInfo"`
+	Constraints      Constraints         `json:"constraints"`
 }
 
 // IPAddress describes an IP address and its network mask.
--- a/pkg/oci/utils.go
+++ b/pkg/oci/utils.go
@@ -454,13 +454,7 @@ func vmConfig(ocispec CompatOCISpec, config RuntimeConfig) (vc.Resources, error)
 			return vc.Resources{}, fmt.Errorf("Invalid OCI cpu period %d", period)
 		}
 
-		// Use some math magic to round up to the nearest whole vCPU
-		// (that is, a partial part of a quota request ends up assigning
-		// a whole vCPU, for instance, a request of 1.5 'cpu quotas'
-		// will give 2 vCPUs).
-		// This also has the side effect that we will always allocate
-		// at least 1 vCPU.
-		resources.VCPUs = uint((uint64(quota) + (period - 1)) / period)
+		resources.VCPUs = vc.ConstraintsToVCPUs(quota, period)
 	}
 
 	return resources, nil
@@ -587,6 +581,14 @@ func ContainerConfig(ocispec CompatOCISpec, bundlePath, cid, console string, det
 		return vc.ContainerConfig{}, err
 	}
 
+	var resources vc.ContainerResources
+	if ocispec.Linux.Resources.CPU != nil &&
+		ocispec.Linux.Resources.CPU.Quota != nil &&
+		ocispec.Linux.Resources.CPU.Period != nil {
+		resources.CPUQuota = *ocispec.Linux.Resources.CPU.Quota
+		resources.CPUPeriod = *ocispec.Linux.Resources.CPU.Period
+	}
+
 	containerConfig := vc.ContainerConfig{
 		ID:             cid,
 		RootFs:         rootfs,
@@ -598,6 +600,7 @@ func ContainerConfig(ocispec CompatOCISpec, bundlePath, cid, console string, det
 		},
 		Mounts:      containerMounts(ocispec),
 		DeviceInfos: deviceInfos,
+		Resources:   resources,
 	}
 
 	cType, err := ocispec.ContainerType()
--- a/qemu.go
+++ b/qemu.go
@@ -35,10 +35,18 @@ type qmpChannel struct {
 	qmp  *govmmQemu.QMP
 }
 
+// CPUDevice represents a CPU device which was hot-added in a running VM
+type CPUDevice struct {
+	// ID is used to identify this CPU in the hypervisor options.
+	ID string
+}
+
 // QemuState keeps Qemu's state
 type QemuState struct {
 	Bridges []Bridge
-	UUID    string
+	// HotpluggedCPUs is the list of CPUs that were hot-added
+	HotpluggedVCPUs []CPUDevice
+	UUID            string
 }
 
 // qemu is an Hypervisor interface implementation for the Linux qemu hypervisor.
@@ -634,8 +642,11 @@ func (q *qemu) hotplugDevice(devInfo interface{}, devType deviceType, op operati
 	case blockDev:
 		drive := devInfo.(Drive)
 		return q.hotplugBlockDevice(drive, op)
+	case cpuDev:
+		vcpus := devInfo.(uint32)
+		return q.hotplugCPUs(vcpus, op)
 	default:
-		return fmt.Errorf("Only hotplug for block devices supported for now, provided device type : %v", devType)
+		return fmt.Errorf("cannot hotplug device: unsupported device type '%v'", devType)
 	}
 }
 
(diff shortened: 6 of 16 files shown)


qemu.go
```diff
@@ -985,6 +996,110 @@ func (q *qemu) hotplugRemoveDevice(devInfo interface{}, devType deviceType) erro
 	return q.pod.storage.storeHypervisorState(q.pod.id, q.state)
 }
 
+func (q *qemu) hotplugCPU(amount uint32, op operation) error {
```
jodh-intel: I was going to say that this function is quite big so might be better split up:

```
if op == addDevice {
    q.addCPU(...)
} else {
    q.removeCPU(...)
}
```
... but then I saw `hotplugAddCPUs()` and `hotplugRemoveCPUs()`, but those two aren't used (yet?)

I assume this is an early branch but we should add some unit tests (even if they only test failure scenarios).

qemu.go
```diff
@@ -37,10 +37,18 @@ type qmpChannel struct {
...
+type CPUDevice struct {
+	// ID is used to identify this CPU in the hypervisor options.
+	ID string
+}
+
 // QemuState keeps Qemu's state
 type QemuState struct {
 	Bridges []Bridge
-	UUID    string
+	/* HotpluggedCPUs is the list of CPUs that were hot-added */
```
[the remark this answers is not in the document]
devimc: lol , I miss you C 😭

qemu.go
devimc: yep, I still need to fix other things

hypervisor.go
```diff
@@ -46,6 +46,8 @@ const (
 	defaultMemSzMiB = 2048
 
 	defaultBridges = 1
+
+	maxVCPUs = 255
```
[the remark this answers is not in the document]
devimc: same as the linux kernel, the limit is 255, but KVM recommends 240

qemu.go
```diff
@@ -985,6 +996,110 @@ func (q *qemu) hotplugRemoveDevice(devInfo interface{}, devType deviceType) erro
...
+		}
+
+		var hotpluggedCPUs uint32
+		for _, hc := range hotpluggableCPUs {
+			// qom-path is the path to the CPU, non-empty means that this CPU is already in use
+			if hc.QOMPath != "" {
+				continue
+			}
+
+			driver := hc.Type
```
mcastelino: @devimc what are you adding here. Core, Socket or Thread. It is not quite clear

qemu.go
```diff
@@ -985,6 +996,110 @@ func (q *qemu) hotplugRemoveDevice(devInfo interface{}, devType deviceType) erro
...
+	// Remove CPUs
+	if amount > uint32(len(q.state.HotpluggedCPUs)) {
+		return fmt.Errorf("Unable to hotunplug %d CPUs, only %d CPUs can be hotunplugged", amount, len(q.state.HotpluggedCPUs))
+	}
+
+	var hotunpluggedCPUs uint32
+	var hotpluggedCPUs []CPUDevice
+	copy(q.state.HotpluggedCPUs, hotpluggedCPUs)
+	for i, huc := range hotpluggedCPUs {
+		if err := q.qmpMonitorCh.qmp.ExecuteDeviceDel(q.qmpMonitorCh.ctx, huc.ID); err != nil {
```
mcastelino: Again which one are you deleting. What is the mapping between this and what got offlined in the guest kernel. How are you co-relating it.

devimc:
@mcastelino @jodh-intel please take a look

pod.go
```diff
@@ -1265,3 +1273,57 @@ func (p *Pod) detachDevices() error {
 
 	return nil
 }
+
+func (p *Pod) updateResources() error {
+	//TODO add support for memory
```
jodh-intel: Could you raise an issue for this and reference it here?

devimc: done, Issue: https://github.com/containers/virtcontainers/issues/578

jodh-intel: Ta - could you add a reference to it in the TODO comment?

devimc:
depends on https://github.com/containers/virtcontainers/pull/581

pod.go
```diff
@@ -1265,3 +1273,58 @@ func (p *Pod) detachDevices() error {
...
+	// Issue: https://github.com/containers/virtcontainers/issues/578
```
devimc: @jodh-intel I added it here, but sure I can move it to the TODO comment

jodh-intel:
@mcastelino - any final comments on this?

qemu.go
```diff
@@ -1007,6 +1017,111 @@ func (q *qemu) hotplugRemoveDevice(devInfo interface{}, devType deviceType) erro
...
+				continue
+			}
+
+			// CPU type, i.e host-x86_64-cpu
+			driver := hc.Type
+			cpuID := fmt.Sprintf("cpu-%d", len(q.state.HotpluggedCPUs))
+			socketID := fmt.Sprintf("%d", hc.Properties.Socket)
+			coreID := fmt.Sprintf("%d", hc.Properties.Core)
+			threadID := fmt.Sprintf("%d", hc.Properties.Thread)
+			// Only sockets are hotplugged, to achieve this, cores and threads MUST BE equal 1 in the Qemu command line
```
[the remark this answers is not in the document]
grahamwhaley: We had a look at this with @mcastelino a few weeks back iirc. I have a feeling what we found was inconclusive - but, yes, previously we have seen and/or been advised that there are performance differences between sockets/cores/cpu's under QEMU.
I already need to re-visit the (iirc related) https://github.com/clearcontainers/packaging/pull/256
Let me see if I can do some work on those tomorrow.

container.go
```diff
@@ -415,6 +425,13 @@ func createContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 	state, err := c.pod.storage.fetchContainerState(c.podID, c.id)
 	if err == nil && state.State != "" {
 		c.state.State = state.State
+		// if we return at this point, the container that is returned
+		// must match with the container in the pod (pod.containers)
+		for i := range pod.containers {
+			if c.id == pod.containers[i].id {
+				pod.containers[i] = c
+			}
+		}
```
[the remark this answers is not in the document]
devimc: the container returned by `fetchContainer` is different to the container in `pod.containers` causing issues updating the container state in `pod.updateResources`

qemu.go
```diff
@@ -1007,6 +1017,111 @@ func (q *qemu) hotplugRemoveDevice(devInfo interface{}, devType deviceType) erro
...
+				// don't fail, let's try with other CPU
+				continue
+			}
+
+			q.state.HotpluggedCPUs = append(q.state.HotpluggedCPUs, CPUDevice{cpuID})
+			hotpluggedCPUs++
+			if hotpluggedCPUs == amount {
+				// All CPUs were hotplugged
+				return nil
+			}
```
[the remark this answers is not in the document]
devimc: lemme find a way to make it simpler

qemu.go
```diff
@@ -566,16 +574,18 @@ func (q *qemu) init(pod *Pod) error {
...
 	}
 
+	// Network IO shows better performance with 1 CPU 1 Socket
 	smp := govmmQemu.SMP{
-		CPUs:    vcpus,
-		Cores:   vcpus,
-		Sockets: defaultSockets,
+		CPUs:    q.config.DefaultVCPUs,
+		Sockets: q.config.DefaultVCPUs,
+		Cores:   defaultCores,
```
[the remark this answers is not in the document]
devimc: sure, I'll do this in other pr

qemu.go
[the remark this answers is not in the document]
devimc: @mcastelino saw better Network IO using 1CPU:1Socket

pkg/hyperstart/types.go
```diff
@@ -201,6 +201,15 @@ type SystemMountsInfo struct {
 	DevShmSize int `json:"devShmSize"`
 }
 
+// Constraints describes the constrains for a container
+type Constraints struct {
+	// CPUQuota specifies the total amount of time in microseconds
```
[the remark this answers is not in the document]
devimc: sure, updating comment

container.go
```diff
@@ -57,6 +58,12 @@ type ContainerStatus struct {
 	Annotations map[string]string
 }
 
+// ContainerResources describes container resources
+type ContainerResources struct {
+	Resources
+	hyperstart.Constraints
```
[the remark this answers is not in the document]
devimc: fixing

qemu.go
```diff
@@ -566,16 +574,18 @@ func (q *qemu) init(pod *Pod) error {
 }
 
 func (q *qemu) setCPUResources(podConfig PodConfig) govmmQemu.SMP {
-	vcpus := q.config.DefaultVCPUs
 	if podConfig.VMConfig.VCPUs > 0 {
-		vcpus = uint32(podConfig.VMConfig.VCPUs)
+		// update the number of vCPUs
```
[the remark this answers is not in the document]
devimc: removed

pkg/hyperstart/types.go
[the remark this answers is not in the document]
devimc: check next line

devimc:
@sameo changes applied, thanks

container.go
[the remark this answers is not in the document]
sameo: That looks like a bug from somewhere else in the vc code, so I'd prefer to fix that first. What's inconsistent? The container state? Or something else?

sameo:
@devimc Also, this is failing the unit tests.

container.go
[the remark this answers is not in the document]
devimc: @sameo https://github.com/containers/virtcontainers/pull/589

devimc:
depends-on https://github.com/containers/virtcontainers/pull/591
https://github.com/containers/virtcontainers/pull/589

sboeuf:
@devimc CI is not passing here. #589 has been solved, and I don't think #591 is the reason why tests are failing, right ?
Could you please make the tests passing on this PR ?

devimc:
@sboeuf let me apply #591 and see if that fixes the CI

devimc:
pssst hey @sboeuf  take a look to the CI

sboeuf:
@devimc thx ! I'll do the review !

api.go
```diff
@@ -440,6 +440,11 @@ func StartContainer(podID, containerID string) (VCContainer, error) {
 		return nil, err
 	}
 
+	// Before starting the container, POD's resources MUST BE updated
+	if err := p.updateResources(); err != nil {
```
sboeuf: Ok let me try to explain here. This call should be called from the function `createContainer()` in `container.go`. This way, it will be called anytime a container needs to be created. Indeed, this function is called after a pod has been created (VM has been started) in order to create new containers if they are provided through the PodConfig, or it is called from `CreateContainer()` here in `api.go`, where we know the VM is already running.
More than factorization, I think it is important to perform the hotplug as soon as possible since the container is really created at the create time, meaning a process is really running inside the agent in case of Kata Containers. And this tend to be the real way to handle this.

pod.go
```diff
@@ -872,6 +876,11 @@ func (p *Pod) start() error {
 		return err
 	}
 
+	// Before starting the container, POD's resources MUST BE updated
+	if err := p.updateResources(); err != nil {
```
sboeuf: Based on my previous explanation, this can be removed from there. (moved to `createContainer()` in `container.go`)

container.go
```diff
@@ -57,6 +57,18 @@ type ContainerStatus struct {
 	Annotations map[string]string
 }
 
+// ContainerResources describes container resources
+type ContainerResources struct {
+	Resources
```
sboeuf: Why do you need this ? The PodConfig already have this info.

hyperstart_agent.go
```diff
@@ -416,6 +416,13 @@ func (h *hyper) startOneContainer(pod Pod, c Container) error {
 		Process: process,
 	}
 
+	if c.config.Resources.VCPUs != 0 {
+		container.Constraints = hyperstart.Constraints{
+			CPUQuota:  c.config.Resources.CPUQuota,
+			CPUPeriod: c.config.Resources.CPUPeriod,
+		}
+	}
+
```
sboeuf: Isn't it great ! Nothing needs to be done for `kata_agent.go` since this is already into the spec :)

pkg/oci/utils.go
```diff
@@ -490,7 +492,7 @@ func PodConfig(ocispec CompatOCISpec, runtime RuntimeConfig, bundlePath, cid, co
 		return vc.PodConfig{}, err
 	}
 
-	resources, err := vmConfig(ocispec, runtime)
```
sboeuf: Please let's keep two separate functions here. One related to the pod (VM) and its global resources, and the other related to each container. 

This way, no need to specify an embedded field `Resources` inside `ContainerConstraints`

pod.go
```diff
@@ -1220,3 +1229,63 @@ func (p *Pod) detachDevices() error {
 
 	return nil
 }
+
+func (p *Pod) updateResources() error {
```
sboeuf: This should take a container as argument, so that it can be called from `createContainer()`.

pod.go
```diff
@@ -1220,3 +1229,63 @@ func (p *Pod) detachDevices() error {
...
+	//TODO add support for memory, Issue: https://github.com/containers/virtcontainers/issues/578
+
+	var vCPUsToRemove, vCPUsToAdd uint
+	// Check containers' resources, hot add/remove as needed
+	for _, container := range p.containers {
```
sboeuf: No need for this loop since this is gonna be called from `createContainer()`.

pod.go
sboeuf: I was expecting this to be called every time we stop a container, am I missing something ?

hypervisor.go
```diff
@@ -467,6 +480,8 @@ type hypervisor interface {
 	addDevice(devInfo interface{}, devType deviceType) error
 	hotplugAddDevice(devInfo interface{}, devType deviceType) error
 	hotplugRemoveDevice(devInfo interface{}, devType deviceType) error
+	hotplugAddCPUs(amount uint32) error
+	hotplugRemoveCPUs(amount uint32) error
```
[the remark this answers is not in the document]
devimc: yep, I think that is not needed, I can use hotplugAddDevice and hotplugRemoveDevice

pod.go
devimc: yep, you're right, adding it

container.go
devimc: to know container's resources, and remove only those resources on stop

sboeuf:
@devimc is that ready for a new review ?

devimc:
@sboeuf changes applied, thanks

container.go
```diff
@@ -592,6 +612,10 @@ func (c *Container) stop() error {
 		return err
 	}
 
+	if err := c.removeResources(); err != nil {
```
sboeuf: By symmetry with `createContainer()`, this call should be placed before `c.detachDevices()`. Indeed, in `createContainer()`, this is called after `c.attachDevices()`

container.go
```diff
@@ -750,3 +774,45 @@ func (c *Container) detachDevices() error {
 
 	return nil
 }
+
+func (c *Container) addResources() error {
+	//TODO add support for memory, Issue: https://github.com/containers/virtcontainers/issues/578
+	if c.config != nil {
```
sboeuf: Please improve indentation with:
```go
if c.config == nil {
    return nil
}
```

container.go
```diff
@@ -750,3 +774,45 @@ func (c *Container) detachDevices() error {
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func (c *Container) removeResources() error {
+	//TODO add support for memory, Issue: https://github.com/containers/virtcontainers/issues/578
+	if c.config != nil {
```
sboeuf: Please improve indentation with:
```go
if c.config == nil {
    return nil
}
```

container.go
```diff
@@ -750,3 +774,45 @@ func (c *Container) detachDevices() error {
...
 }
+
+func (c *Container) addResources() error {
+	//TODO add support for memory, Issue: https://github.com/containers/virtcontainers/issues/578
+	if c.config != nil {
+		cType := ContainerType(c.config.Annotations[vcAnnotations.ContainerTypeKey])
+		if cType == PodSandbox {
+			virtLog.Debugf("Cannot hot add vCPUs to sandbox containers")
+			return nil
+		}
```
sboeuf: I think this check must not be done. The sandbox type will never trigger a call to `createContainer()` unless there is a real container to create, and in that case, we want to hotplug the right amount of resources. By doing this, you are preventing the Docker case from working properly, meaning it will run only on default number of VCPUs.

container.go
```diff
@@ -750,3 +774,45 @@ func (c *Container) detachDevices() error {
...
+
+		vCPUs := c.config.Resources.VCPUs
```
sboeuf: Ok now I understand why you have embedded `Resources` into `ContainerResources`.
But I still think you should not need this since you have cpu quota and period, which should be enough to determine the numbers of vCPUs needed for this container, right ??

container.go
```diff
@@ -750,3 +774,45 @@ func (c *Container) detachDevices() error {
...
+}
+
+func (c *Container) removeResources() error {
+	//TODO add support for memory, Issue: https://github.com/containers/virtcontainers/issues/578
+	if c.config != nil {
+		cType := ContainerType(c.config.Annotations[vcAnnotations.ContainerTypeKey])
+		if cType == PodSandbox {
+			virtLog.Debugf("Cannot hot remove vCPUs from sandbox containers")
+			return nil
+		}
```
sboeuf: Same comment than above, this is not needed, and more than that, this is preventing the Docker case to work as expected.

container.go
```diff
@@ -750,3 +774,45 @@ func (c *Container) detachDevices() error {
...
+
+		vCPUs := c.config.Resources.VCPUs
```
sboeuf: Same comment, we should be able to rely on cpu quota and period I think.

hyperstart_agent.go
```diff
@@ -411,6 +411,13 @@ func (h *hyper) startOneContainer(pod Pod, c *Container) error {
 		Process: process,
 	}
 
+	if c.config.Resources.VCPUs != 0 {
```
sboeuf: You should test `c.config.Resources.CPUQuota` and `c.config.Resources.CPUPeriod` instead of `c.config.Resources.VCPUs`

pkg/oci/utils.go
```diff
@@ -587,6 +587,19 @@ func ContainerConfig(ocispec CompatOCISpec, bundlePath, cid, console string, det
 		return vc.ContainerConfig{}, err
 	}
 
+	var resources vc.ContainerResources
+	resources.Resources, err = vmConfig(ocispec, RuntimeConfig{VMConfig: resources.Resources})
```
sboeuf: This is not needed. The computation translating cpu quota and cpu period into vCPUs should be done inside virtcontainers directly. Those are Linux generic metrics, no need to translate them before to provide them as container resources.

container.go
devimc: currently in docker the CPUs are not hot plugged

container.go
devimc: currently in docker the vCPUs are cold plugged, this code is to prevent adding double resources

container.go
devimc: the problem with that is if the formula changes, let's suppose you start a container with a constrain of 4.5 CPUs, currently 5 vCPUs are added, but if the formula changes in the future to return 6 vCPUs and you update the runtime, when that container ends, the runtime will try to hot-remove 6 vCPUs, not 5. From my point of view the resources assigned to the containers should be saved to avoid this kind of issues.

container.go
devimc: agree

container.go
devimc: fixed

container.go
devimc: fixed

container.go
sboeuf: Well this should not be an issue. The docker case will start the VM with the default CPUs, and when the container (right after) will be created/added, the code should evaluate if it needs more/less CPUs. Why do we need to make a particular case here ?

container.go
sboeuf: The way I see it, you get the VM started with default CPUs for any case (unless the configuration.toml is saying something else), and every container that we start will be defined with cpu quotas and period, which will help us determine if we have to add some CPUs.

container.go
sboeuf: If the formula changes (thing very odd IMO), then we'll change it. But it's easier to get pods being started with a default number of CPUs (unless otherwise noticed by configuration.toml), and add hotplug CPUs for new containers. I would like the Docker case to follow the same logic.

devimc:
@sboeuf changes applied, thanks

depends on https://github.com/clearcontainers/packaging/pull/279

container.go
```diff
@@ -754,3 +775,37 @@ func (c *Container) detachDevices() error {
...
 }
+
+func (c *Container) addResources() error {
+	//TODO add support for memory, Issue: https://github.com/containers/virtcontainers/issues/578
+	if c.config == nil {
+		return nil
+	}
+
+	vCPUs := ConstraintsToVCPUs(c.config.Resources.CPUQuota, c.config.Resources.CPUPeriod)
+	if vCPUs != 0 {
```
sboeuf: We should add an additional check here, looking for the current amount of CPUs on the VM. If the amount available is equal or higher than what is needed by the container (the variable `vCPUs` in this case), then we have nothing to do here.

container.go
```diff
@@ -754,3 +775,37 @@ func (c *Container) detachDevices() error {
...
+}
+
+func (c *Container) removeResources() error {
+	//TODO add support for memory, Issue: https://github.com/containers/virtcontainers/issues/578
+	if c.config == nil {
+		return nil
+	}
+
+	vCPUs := ConstraintsToVCPUs(c.config.Resources.CPUQuota, c.config.Resources.CPUPeriod)
+	if vCPUs != 0 {
```
sboeuf: This cannot be as simple as it is. First, we should check that removing the numbers `vCPUs` here won't fail because we're trying to remove the coldplugged default CPUs (in case we didn't need to hotplug when `addResources` has been called). And second, we should have saved the highest number of CPUs needed by any container on this pod, in order to prevent the removal of CPUs still needed by this container.

devimc:
@sboeuf I think your design is different than mine, you want to share vCPUs between containers while I don't. Correct me if I''m wrong, but in your design if we have 1 container with a constraint of 2 vCPUs, the VM should have 2 vCPUs, but if a new container is created with a constrain of 3 vCPUs, then only 1 vCPUs will be added, the idea is good but is not the correct solution. In order to don't have containers fighting for the CPU usage, each container should have assigned the number of vCPUs that it needs, we can not exceed the maximum number of actual physical cores because docker and kubernetes validates that before starting the container, if `quota/period` is greater than the actual number of physical cores the container won't start, also in `qemu.go` I have a validation to do not exceed the maximum number of vCPUs, in that way when the container is destroyed only its resources are hot removed (if it has any), that means we will not remove CPUs used by other containers neither the runtime will try to remove cold plugged vCPUs.

devimc:
@sboeuf 
> Do you realize that we're always going to end up with number_of_CPUs_on_the_VM = coldplugged_CPUs + (container_i * number_of_CPUs_for_container_i) ?

Yes, when a CPU constrain is not specified for a container, then it will use the CPUs cold plugged.

> How are we going to handle the constraint of the container process inside the VM so that we won't end up using only the first CPUs (0,1,2 for instance for a second container that needs 3 CPUs, but the first container needed 2 CPUs) ?

Before creating a container, the runtime check if a CPU constraint was specified, if so, then N cpus are hot added and the agent places the container inside a CPU cgroup, containers without CPU constraint have less priority than containers with CPUs constraint. For example, if `default_vcpus` is 1 and this pod configuration is used:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: cpu-demo
  namespace: cpu
spec:
  containers:
  - name: container1
    image: vish/stress
    args:
    - -cpus
    - "4"
  - name: container2
    image: vish/stress
    resources:
      limits:
        cpu: "4"
    args:
    - -cpus
    - "4"
```
both containers `container1` and `container2` will try to consume 400% of CPU. When `container1` is created the VM only have 1 vCPU, hence `container1` uses 100% of CPU. When `container2` is created, 4 vCPUs are hot added, `container2` is placed inside a CPU cgroup, and it consumes 400% of CPU while `container1` only 100%

sboeuf:
@devimc Ok this is pretty clear, but I thought we had to specify which CPU (0, 1, 2, ..., N) we're gonna use for each container. If the kernel handles the workload on its own, then that's fine to only specify the number of CPU for each process and we don't have to bother with this. Is that right ?

Now, about the coldplugged CPU, I think we can do better here. In case all the containers running will have some CPU constraints, the number of CPUs coldplugged will never be used. We're wasting resources here.
So I'd like that we take into account the number of coldplugged CPUs before we start hotplugging more CPUs. Also, I think that we should handle the non-constrained containers as if they had some CPUs to hotplug. The only way to do that is to define a default amount of CPUs that we expect acceptable to run a non-constrained container.
Does that make sense ?

devimc:
@sboeuf 
> I thought we had to specify which CPU (0, 1, 2, ..., N) we're gonna use for each container. 

when I started with this, I thought the same, but that is not needed because the kernel automatically runs the workload in N vCPUs. Also we cannot decide what vCPUs the container have to use, the user can decide this `docker --cpuset-cpus`(in another PR I'll add support for this option).

> If the kernel handles the workload on its own, then that's fine to only specify the number of CPU for each process and we don't have to bother with this. Is that right ?

that's correct.

> Now, about the coldplugged CPU, I think we can do better here. In case all the containers running will have some CPU constraints, the number of CPUs coldplugged will never be used. We're wasting resources here.

It depends, to deal with this limitation, in https://github.com/clearcontainers/runtime/pull/1041 I'm changing `default_vcpus` to 1, this vCPU can be used for other processes, remember that our containers are not running alone in the VM, we have other process running, like the kernel, systemd(journal, etc) and the agent.
I agree with you, if the user changes `default_vcpus` to a number greater than 1, then we will be wasting resources, for that reason I'll document this in https://github.com/clearcontainers/runtime/pull/1041

> So I'd like that we take into account the number of coldplugged CPUs before we start hotplugging more CPUs. Also, 

and we do, these coldplugged vCPUs are used by containers without CPU constraints, it's up to the user decides the resources available for the containers without a CPU constraint.

> I think that we should handle the non-constrained containers as if they had some CPUs to hotplug. The only way to do that is to define a default amount of CPUs that we expect acceptable to run a non-constrained container.

In runC, if containers don't have a CPU constraint, they will try to consume all the resources needed to complete theirs task, they will compete for the resources. Cold plugged vCPUs are used and shared between  the containers without CPU constraint, if the user want to give more priority to one container over these resources, He/She can use `docker --cpu-shares` (I'll add support for this option in other PR)

sboeuf:
Following our discussion on IRC with @devimc, the change looks good and can now be merged as long as we document very well all the behavior we expect with unconstrained vs constrained containers.
LGTM

sboeuf:
@devimc please take a look why this is failing so that we can move forward and get this merged ;)

PR #520, opened 2017-12-16
Network refactoring/cleanup

egernst:
Some of the function naming and process around CNI's add() were overly complicated and/or obfuscated.
    
    Initially, the CNI "add" function would create a virtual device which
    would then be used to store the results obtained from calling the CNI
    plugin.  This patch removes that, and instead keeps the netinfo results
    for application to the endpoint device when we identify it post-scan.
    
    Renamed addVirtInterfaces and deleteVirtInterfaces functions.
    add is now invokePlugins, since this is what is carried out, and
    the "virt" is removed from delete since there is nothing specific
    to virtual in this case (it just invokes delete command to the plugin).
    The function prototype for invokePlugins was also modified to return
    the resulting netinfo from the CNI call, allowing the DNS information
    to be used when updating the endpoints after the network namespace
    is scanned.
    
    The logic has been updated to reflect that a pod will only have a single
    network namespace.  This simplified the code in a couple of locations,
    since we know the initial numInterfaces is always '1' - no need to loop.
    This single interfaces is described as 'CniPrimaryInterface' in the
    sources.

-Also, fixed bug in how uniqueID was used for createVirtualNetworkEndpoint.
-Rewrote network scan portion to avoid looping twice unnecessarily.

### DIFF
--- a/cni.go
+++ b/cni.go
@@ -70,7 +70,7 @@ func convertCNIResult(cniResult cniTypes.Result) (NetworkInfo, error) {
 	}
 }
 
-func (n *cni) invokePlugins(pod Pod, networkNS *NetworkNamespace) (*NetworkInfo, error) {
+func (n *cni) invokePluginsAdd(pod Pod, networkNS *NetworkNamespace) (*NetworkInfo, error) {
 	netPlugin, err := cniPlugin.NewNetworkPlugin()
 	if err != nil {
 		return nil, err
@@ -97,7 +97,7 @@ func (n *cni) invokePlugins(pod Pod, networkNS *NetworkNamespace) (*NetworkInfo,
 	return &netInfo, nil
 }
 
-func (n *cni) deleteInterfaces(pod Pod, networkNS NetworkNamespace) error {
+func (n *cni) invokePluginsDelete(pod Pod, networkNS NetworkNamespace) error {
 	netPlugin, err := cniPlugin.NewNetworkPlugin()
 	if err != nil {
 		return err
@@ -148,7 +148,7 @@ func (n *cni) add(pod Pod, config NetworkConfig, netNsPath string, netNsCreated 
 		NetNsCreated: netNsCreated,
 	}
 
-	netInfo, err := n.invokePlugins(pod, &networkNS)
+	netInfo, err := n.invokePluginsAdd(pod, &networkNS)
 	if err != nil {
 		return NetworkNamespace{}, err
 	}
@@ -171,7 +171,7 @@ func (n *cni) remove(pod Pod, networkNS NetworkNamespace) error {
 		return err
 	}
 
-	if err := n.deleteInterfaces(pod, networkNS); err != nil {
+	if err := n.invokePluginsDelete(pod, networkNS); err != nil {
 		return err
 	}
 
--- a/cni.go
+++ b/cni.go
@@ -26,6 +26,10 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// CniPrimaryInterface Name chosen for the primary interface
+// If CNI ever support multiple primary interfaces this should be revisited
+const CniPrimaryInterface = "eth0"
+
 // cni is a network implementation for the CNI plugin.
 type cni struct{}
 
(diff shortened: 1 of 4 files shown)


egernst:
@mcastelino

egernst:
@sboeuf @jcvenegas @amshinde  -- can ya'll take a look?

sboeuf:
@egernst yep that's on my list :)

jcvenegas:
@egernst  sure !

amshinde:
lgtm

cni.go
```diff
@@ -66,84 +70,63 @@ func convertCNIResult(cniResult cniTypes.Result) (NetworkInfo, error) {
...
+	// We do not care about this for now but
+	// If present, the CNI DNS result has to be updated in resolv.conf
+	// if the kubelet has not supplied it already
+	n.Logger().Infof("AddNetwork results %s", result.String())
+
+	return &netInfo, nil
 }
 
-func (n *cni) deleteVirtInterfaces(pod Pod, networkNS NetworkNamespace) error {
+func (n *cni) deleteInterfaces(pod Pod, networkNS NetworkNamespace) error {
```
sboeuf: Now that you have replaced `addVirtInterfaces()` with `invokePlugins()`, I think it would be better to add a parameter to `invokePlugins()` so that you can do both `add` or `remove` action from the same function.
Or I would prefer to have `invokePluginsAdd()` and `invokePluginsDelete()` for more consistency.

network.go
```diff
@@ -184,6 +184,8 @@ func networkLogger() *logrus.Entry {
 
 // Attach for virtual endpoint bridges the network pair and adds the
 // tap interface of the network pair to the hypervisor.
+// TODO: This should be split to multiple end point types with
+// therr own attach methods
```
sboeuf: But this is one implementation of the `Endpoint` interface already, I don't get the comment here.

network.go
```diff
@@ -1014,13 +1014,12 @@ func deleteNetNS(netNSPath string, mounted bool) error {
 	return nil
 }
 
-func createVirtualNetworkEndpoint(idx int, uniqueID string, ifName string) (*VirtualEndpoint, error) {
```
sboeuf: You're modifying the behavior here, meaning we will end up with a new uniqueID for every endpoint. What you're doing is harmless but I'd like to understand where was the bug ?
Also, if we go with this solution, then I don't think we need to concatenate the unique ID with the actual ID.

network.go
```diff
@@ -1128,16 +1099,29 @@ func createEndpointsFromScan(networkNSPath string) ([]Endpoint, error) {
...
 
-	netInfoList, err := networkInfoListFromNetworkScan(netlinkHandle)
+	linkList, err := netlinkHandle.LinkList()
 	if err != nil {
 		return []Endpoint{}, err
 	}
 
 	idx := 0
-	for _, netInfo := range netInfoList {
+	for _, link := range linkList {
```
sboeuf: This looks good, thanks for the optimization here!
We might need to break `createEndpointsFromScan()` into subfunctions because it seems to me that we're close to the `gocyclo 15` limit.

network.go
egernst: I wrote some description in the commit message.  Prior, the 'unique id' logic was being done before calling the create function.  Since this wasn't being used anywhere else, it makes more sense to do the exact same logic, only inside the function itself.

The bug that existed in the past was that the unique ID wasn't necessarily being created for each endpoint, as was intended (per discussion with @amshinde).  It was being called once, and then multiple virtual endpoints would potentially use this same "unique" ID.

network.go
egernst: I'll clean up the typo above, @sboeuf.  For content, I think the intention of Manohar on this comment is that for a "virtual endpoint", there can by multiple ways that you'd end up attaching them to the hypervisor itself (bridge, TC, macvtap), and that this should be explicitly handled here.

cni.go
egernst: Yeah, agreed @sboeuf

network.go
sboeuf: Yes, we were using one uniqueID per network namespace but the ID seems to me like it was unique since we were concatenating the uniqueID with the endpoint ID. I'll talk with @amshinde to understand the bug.

network.go
egernst: Looks like its at 11 with this change.  There'll be more changes later on which should simplify it (I think one function for determining endpoint type - whether it be physical, virtual or vhost-user based), a case instead of if/else..

sboeuf: Ok !

cni.go
sboeuf: Thx

network.go
sboeuf: Ok I understand. This will be for a follow up PR, no need to discuss further on this point !

network.go
sboeuf: Talking with @amshinde , it seems there was no bug, but a feeling there might be a bug :)
Ok anyway, I think we can keep this change but to go one step further, I think we should should concatenate with the endpoint ID, there is no point doing this anymore.

egernst: Okay - - repushed.  Saw a build failure, though I don't see this as related to my changes.  Repushed in the hopes that it was a transient issue (I can build locally, and the same changes were fine on prior pushes) . Will try to check in on this later.  Thanks for the review, @sboeuf

sboeuf:
LGTM

sboeuf:
@egernst the failure is not related to your changes. Virtcontainers needed https://github.com/clearcontainers/runtime/pull/870 to be merged.
Could you re-push in order to re-trigger the whole CI ?
Also make sure you rebase on the latest `origin/master`.

egernst:
@sboeuf:  Done.

sboeuf:
@egernst thanks, I'll merge it now !

PR #522, opened 2017-12-18
agent: Add kata exec, stopContainer and killContainer.

### LINKED ISSUE #521: Add missing kata containers agent calls
Add basic kata agent support for the following:

- `exec()`
- `stopContainer()`
- `killContainer()`

jodh-intel:
Basic kata agent implementations for exec(), stopContainer() and
killContainer().

Fixes #521.


sameo:
LGTM

devimc:
lgtm

kata_agent.go
```diff
@@ -134,8 +135,37 @@ func (k *kataAgent) createPod(pod *Pod) error {
 	return pod.hypervisor.addDevice(sharedVolume, fsDev)
 }
 
+func cmdToKataProcess(cmd Cmd) *grpc.Process {
+	return &grpc.Process{
+		Terminal: cmd.Interactive,
+		User: grpc.User{
+			Username: cmd.User,
```
jodh-intel: I've just looked at this again and realised that it's incorrect: `grpc.User.Username` represents `ocispec.Process.User.Username`, which is [Windows-only](https://github.com/opencontainers/runtime-spec/blob/master/config.md#windows-user), meaning that `Username` will be silently ignored I think. I can fix this simply enough but stepping back...

- The OCI runtime spec encodes the UID/GID details as `int` values.
- `vc.ContainerConfig()` converts those values into a `vc.Cmd`, which encodes UID/GID as strings.
- The runtime currently passes `cmd.User` from `ocispec.Process.User.Username` to `vc.EnterContainer()`. This is very confusing given that:
  - `Username` is reserved for Windows only.
  - `runtime.generateExecParams()` takes the `exec --user $user` and sets `Username` (Windows-only) which is then assigned as `vc.Cmd.User = `Username`.
- The kata agent (`oci.proto`) specifies UID/GID as `uint32` values.
- The hyperstart agent specifies UID/GID as string values.

Hence, I think we need to:

- Stop the runtime from using `Username`. I'm *guessing* the reason we use it currently is as a convenient way to pass the string UID value to `vc.Cmd`?
- Change `vc.Cmd` to specify the UID/GID as a `uint32` to avoid all the conversions for Kata.
- Change `buildHyperContainerProcess()` so that it convers the (now) int UID/GID values to strings.

That will make kata the "fast path" and we can convert the numeric into a string purely for hyperstart.

@sameo, @sboeuf - thoughts?

jodh-intel:
Branch updated fairly significantly, so this will need a re-review.

sameo:
still LGTM

jodh-intel:
Hi @sameo - this builds locally for me so I suspect the CI fails relate to vendoring changes.

jodh-intel:
... that still need to be applied to the runtime, used in testing vc.

jodh-intel:
@sameo - I'm struggling to convince `dep` to use the `sameo/topic/kata` branch rather than `master` for the `Depends-on:` runtime test re-vendor to allow this PR to pass the CI checks.

kata_agent.go
jodh-intel: Annoyingly, afaics we can't switch `Cmd` to use `uint32` for `UID` and `GID` as that will break invoking the runtime standalone using `--user=` (see https://github.com/clearcontainers/runtime/pull/873).

The only alternative I can think of right now (aside from rewriting everything in C [again!] and using a union :-) is to keep have the `UID` and `GID` as `uint32`'s but also add a `UserOverride string` field to `Cmd` that would take priority over `UID` and `GID`, but that's pretty horrid.

jodh-intel:
Removed the commit that converted the user+group fields in `Cmd` from `string` to `uint32` and raised #534 to explain why we cannot do that.

sboeuf:
@jodh-intel Could you please rework this PR according to the changes introduced here: https://github.com/kata-containers/agent/pull/83
Also, I think you should open a PR against `master` so that we can review it and merge it into virtcontainers.

jodh-intel:
Closing - replaced by #538.

PR #523, opened 2017-12-19
agent: Check for -v /dev:/dev case

amshinde:
Introduce a new variable "BindMountDev" to indicate the case where
/dev is passed as a volume on the docker command line.
This case needs to be handled by the agent.

### DIFF
--- a/container.go
+++ b/container.go
@@ -99,6 +99,16 @@ func (c *ContainerConfig) valid() bool {
 	return true
 }
 
+// SystemMountsInfo describes additional information for system mounts that the agent
+// needs to handle
+type SystemMountsInfo struct {
+	// Indicates if /dev has been passed as a bind mount for the host /dev
+	BindMountDev bool
+
+	// Size of /dev/shm assigned on the host.
+	DevShmSize uint
+}
+
 // Container is composed of a set of containers and a runtime environment.
 // A Container can be created, deleted, started, stopped, listed, entered, paused and restored.
 type Container struct {
@@ -122,6 +132,8 @@ type Container struct {
 	mounts []Mount
 
 	devices []Device
+
+	systemMountsInfo SystemMountsInfo
 }
 
 // ID returns the container identifier string.
@@ -482,6 +494,19 @@ func (c *Container) fetchState(cmd string) (State, error) {
 	return state, nil
 }
 
+func (c *Container) getSystemMountInfo() {
+	// check if /dev needs to be bind mounted from host /dev
+	for _, m := range c.mounts {
+		if m.Source == "/dev" && m.Destination == "/dev" && m.Type == "bind" {
+			c.systemMountsInfo.BindMountDev = true
+		} else {
+			c.systemMountsInfo.BindMountDev = false
+		}
+	}
+
+	// TODO Deduce /dev/shm size. See https://github.com/clearcontainers/runtime/issues/138
+}
+
 func (c *Container) start() error {
 	state, err := c.fetchState("start")
 	if err != nil {
@@ -518,6 +543,10 @@ func (c *Container) start() error {
 	if err := c.attachDevices(); err != nil {
 		return err
 	}
+
+	// Deduce additional system mount info that should be handled by the agent
+	// inside the VM
+	c.getSystemMountInfo()
 
 	if err = c.pod.agent.startContainer(*(c.pod), *c); err != nil {
 		c.Logger().WithError(err).Error("Failed to start container")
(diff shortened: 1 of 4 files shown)


amshinde:
cc @sboeuf

pkg/hyperstart/types.go
```diff
@@ -186,6 +186,7 @@ type Container struct {
 	Process       *Process            `json:"process"`
 	RestartPolicy string              `json:"restartPolicy"`
 	Initialize    bool                `json:"initialize"`
+	BindMountDev  bool                `json:"bindMountDev"`
```
sboeuf: This should be part of a more generic structure like:
```go
type IgnoredMounts struct {
        Dev bool
        Proc bool
        ...
}
```

Make this structure the way you want, but the point is that I don't want to see `Container` structure growing too much. And this is my feeling if we add a new field every time we have to handle a specific case.

amshinde: @sboeuf How about 
``` 
type IgnoredMountsInfo struct {
  BindMountDev bool    // Flag which says if -v /dev:/dev is true (just Dev does not make it clear at all to me)
  DevShmSize int   // For handing /dev/shm size in the future, currently hardcoded
}
```
I dont know what we will be doing about proc, or if we can do anything about it, so will leave that out for now, we can add those as we start handling them.

sboeuf: Yes this looks good. I was only using `proc` as an example, but I was not expecting you to use it.
The case with `Shm` looks good and illustrates the fact that we can expect several special handling cases like this to come in the future.

amshinde: @sboeuf I have pushed changes. Have a look.

pkg/hyperstart/types.go
```diff
@@ -173,19 +173,29 @@ type Process struct {
...
+	Fstype            string              `json:"fstype,omitempty"`
+	Image             string              `json:"image"`
+	Addr              string              `json:"addr,omitempty"`
+	Volumes           []*VolumeDescriptor `json:"volumes,omitempty"`
+	Fsmap             []*FsmapDescriptor  `json:"fsmap,omitempty"`
+	Sysctl            map[string]string   `json:"sysctl,omitempty"`
+	Process           *Process            `json:"process"`
+	RestartPolicy     string              `json:"restartPolicy"`
+	Initialize        bool                `json:"initialize"`
+	BindMountDev      bool                `json:"bindMountDev"`
```
sboeuf: You forgot to remove this.

amshinde: Oops! Fixed

amshinde:
@sboeuf  Agree. Deducing information about the system/ignored mounts should be done in container.go , now that we know of atleast two cases, especially deducing the shm size to be done in the future. The dev bind mount was pretty trivial enough, so I had left that to the agent.
I'll add the changes.

amshinde:
@sboeuf Changes pushed

container.go
```diff
@@ -99,6 +99,16 @@ func (c *ContainerConfig) valid() bool {
 	return true
 }
 
+// SystemMountsInfo describes additional information for system mounts that the agent
+// needs to handle
+type SystemMountsInfo struct {
+	// Indicates if /dev has been passed as a bind mount for the host /dev
+	BindMountDev bool
+
+	// Size of /dev/shm assigned by docker.
```
sboeuf: No need to mention Docker from this comment. I would say `Size of /dev/shm assigned by the container specification.`

container.go
```diff
@@ -482,6 +494,17 @@ func (c *Container) fetchState(cmd string) (State, error) {
 	return state, nil
 }
 
+func (c *Container) getSystemMountInfo() {
+	// check if /dev needs to be bind mounted from host /dev
+	for _, m := range c.mounts {
+		if m.Source == "/dev" && m.Destination == "/dev" && m.Type == "bind" {
+			c.systemMountsInfo.BindMountDev = true
```
sboeuf: Don't you think we should maybe remove the mount from the list after we identified it as a special case ?

amshinde: Our hyperstart agent anyways ignores /dev in any case, if /dev is passed as a bind mount or not. I think we should leave it to the agent and not do any removal here.

sboeuf: Yes I agree, let's not remove too much info.

container.go
amshinde: fixed

sboeuf:
LGTM

Issue #524, opened 2017-12-19
Rework of agent interface


sboeuf:
A light rework of this interface is needed in order to make future proxy changes easier.
We need to add two new functions to the agent interface:
```go
type agent interface {
        ...
        // vmURL returns the agent URL exposed by the hypervisor. This URL is
	// directly connected to the agent running inside the virtual machine.
	vmURL() (string, error)

	// proxyURL sets the URL virtcontainers should connect in order to
	// communicate with the agent. This URL can be either the proxy URL
	// if it actually needs to go through a proxy, or it can be the VM
	// URL in case no proxy is needed. Both ways, this has to be
	// transparent for the agent implementation.
	proxyURL(url string) error
        ...
}
```

Issue #516, opened 2017-12-13
proxy: Implement a no-proxy


sboeuf:
We need to implement the generic case where no proxy is needed to interact with the agent. This case happens for example when our host system is able to support VSOCK, meaning that our VM will handle several connections without need of a proxy to handle the multiplexing.
This implementation should be very simple and should only return the VM socket every time we expect a proxyURL to be returned.

PR #525, opened 2017-12-19
proxy: Introduce no_proxy and kata_proxy implementations

### LINKED ISSUE #515: proxy: Implement Kata Containers proxy
We need to implement the Kata Containers proxy for our proxy interface. This will be able to launch and interact with the Kata Containers proxy component. This needs to be carefully implemented so that we don't break the existing implementation of `cc_proxy.go` by changing some global calls into the proxy interface.

sboeuf:
This PR implements proxies for Kata Containers cases with and without proxy.
In order to achieve this, it slightly modifies the agent interface by adding two new functions.

Fixes #515 #516 #524

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -133,6 +133,22 @@ type agent interface {
 	// to handle all other Agent interface methods.
 	init(pod *Pod, config interface{}) error
 
+	// vmURL returns the agent URL exposed by the hypervisor. This URL has
+	// been previously built from the agent implementation and provided to
+	// the hypervisor through a specific hypervisor interface function. It
+	// represents the entry point to connect to the agent through the VM.
+	// This function is particularly useful in case there is no proxy
+	// needed, meaning the proxy implementation will have to provide this
+	// URL instead of a real proxy URL.
+	vmURL() (string, error)
+
+	// setProxyURL sets the URL virtcontainers should connect in order to
+	// communicate with the agent. This URL can be either the proxy URL
+	// if it actually needs to go through a proxy, or it can be the VM
+	// URL in case no proxy is needed. Both ways, this has to be
+	// transparent for the agent implementation.
+	setProxyURL(url string) error
+
 	// capabilities should return a structure that specifies the capabilities
 	// supported by the agent.
 	capabilities() capabilities
--- a/kata_proxy.go
+++ b/kata_proxy.go
@@ -1,0 +1,98 @@
+//
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package virtcontainers
+
+import (
+	"fmt"
+	"os/exec"
+	"path/filepath"
+)
+
+// This is the Kata Containers implementation of the proxy interface.
+// This is pretty simple since it provides the same interface to both
+// runtime and shim as if they were talking directly to the agent.
+type kataProxy struct {
+	proxyURL string
+}
+
+// start is kataProxy start implementation for proxy interface.
+func (p *kataProxy) start(pod Pod) (int, string, error) {
+	config, err := newProxyConfig(pod.config)
+	if err != nil {
+		return -1, "", err
+	}
+
+	// construct the socket path the proxy instance will use
+	socketPath := filepath.Join(runStoragePath, pod.id, "kata_proxy.sock")
+	proxyURL := fmt.Sprintf("unix://%s", socketPath)
+
+	vmURL, err := pod.agent.vmURL()
+	if err != nil {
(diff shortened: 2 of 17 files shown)


sboeuf:
cc @sameo @jodh-intel

no_proxy.go
```diff
@@ -0,0 +1,70 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package virtcontainers
+
+type noProxy struct {
```
jodh-intel: Now that this file has been introduced, I think it might be a good idea to add a description in this file explaining what `no_proxy.go` is for.

The name does admittedly give a clue, but `noop_proxy.go` and `no_proxy.go` are pretty similar, as are the file contents apart from the crucial `start()` call below.

It might even be worth expanding on the brief comments in `noop_proxy.go` to add a descriptive header explaining that this is a dummy proxy whereas `no_proxy.go` provides the (mostly) empty proxy implementation required by vc, but which is specifically for systems that don't actually require a proxy process (namely kata containers environments running on systems whose host kernel supports vsock).

agent.go
```diff
@@ -133,6 +133,17 @@ type agent interface {
...
+	// vmURL returns the agent URL exposed by the hypervisor. This URL is
+	// directly connected to the agent running inside the virtual machine.
+	vmURL() (string, error)
+
+	// proxyURL sets the URL virtcontainers should connect in order to
+	// communicate with the agent. This URL can be either the proxy URL
+	// if it actually needs to go through a proxy, or it can be the VM
+	// URL in case no proxy is needed. Both ways, this has to be
+	// transparent for the agent implementation.
+	proxyURL(url string) error
```
jodh-intel: This should be `setProxyURL()` to be idiomatic go I think.

sboeuf: Oh yeah you're right, otherwise this looks like a getter.

no_proxy.go
sboeuf: @jodh-intel I fully agree! I'll add some more descriptions and more unit tests.

agent.go
sboeuf: Done

no_proxy.go
sboeuf: Done

sboeuf:
@jodh-intel @sameo PR updated.

jodh-intel:
lgtm but let's wait to see if @sameo has any comments.

kata_proxy.go
```diff
@@ -0,0 +1,98 @@
...
+	config, err := newProxyConfig(pod.config)
+	if err != nil {
+		return -1, "", err
+	}
+
+	// construct the socket path the proxy instance will use
+	socketPath := filepath.Join(runStoragePath, pod.id, "kata_proxy.sock")
+	proxyURL := fmt.Sprintf("unix://%s", socketPath)
+
+	vmURL, err := pod.agent.vmURL()
```
sameo: So here we assume `agent.init()` has been called already, right ?

kata_proxy.go
```diff
@@ -0,0 +1,98 @@
...
+func (p *kataProxy) register(pod Pod) ([]ProxyInfo, string, error) {
+	var proxyInfos []ProxyInfo
+
+	for i := 0; i < len(pod.containers); i++ {
+		proxyInfo := ProxyInfo{}
+
+		proxyInfos = append(proxyInfos, proxyInfo)
+	}
+
+	return proxyInfos, p.proxyURL, nil
```
sameo: Do we really need to return an array of empty `proxyInfo` instances?

agent.go
```diff
@@ -133,6 +133,17 @@ type agent interface {
 	// to handle all other Agent interface methods.
 	init(pod *Pod, config interface{}) error
 
+	// vmURL returns the agent URL exposed by the hypervisor. This URL is
+	// directly connected to the agent running inside the virtual machine.
```
sameo: Could we please add some more information here: In particular, could we clearly explain that the VM URL needs to be built and passed to the hypervisor by the agent.

sameo:
A few comments to be answered/addressed.

kata_proxy.go
sboeuf: Yes. `proxy.start()` is called here: https://github.com/containers/virtcontainers/blob/master/pod.go#L768 from a higher level call here: https://github.com/containers/virtcontainers/blob/master/api.go#L90
And `agent.init()` is called here: https://github.com/containers/virtcontainers/blob/master/pod.go#L653 from https://github.com/containers/virtcontainers/blob/master/pod.go#L561 which is called from here: https://github.com/containers/virtcontainers/blob/master/api.go#L46

Now the question could be, do we want a proper check to make sure the agent has been initialized ? In that case, we might want to introduce `isInitialized()` function to the agent interface.

kata_proxy.go
sboeuf: Yes, because of this https://github.com/containers/virtcontainers/blob/master/pod.go#L798.

agent.go
sboeuf: Yes, I'll try to be more explicit and verbose about that.

sboeuf:
@sameo PR updated.

kata_proxy.go
sameo: Ok, I guess we'll fix that with the 2.0 API...

sameo:
LGTM

Issue #527, opened 2017-12-20
Allow change default networking model at pod level


jcvenegas:
The default model is macvtap but it can not be changed by vc api. We need to be able to configure it 

This will allow a runtime be able to choose the networking model at a per pod level.

PR #528, opened 2017-12-20
network: pod: Get default internetwork model from pod configuration

jcvenegas:
This patch change the default internetwork model based in the pod
configuration.

Fixes: #527

Signed-off-by: Jose Carlos Venegas Munoz <jose.carlos.venegas.munoz@intel.com>

### DIFF
--- a/network.go
+++ b/network.go
@@ -117,8 +117,9 @@ type NetworkInterfacePair struct {
 
 // NetworkConfig is the network configuration related to a network.
 type NetworkConfig struct {
-	NetNSPath     string
-	NumInterfaces int
+	NetNSPath         string
+	NumInterfaces     int
+	InterworkingModel NetInterworkingModel
 }
 
 // Endpoint represents a physical or virtual network interface.
@@ -1103,7 +1104,7 @@ func deleteNetNS(netNSPath string, mounted bool) error {
 	return nil
 }
 
-func createVirtualNetworkEndpoint(idx int, ifName string) (*VirtualEndpoint, error) {
+func createVirtualNetworkEndpoint(idx int, ifName string, interworkingModel NetInterworkingModel) (*VirtualEndpoint, error) {
 	if idx < 0 {
 		return &VirtualEndpoint{}, fmt.Errorf("invalid network endpoint index: %d", idx)
 	}
@@ -1126,6 +1127,7 @@ func createVirtualNetworkEndpoint(idx int, ifName string) (*VirtualEndpoint, err
 			TAPIface: NetworkInterface{
 				Name: fmt.Sprintf("tap%d", idx),
 			},
+			NetInterworkingModel: interworkingModel,
 		},
 		EndpointType: VirtualEndpointType,
 	}
@@ -1158,7 +1160,7 @@ func networkInfoFromLink(handle *netlink.Handle, link netlink.Link) (NetworkInfo
 	}, nil
 }
 
-func createEndpointsFromScan(networkNSPath string) ([]Endpoint, error) {
+func createEndpointsFromScan(networkNSPath string, config NetworkConfig) ([]Endpoint, error) {
 	var endpoints []Endpoint
 
 	netnsHandle, err := netns.GetFromPath(networkNSPath)
@@ -1228,7 +1230,7 @@ func createEndpointsFromScan(networkNSPath string) ([]Endpoint, error) {
 					cnmLogger().WithField("interface", netInfo.Iface.Name).Info("VhostUser network interface found")
 					endpoint, err = createVhostUserEndpoint(netInfo, socketPath)
 				} else {
-					endpoint, err = createVirtualNetworkEndpoint(idx, netInfo.Iface.Name)
+					endpoint, err = createVirtualNetworkEndpoint(idx, netInfo.Iface.Name, config.InterworkingModel)
 				}
 			}
 
--- a/network.go
+++ b/network.go
@@ -1137,22 +1137,6 @@ func createVirtualNetworkEndpoint(idx int, ifName string) (*VirtualEndpoint, err
 	return endpoint, nil
 }
 
-func createNetworkEndpoints(numOfEndpoints int) (endpoints []Endpoint, err error) {
-	if numOfEndpoints < 1 {
-		return endpoints, fmt.Errorf("Invalid number of network endpoints")
-	}
-
-	for i := 0; i < numOfEndpoints; i++ {
-		endpoint, err := createVirtualNetworkEndpoint(i, "")
-		if err != nil {
-			return nil, err
-		}
-		endpoints = append(endpoints, endpoint)
-	}
-
-	return endpoints, nil
-}
-
 func networkInfoFromLink(handle *netlink.Handle, link netlink.Link) (NetworkInfo, error) {
 	addrs, err := handle.AddrList(link, netlink.FAMILY_ALL)
 	if err != nil {
--- a/network.go
+++ b/network.go
@@ -675,7 +675,7 @@ func getLinkByName(netHandle *netlink.Handle, name string, expectedLink netlink.
 
 // The endpoint type should dictate how the connection needs to be made
 func xconnectVMNetwork(netPair *NetworkInterfacePair, connect bool) error {
-	switch DefaultNetInterworkingModel {
+	switch netPair.NetInterworkingModel {
 	case ModelBridged:
 		netPair.NetInterworkingModel = ModelBridged
 		if connect {
@@ -691,7 +691,7 @@ func xconnectVMNetwork(netPair *NetworkInterfacePair, connect bool) error {
 	case ModelEnlightened:
 		return fmt.Errorf("Unsupported networking model")
 	default:
-		return fmt.Errorf("Invalid networking model")
+		return fmt.Errorf("Invalid internetworking model: '%s'", netPair.NetInterworkingModel)
 	}
 }
 
--- a/network.go
+++ b/network.go
@@ -71,30 +71,21 @@ func (n NetInterworkingModel) IsValid() bool {
 	return 0 <= int(n) && int(n) < int(NetXConnectInvalidModel)
 }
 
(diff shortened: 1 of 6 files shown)


api.go
```diff
@@ -55,6 +55,7 @@ func createPodFromConfig(podConfig PodConfig) (*Pod, error) {
 	}
 
 	// Initialize the network.
+	DefaultNetInterworkingModel = p.config.NetworkConfig.DefaultNetInterworkingModel
```
jodh-intel: A few questions:

- Where is `p.config.NetworkConfig.DefaultNetInterworkingMode` being set?
- Doesn't this mean that the initial value of `DefaultNetInterworkingModel` (`ModelMacVtap`) isn't used now?
- If we need to retain the initial `ModelMacVtap`, maybe you could add a log call along the lines of:
  ```
  newNetModel := p.config.NetworkConfig.DefaultNetInterworkingModel

  if newNetModel != DefaultNetInterworkingModel {
      virtLog.WithFields(logrus.Fields{
          "old-network-model": DefaultNetInterworkingModel,
          "new-network-model": newNetworkModel,
      }).Info("switching network model")

      DefaultNetInterworkingModel = newNetModel
  }
  ```

jcvenegas: - p.config.NetworkConfig.DefaultNetInterworkingMode,  I am setting from the runtime. I mean the configuration provided by the caller to api.CreatePod should set it.
- Yes, DefaultNetInterworkingModel is being used now, but the api user (the runtime) should be able to modify what network interface need to be used ( as we support differents today). 
- Agree to log what is the default networking model. I will add it

network.go
```diff
@@ -117,8 +117,9 @@ type NetworkInterfacePair struct {
 
 // NetworkConfig is the network configuration related to a network.
 type NetworkConfig struct {
-	NetNSPath     string
-	NumInterfaces int
+	NetNSPath                   string
+	NumInterfaces               int
+	DefaultNetInterworkingModel NetInterworkingModel
```
sboeuf: Well I think you're addressing this in the wrong way here. I would prefer to see a `NetInterworkingModel` being defined through the network configuration (coming from the `configuration.toml` from the runtime), and having virtcontainers deciding about a default one in case nothing is provided.
Or we could go one step further, and say that virtcontainers needs to be provided with a valid model, and handle the default value at the runtime level. The runtime will by default choose something in case the model is not provided through the `configuration.toml`.
Did I miss something ?

sameo: Yes, `DefaultNetInterworkingModel` is not the right name here, it should be something like `InterworkingModel`. It's not the default virtcontainers model, but the one we've been asked to use from the caller.

Then probably in `initNetworkCommon` you want to set the `config.InterworkingModel` to `DefaultNetInterworkingModel` iff the provided value is invalid.

jcvenegas:
@sameo @sboeuf  added changes you suggested. 

I created a temporal PR in runtime repository that  uses this changes to allow modify the model from the runtime. 
https://github.com/clearcontainers/runtime/pull/887/files

hypervisor.go
```diff
@@ -190,6 +190,10 @@ type HypervisorConfig struct {
 	// For example, if there is a value for the "kernel" key in this map,
 	// it will be used for the pod's kernel path instead of KernelPath.
 	customAssets map[assetType]*asset
+
+	//Determines how the VM should be connected to the
+	//the container network interface
+	InterNetworkModel NetInterworkingModel
```
sboeuf: Why do you need this ?

jcvenegas: I was looking for a place to be semantically correct for the information provided vc user. 

Actually I would like to ask for feedback for the best place to put it:

Here the context:
From oci pkg we create the `podConfig` we do it based on `RuntimeConfig`
```go
type RuntimeConfig struct {
	VMConfig vc.Resources

	HypervisorType   vc.HypervisorType
	HypervisorConfig vc.HypervisorConfig

	AgentType   vc.AgentType
	AgentConfig interface{}

	ProxyType   vc.ProxyType
	ProxyConfig vc.ProxyConfig

	ShimType   vc.ShimType
	ShimConfig interface{}

	Console string
}
```
https://github.com/containers/virtcontainers/blob/master/pkg/oci/utils.go#L88

From oci  vc user looks good me that the property be placed in:
`config.HypervisorConfig.InterNetworkModel`
Because the property is related on how the Hypervisor will be connected.

On the other hand when the `PodConfig` is created, the struct that mainly carry the network configuration is:
`podConfig.NetworkConfig` so I added the in the property to `podConfig.NetworkConfig.InterNetworkModel`
`networkConfig, err := networkConfig(ocispec, config)`
https://github.com/containers/virtcontainers/pull/528/files#diff-5f81144f5833a1f9b3e10ca1e3039207R258


What I did was to add the property `HypervisorConfig` and then when the `PodConfig` is created use it to properly create/set `podConfig.NetworkConfig.InterNetworkModel`.

But now `config.HypervisorConfig.InterNetworkModel` it is just used to create the network config. `networkConfig, err := networkConfig(ocispec, config)` then is not used anymore.

So I wonder if should I not add it here or added `type RuntimeConfig struct`.

sboeuf: @jcvenegas Thanks for this summary, I understand the problem here and I think that we should make a decision in one direction or another, but keeping both ways is too confusing IMO.
I think we should not use `HypervisorConfig` because the VM itself does not need to know about the way the connection is done with the interface inside the network namespace. Instead, I would prefer that we define a new field in the `configuration.toml`, under the runtime section, something like that:
```toml
[runtime]
enable_debug = true
net_xconnect_model = "bridge"
```
From this information, you should be able to provision `podConfig.NetworkConfig` with the correct information.

This is the cleanest solution IMO. @sameo @jodh-intel WDYT ?

jodh-intel: That sounds good to me, but we might also want to also get agreement from @mcastelino first.

sboeuf: Oh yes of course, cc @mcastelino

mcastelino: @jcvenegas thanks for doing this. As such it looks good. I wanted to clarify why I added this configurable option. It was added so that we can switch back to the legacy bridge method if we found issues with macvtap. 

Once we found macvtap to be reliable I planned to just remove the bridge networking model. And now we think macvtap has proved to be good enough (except the bug that @jcvenegas for containerd, an issue that should be root caused). So we cannot remove it quite yet :(

So longer term I think we will have this value be configurable for only one reason. 
1. Use tc based interconnection model (as done in runV)
2. Use macvtap

However there are special network interfaces (VFs or Physical NICs), vhost-user-socket) where neither tc or macvtap will work well, and we use an interconnection model based on the type of the incoming interface and not driven by the defined interconnection type. (VFIO, vhost-user).

@sboeuf @egernst added support to get and store the incoming endpoint type. @sboeuf @egernst can you confirm the same please.

So the interconnection model will be used in the case of interfaces where either tc or macvtap can be used. For the end points where neither works, we will connect it to the VM in a way that is specific to the end point.

But if our performance measurements indicate the tc and macvtap give the exact same performance and there is no feature gap in the tc method, I wonder if we should just switch to tc, at which point we remove this. @sameo ?

So the final implementation will be

- tc + multiqueue tap (for normal interfaces)
- end point specific connection for "special endpoints"

/cc @sameo @sboeuf @egernst @jcvenegas @amshinde

egernst: @mcastelino -- correct -- we do not take the interconnection model into account in the event we are using a VFIO or vhost-user based network interface.  Its less about "work well" and more of "work at all," for these types.

sboeuf: @mcastelino I agree with the final implementation, but for the time being, and to be able to switch easily (for testing purpose for instance), I think we can make this interconnect model configurable.
With this in mind, I was saying that I would prefer a new field in our `configuration.toml`, rather than adding this to the `HypervisorConfig`.

mcastelino: @sboeuf agreed.

jcvenegas: @mcastelino  thanks for the clarification. Makes totally sense to me to work to just let one implementation.  For the moment to allow users change it (in any case need, debug, testing, issues with default ). What should be the place to add the option at runtime level or hypervisor level. See comment above from @sboeuf .

sboeuf: @jcvenegas I think we all agreed on the same thing here. You can go ahead and modify runtime/virtcontainers PRs so that you will rely on a new field from the configuration.toml to determine which xconnect model you need. We will need a default value in case this model is not provided through the config file.

jcvenegas:
@sboeuf  @mcastelino @sameo  @jodh-intel updated PR to provide the model using `runtimeConf.InterNetworkModel`

jcvenegas:
Updated vendor in https://github.com/clearcontainers/runtime/pull/887

network.go
jcvenegas: @sboeuf  the current changes now make sense to you?

sboeuf: @jcvenegas could you ping me back on Monday about that, I'll do the review properly.

sboeuf:
@jcvenegas any updates on this one ?

sboeuf:
@jcvenegas cool thx

network.go
```diff
@@ -58,8 +61,44 @@ const (
 	// when requested by the runtime
 	// This will be used for vethtap, macvtap, ipvtap
 	ModelEnlightened
+
+	// InvalidNetInterworkingModel is the last item to check valid values by IsValid()
+	InvalidNetInterworkingModel
 )
 
+//IsValid checks if a model is valid
+func (n NetInterworkingModel) IsValid() bool {
```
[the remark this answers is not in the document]
jcvenegas: @sboeuf will be used by vc caller, in this case the runtime to allow choose the default model. https://github.com/clearcontainers/runtime/pull/887/files#diff-b4bda758a2aef091432646c354b4dc59R397

It could set it using any of the NetXConnect*Model values but as this is going to be user defined  (From configuration.toml) I opted to set the model by using an string representation.

sboeuf:
Thanks for addressing comments !
LGTM

PR #531, opened 2017-12-21
Kata Containers shim support

sameo:
The Kata Containers shim needs the following parameters:

* The agent URL (Or the proxy one when there is a proxy)
* The container ID
* The guest process token it monitors

As a consequence, the virtcontainers support for the Kata Containers shim needs:

* [X] Addition of a new shim type for Kata Containers
* [X] Implementation of the Kata Container shim (i.e. its `start()` method)


container.go
```diff
@@ -29,7 +29,8 @@ import (
 // Process gathers data related to a container process.
 type Process struct {
 	Token     string
-	Pid       int
+	HostPID   int
+	GuestPID  int
```
jodh-intel: And this is the guest kernel PID (from the root namespace perspective) of the workload? Given the potential for confusion, I think we should add comments explaining which "level" all the new PID fields refer to.

hyperstart.go
```diff
@@ -406,10 +406,10 @@ func (h *hyper) exec(pod *Pod, c Container, process Process, cmd Cmd) error {
 	}
 
 	if _, err := h.proxy.sendCmd(proxyCmd); err != nil {
-		return err
+		return guestProcess{}, err
 	}
 
-	return nil
+	return guestProcess{}, nil
```
jodh-intel: I'm not clear why hyperstart always returns an empty `guestProcess`?

sboeuf: Because the shim does not need this information. Also because no process ID is returned from the protocol.

agent.go
```diff
@@ -157,7 +169,7 @@ type agent interface {
 	createPod(pod *Pod) error
 
 	// exec will tell the agent to run a command in an already running container.
-	exec(pod *Pod, c Container, process Process, cmd Cmd) error
+	exec(pod *Pod, c Container, process Process, cmd Cmd) (guestProcess, error)
```
sboeuf: The change is okay but this means that we'll have to change the sequence from the caller, because right now, we're starting the shim first and calling into `agent.exec()` as a second step.
Also, I am worried about the `cc-shim` case where we were expecting the shim to be connected so that we won't miss any output. I am not sure if this will cause any trouble.

sameo:
@jodh-intel Good points. I added some comments, let me know if that's clearer now.

agent.go
```diff
@@ -169,7 +181,8 @@ type agent interface {
 	createContainer(pod *Pod, c *Container) error
 
 	// startContainer will tell the agent to start a container related to a Pod.
-	startContainer(pod Pod, c Container) error
+	// It returns the guest PID for the container init process.
+	startContainer(pod Pod, c Container) (guestProcess, error)
```
sboeuf: This is an issue. I mean it won't be used since the shim is expected to be started when the container is created... That's the topic of this discussion https://github.com/kata-containers/agent/issues/38#issuecomment-353249666 and here is a possible solution https://github.com/kata-containers/agent/issues/38#issuecomment-353284358.
But in any case, I am not sure the guestProcess from the StartContainer will be used since this will be too late.

sboeuf:
@sameo according to https://github.com/kata-containers/agent/pull/83, I think you can remove your first commit `agent: Return guest process information when starting a new one`.
We need another PR for virtcontainers in order to generate some unique IDs for every processes.

sameo:
@sboeuf @jodh-intel This is ready to review.

kata_shim.go
```diff
@@ -0,0 +1,110 @@
...
+		return -1, fmt.Errorf("Process token cannot be empty")
+	}
+
+	args := []string{config.Path, "-agent", params.URL, "-container", params.Container, "-exec-id", params.Token}
+	if config.Debug {
+		args = append(args, "-d")
+	}
+
+	cmd := exec.Command(args[0], args[1:]...)
+	cmd.Env = os.Environ()
```
jodh-intel: Is this really required? I'm wondering because we don't do this for the proxy.

jodh-intel:
CI is unhappy:

```
13:57:32 Run go test and generate coverage:
14:02:32 Build timed out (after 5 minutes). Marking the build as aborted.
14:02:32 Build was aborted
```

kata_shim.go
```diff
@@ -0,0 +1,110 @@
...
+
+	if params.Container == "" {
+		return -1, fmt.Errorf("Container cannot be empty")
+	}
+
+	if params.Token == "" {
+		return -1, fmt.Errorf("Process token cannot be empty")
+	}
+
+	args := []string{config.Path, "-agent", params.URL, "-container", params.Container, "-exec-id", params.Token}
```
sboeuf: I am thinking it would be cleaner to define those strings (`"-container"`, `"-exec-id"`, ...) into `github.com/kata-containers/shim` so that we can import them. This would avoid any synchronization issues between the shim and the virtcontainers shim implementation.

kata_shim.go
sboeuf: I agree we don't need it since we don't rely on any environment variable from the shim, and it might be better for security not to pass more information if this is not needed.

sameo:
@jodh-intel 

> CI is unhappy

This should be fixed now.

kata_shim.go
```diff
@@ -0,0 +1,110 @@
...
+			Setctty: true,
+			Ctty:    int(f.Fd()),
+		}
+
+	}
+	defer func() {
+		if f != nil {
+			f.Close()
+		}
+	}()
```
sboeuf: @sameo don't you think we could factorize this whole part (above my comment) into a function provided by `shim.go`. I am asking because this is exactly the same code with `cc-shim.go`.
Or maybe we could create a `commonStart()` function, taking as parameters the name of each parameter, since I have the feeling this is the only difference between both implementations.

shim.go
```diff
@@ -100,6 +110,13 @@ func newShimConfig(config PodConfig) interface{} {
 			return err
 		}
 		return ccConfig
+	case KataShimType:
+		var kataConfig KataShimConfig
```
sboeuf: Same thing I did for the proxy, I think we should not create different config structures for each implementation. We could simplify everything into a `ShimConfig` structure because I don't see any reason why we would need different configs (we expect a shim to be kind of the same thing). This can be done as a follow up PR I guess.

kata_shim.go
sameo: This is what we do for `cc_shim`, so I took it from there. I think it's probably not needed, yes.

kata_shim.go
sameo: @sboeuf Sure. This is what I did with the mock code, I'll do the same with the actual shim one as well.

kata_shim.go
sameo: Yes, I'll send a Kata PR for this.

sboeuf:
LGTM

PR #534, opened 2017-12-22
pod: Add important comment for Cmd type

### LINKED ISSUE #533: Add comment explaining why Cmd type must encode user and groups as strings
The `Cmd` type encodes the user and group details that the workload will be run as.

It currently encodes those values as strings, although since the OCI spec and the Kata gRPC protocol encode such values as integers, we would ideally like to change `Cmd` so that the user and group details are also encoded as integer values (with a conversion to strings for sending such details to hyperstart which uses string types for them).

However, we've realised this isn't possible so add a comment explaining why they are encoded as strings and must remain that way.

See:

- https://github.com/containers/virtcontainers/pull/522#discussion_r157538887
- https://github.com/clearcontainers/runtime/pull/873

jodh-intel:
The `Cmd` type encodes the user and group details that the workload will
be run as.

It currently encodes those values as strings, although since the OCI
spec and the Kata gRPC protocol encode such values as integers, we would
ideally like to change `Cmd` so that the user and group details are also
encoded as integer values (with a conversion to strings for sending such
details to hyperstart which uses string types for them).

Add a comment explaining why we cannot change the type of the user and
group fields to an integer type.

Fixes #533.


jodh-intel:
A few more details here:

- https://github.com/jodh-intel/virtcontainers/commit/a8e6aff149a2f0db033ca8ee07cc8eb0c58bc41c

devimc:
lgtm

sboeuf:
LGTM

PR #536, opened 2018-01-02
vhostuser enabling

egernst:
Changes for introducing vhost-user enabling to virtcontainers.

In future still need to add support for generic vhost-user-block devices and to test vhost-user SCSI end to end.

### DIFF
--- a/container.go
+++ b/container.go
@@ -834,7 +834,7 @@ func (c *Container) hotplugDrive() error {
 	}).Info("Block device detected")
 
 	// Add drive with id as container id
-	devID := makeBlockDevIDForHypervisor(c.id)
+	devID := makeNameID("drive", c.id)
 	drive := Drive{
 		File:   devicePath,
 		Format: "raw",
@@ -870,7 +870,7 @@ func (c *Container) removeDrive() (err error) {
 	if c.isDriveUsed() && c.state.HotpluggedDrive {
 		c.Logger().Info("unplugging block device")
 
-		devID := makeBlockDevIDForHypervisor(c.id)
+		devID := makeNameID("drive", c.id)
 		drive := Drive{
 			ID: devID,
 		}
--- a/device.go
+++ b/device.go
@@ -177,15 +177,6 @@ func newBlockDevice(devInfo DeviceInfo) *BlockDevice {
 	}
 }
 
-func makeBlockDevIDForHypervisor(deviceID string) string {
-	devID := fmt.Sprintf("drive-%s", deviceID)
-	if len(devID) > maxDevIDSize {
-		devID = string(devID[:maxDevIDSize])
-	}
-
-	return devID
-}
-
 func (device *BlockDevice) attach(h hypervisor, c *Container) (err error) {
 	// If VM has not been launched yet, return immediately.
 	// This is because we always want to hotplug block devices.
@@ -206,7 +197,7 @@ func (device *BlockDevice) attach(h hypervisor, c *Container) (err error) {
 	drive := Drive{
 		File:   device.DeviceInfo.HostPath,
 		Format: "raw",
-		ID:     makeBlockDevIDForHypervisor(device.DeviceInfo.ID),
+		ID:     makeNameID("drive", device.DeviceInfo.ID),
 	}
 
 	// Increment the block index for the pod. This is used to determine the name
@@ -246,7 +237,7 @@ func (device BlockDevice) detach(h hypervisor) error {
 		deviceLogger().WithField("device", device.DeviceInfo.HostPath).Info("Unplugging block device")
 
 		drive := Drive{
-			ID: makeBlockDevIDForHypervisor(device.DeviceInfo.ID),
+			ID: makeNameID("drive", device.DeviceInfo.ID),
 		}
 
 		if err := h.hotplugRemoveDevice(drive, blockDev); err != nil {
--- a/device.go
+++ b/device.go
@@ -174,18 +174,72 @@ const (
 	VhostUserBlk = "vhost-user-blk-pci"
 )
 
-// VhostUserDevice refers to a vhost-user device implemntation.
-type VhostUserDevice struct {
-	DeviceType    string
-	DeviceInfo    DeviceInfo
-	SocketPath    string
-	HardAddr      string //valid only if VhostUserType is VhostUserNet
-	ID            string
-	VhostUserType VhostUserDeviceType
+// VhostUserDevice represents a vhost-user device. Shared
+// attributes of a vhost-user device can be retrieved using
+// the Attrs() method. Unique data can be obtained by casting
+// the object to the proper type.
+type VhostUserDevice interface {
+	Attrs() *VhostUserDeviceAttrs
+	Type() string
 }
 
-func (device *VhostUserDevice) attach(h hypervisor, c *Container) (err error) {
+// VhostUserDeviceAttrs represents data shared by most vhost-user devices
+type VhostUserDeviceAttrs struct {
+	DeviceType string
+	DeviceInfo DeviceInfo
+	SocketPath string
+	ID         string
+}
 
+// VhostUserNetDevice is a network vhost-user based device
+type VhostUserNetDevice struct {
+	VhostUserDeviceAttrs
+	MacAddress string
+}
+
+// Attrs returns the VhostUserDeviceAttrs associated with the vhost-user device
+func (vhostUserNetDevice *VhostUserNetDevice) Attrs() *VhostUserDeviceAttrs {
+	return &vhostUserNetDevice.VhostUserDeviceAttrs
+}
+
+// Type returns the type associated with the vhost-user device
+func (vhostUserNetDevice *VhostUserNetDevice) Type() string {
+	return VhostUserNet
+}
+
+// VhostUserSCSIDevice is a SCSI vhost-user based device
+type VhostUserSCSIDevice struct {
+	VhostUserDeviceAttrs
+}
+
+// Attrs returns the VhostUserDeviceAttrs associated with the vhost-user device
+func (vhostUserSCSIDevice *VhostUserSCSIDevice) Attrs() *VhostUserDeviceAttrs {
+	return &vhostUserSCSIDevice.VhostUserDeviceAttrs
+}
+
+// Type returns the type associated with the vhost-user device
+func (vhostUserSCSIDevice *VhostUserSCSIDevice) Type() string {
+	return VhostUserSCSI
+}
+
+// VhostUserBlkDevice is a block vhost-user based device
+type VhostUserBlkDevice struct {
+	VhostUserDeviceAttrs
+}
+
+// Attrs returns the VhostUserDeviceAttrs associated with the vhost-user device
+func (vhostUserBlkDevice *VhostUserBlkDevice) Attrs() *VhostUserDeviceAttrs {
+	return &vhostUserBlkDevice.VhostUserDeviceAttrs
+}
+
+// Type returns the type associated with the vhost-user device
+func (vhostUserBlkDevice *VhostUserBlkDevice) Type() string {
+	return VhostUserBlk
+}
+
+// vhostUserAttach handles the common logic among all of the vhost-user device's
+// attach functions
+func vhostUserAttach(device VhostUserDevice, h hypervisor, c *Container) (err error) {
 	// generate a unique ID to be used for hypervisor commandline fields
 	randBytes, err := generateRandomBytes(8)
 	if err != nil {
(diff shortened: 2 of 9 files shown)


egernst:
...testing still WIP...

egernst:
@mcastelino -- take a look?

network.go
```diff
@@ -1184,6 +1290,27 @@ func isPhysicalIface(ifaceName string) (bool, error) {
...
+	if netInfo.Iface.Name == "lo" {
+		return false, "", nil
+	}
+
+	// for now, check for socket file existence at known location.
+	// Identify this be searching for a file in a path which contains
+	// the interfaces IP address.  ie:
+	//  /tmp/vhostuser_<ipv4 address>/vhu.sock
+	for _, addr := range netInfo.Addrs {
+		socketPath := fmt.Sprintf("/tmp/vhostuser_%s/vhu.sock", addr.IPNet.IP)
```
sboeuf: I think we had a discussion some time ago, about not putting things into /tmp since it can be accessed by anyone on the system. It'd be safer to use a path like `/var/run/virtcontainers/pods/<podID>/vhostuser_<IP>/vhu.sock`.

egernst: @sboeuf  

Let me see what I can do here.  

This is actually not a function of the runtime, but a function of where the CNI and/or CNM plugin places these.  Let me see what I can change --  I'm consuming these sockets, not creating them.

sboeuf: Oh I see, then ignore this comment.

qemu.go
```diff
@@ -1028,3 +1028,13 @@ func (q *qemu) getPodConsole(podID string) string {
 func (q *qemu) getState() interface{} {
 	return q.state
 }
+
+//Generic function for creating a named-id for passing on the hypervisor commandline
+func makeNameID(namedType string, id string) string {
```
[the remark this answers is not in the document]
egernst: @sboeuf  -- I think moving it to hypervisor.go is the most straight forward at this point.  You okay with that?

qemu.go
```diff
@@ -286,6 +286,28 @@ func (q *qemu) appendBlockDevice(devices []govmmQemu.Device, drive Drive) []govm
...
 
+func (q *qemu) appendVhostuserDevice(devices []govmmQemu.Device, vhostuserDevice VhostUserDevice) []govmmQemu.Device {
+
+	typeDev := ""
+	// Only two types of vhost-user devices supported: net or scsi
+	if vhostuserDevice.VhostUserType == govmmQemu.VhostUserNet {
+		typeDev = "net"
+	} else {
+		typeDev = "scsi"
+	}
```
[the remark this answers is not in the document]
egernst: I think this'll soon be a case statement, as there'll be support for vhostuser-block as well.

qemu.go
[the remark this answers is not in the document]
sboeuf: Yes sure, the second proposal can be part of a separate PR.

qemu.go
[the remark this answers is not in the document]
sboeuf: Fair enough.

egernst:
@sboeuf -- agreed this is non-marginally sub-optimal method (the discovery of the socket in the network endpoint case).  AFAICT it can't be made that generic, however, as this isn't just a socket being passed to qemu.  In govmm we are taking this device type and appending a net device onto the command line, including setting/passing ip, route, Mac, etc.  @mcastelino  -- you have further thoughts here?

device.go
```diff
@@ -161,6 +161,48 @@ func (device *VFIODevice) deviceType() string {
...
+// VhostUserDevice refers to a vhost-user device implemntation.
+type VhostUserDevice struct {
+	DeviceType    string
+	DeviceInfo    DeviceInfo
+	SocketPath    string
+	HardAddr      string //valid only if VhostUserType is VhostUserNet
+	ID            string
+	VhostUserType VhostUserDeviceType
+}
+
```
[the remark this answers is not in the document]
egernst: This pull is just enabling adding the network based device, not a generic vhost-user at this point.  The device is created during "addDevice," called in the endpoints' Attach function.  For the case of qemu, it is done in appendVhostUserDevice.  When looking at non-network based vhost-user, we can plan on adding this kind of ```newVhostUserDevice``` function.

qemu.go
```diff
@@ -286,6 +286,28 @@ func (q *qemu) appendBlockDevice(devices []govmmQemu.Device, drive Drive) []govm
 	return devices
 }
 
+func (q *qemu) appendVhostuserDevice(devices []govmmQemu.Device, vhostuserDevice VhostUserDevice) []govmmQemu.Device {
```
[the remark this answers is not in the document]
egernst: done

qemu.go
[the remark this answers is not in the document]
egernst: done

device.go
```diff
@@ -100,6 +100,16 @@ type DeviceInfo struct {
 	ID string
 }
 
+//Generic function for creating a named-id for passing on the hypervisor commandline
+func makeNameID(namedType string, id string) string {
```
sboeuf: It is still in `device.go`. I thought we had agreed on moving this to `hypervisor.go`
Also, please update your commit message.

device.go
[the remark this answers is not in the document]
sboeuf: My understanding of your PR was that you're adding a global support for vhost-user, regardless of "non-network vs network" cases. Your commit message says `Generic vhost-user enabling for non-network devices (like SCSI) still
needs to be tested.`, that's why I was assuming you were covering this case in this commit. Also, you've added both `net` and `scsi` in `qemu.go`, that's why it is a bit confusing.
If it's clear that you want to add only vhost-user support for the network case, I am fine with that, but make sure it's clear from the code and the commit message.

egernst:
Good points, Sebastien.  Let me make some updates.

network.go
mcastelino: This ``/tmp/vhostuser_%s/vhu.sock`` is based on the CNI/CNM plugin we use. Ideally this should be configurable. vhostuser is very flexible in that the socket can be in its own dir or all sockets can also be created in the same directory. We can make an assumption that there is a 
- single socket / dir
What happens when there are two vhost user networks (say VPP with one L2 switch for inbound and L2 switch for outbound). 

When there are unique IP addresses across networks this scheme will work.

The only issue I can see is if there are 2 networks and 2 containers. But both the networks have the same IP, but each container connects to only one network. This is a corner case which we can ignore now as this will not show up in the case of kubernetes.

mcastelino: So in summary let us make this search location configurable.

qemu.go
```diff
@@ -286,6 +286,30 @@ func (q *qemu) appendBlockDevice(devices []govmmQemu.Device, drive Drive) []govm
 	return devices
 }
 
+func (q *qemu) appendVhostUserDevice(devices []govmmQemu.Device, vhostUserDevice VhostUserDevice) []govmmQemu.Device {
```
mcastelino: Having this function and having different types for each vhost-user type means you have two options. Implement an interface or have multiple functions. Or send in the type as an interface and perform a type switch internally.

qemu.go
```diff
@@ -1009,6 +1033,9 @@ func (q *qemu) addDevice(devInfo interface{}, devType deviceType) error {
 	case blockDev:
 		drive := devInfo.(Drive)
 		q.qemuConfig.Devices = q.appendBlockDevice(q.qemuConfig.Devices, drive)
+	case vhostuserDev:
```
mcastelino: Again this will be based on the actual device type

mcastelino: @sboeuf why are we not doing a straight type switch here?

``
switch v := devInfo .(type) {
	default:
		fmt.Printf("I don't know about type %T!\n", v)
``

qemu.go
egernst: I'll end up moving to using a VhostUserDevice base type, and switch on the subtype (VhostUserNetDevice, VhostUserSCSIDevice, etc).

network.go
```diff
@@ -200,6 +212,69 @@ func (endpoint *VirtualEndpoint) Detach() error {
...
+	networkLogger().Info("Attaching vhostuser based endpoint")
+
+	// generate a unique ID to be used for hypervisor commandline fields
+	randBytes, err := generateRandomBytes(8)
+	if err != nil {
+		return err
+	}
+	id := hex.EncodeToString(randBytes)
+
+	d := VhostUserDevice{
```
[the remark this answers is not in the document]
egernst: done...

egernst:
@sboeuf @mcastelino  -- update to address our comments.

sboeuf:
@egernst Thanks, I'll take a look !

device.go
```diff
@@ -161,6 +161,142 @@ func (device *VFIODevice) deviceType() string {
...
+
+// vhostUserAttach handles the common logic among all of the vhost-user device's
+// attach functions
+func vhostUserAttach(device VhostUserDevice, h hypervisor, c *Container) (err error) {
+	// generate a unique ID to be used for hypervisor commandline fields
+	randBytes, err := generateRandomBytes(8)
+	if err != nil {
+		return err
+	}
+	id := hex.EncodeToString(randBytes)
```
mcastelino: does this need the prefix of the type? I see prefixes being used in some places

egernst: @mcastelino 

No -- the "prefix-ID" is created further down the line.  In our case it is handled in the qemu appendVhostUserDevice function.

network.go
```diff
@@ -1184,6 +1290,31 @@ func isPhysicalIface(ifaceName string) (bool, error) {
...
+// for a vhost-user socket
+func isVhostuserIface(netInfo NetworkInfo) (bool, string, error) {
+	if netInfo.Iface.Name == "lo" {
+		return false, "", nil
+	}
+
+	// for now, check for socket file existence at known location.
+	// Identify this be searching for a file in a path which contains
+	// the interfaces IP address.  ie:
+	//  /tmp/vhostuser_<ipv4 address>/vhu.sock
```
sameo: Here we should explain where the `/tmp/vhostuser_ipv4/vhu.sock` pattern is coming [from](https://github.com/clearcontainers/vpp/pull/5/files).

network.go
```diff
@@ -1184,6 +1290,31 @@ func isPhysicalIface(ifaceName string) (bool, error) {
...
+	for _, addr := range netInfo.Addrs {
+		socketPath := fmt.Sprintf(hostSocketSearchPath, addr.IPNet.IP)
```
sameo: It would make sense to have a `vhosUserSocketPath(info interface{}) (string, error)` function that would try all known heuristics (one for now) for finding a  vhost-user socket path and return it if it can find one.

hyperstart_agent.go
```diff
@@ -53,6 +53,16 @@ func (c *HyperConfig) Logger() *logrus.Entry {
 	return virtLog.WithField("subsystem", "hyperstart")
 }
 
+//Generic function for creating a named-id for passing on the hypervisor commandline
```
sameo: That's not hyperstart specific afaict. Please move it to e.g. hypervisor.go.

qemu.go
```diff
@@ -1018,31 +1018,25 @@ func (q *qemu) resumePod() error {
...
+	case Volume:
+		q.qemuConfig.Devices = q.appendVolume(q.qemuConfig.Devices, v)
+	case Socket:
+		q.qemuConfig.Devices = q.appendSocket(q.qemuConfig.Devices, v)
+	case Endpoint:
+		q.qemuConfig.Devices = q.appendNetwork(q.qemuConfig.Devices, v)
+	case Drive:
+		q.qemuConfig.Devices = q.appendBlockDevice(q.qemuConfig.Devices, v)
+
+	//vhostUserDevice is an interface, hence the pointer for Net, SCSI and Blk:
```
sboeuf: Could not you simplify like this:
```go
        case VhostUserNetDevice, VhostUserSCSIDevice, VhostUserBlkDevice:
		q.qemuConfig.Devices = q.appendVhostUserDevice(q.qemuConfig.Devices, &v)
```

sboeuf: I like the refactoring BTW.

hyperstart_agent.go
sboeuf: @egernst I think you put that in the wrong file. We had agreed the most simple/generic would be `hypervisor.go`.

qemu.go
egernst: No, you cannot in the case of a type switch, unfortunately.  It pained me to not be able to!

network.go
egernst: Sure. @sameo  - you mean to say add a function to the vhostUserDevice interface, right?  Good call...

network.go
egernst: Sure.  Will do.

device.go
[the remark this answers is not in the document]
egernst: @sboeuf I made changes already based on what @mcastelino had described.  VhostUserDeviceAttrs is a "base type" structure which we now embed in resulting SCSI, net and block vhostuser devices.  VhostUserDevice is the interface which the SCSI, net and block devices "implement."

network_test.go
```diff
@@ -298,6 +330,83 @@ func TestCreateNetworkEndpointsFailure(t *testing.T) {
...
+	}
+	netinfo := NetworkInfo{
+		Addrs: addresses,
+	}
+
+	isVhost, path, _ := isVhostuserIface(netinfo)
+
+	if isVhost != true {
+		t.Fatalf("Got %+v\nExpecting %+v", isVhost, true)
+		return
```
amshinde: You should not require a return here.

network_test.go
```diff
@@ -298,6 +330,83 @@ func TestCreateNetworkEndpointsFailure(t *testing.T) {
...
+	}
+	if path != expectedResult {
+		t.Fatalf("Got %+v\nExpecting %+v", path, expectedResult)
+		return
```
amshinde: Same as above.

qemu.go
```diff
@@ -287,6 +287,30 @@ func (q *qemu) appendBlockDevice(devices []govmmQemu.Device, drive Drive) []govm
...
+func (q *qemu) appendVhostUserDevice(devices []govmmQemu.Device, vhostUserDevice VhostUserDevice) []govmmQemu.Device {
+
+	typeDev := ""
+	switch vhostUserDevice.VhostUserType {
+	case VhostUserNet:
+		typeDev = makeNameID("net", vhostUserDevice.ID)
+	case VhostUserSCSI:
+		typeDev = makeNameID("scsi", vhostUserDevice.ID)
+	case VhostUserBlk:
+		typeDev = "" //vhost-user-blk-pci doesn't include a device ID
```
amshinde: You should consider adding a default case here for completeness.

qemu.go
amshinde: You can use "fallthrough" in a switch iirc.

container.go
```diff
@@ -834,7 +834,7 @@ func (c *Container) hotplugDrive() error {
 	}).Info("Block device detected")
```
amshinde: Related to the commit message. Typo /idenitfy/identify/

qemu.go
egernst: The issue is that the resolution of &v isn't carried out correctly when it isn't explicitly done this way (because it is a type switch and VhostUserDevice is an interface).

network_test.go
egernst: done

network_test.go
egernst: done

qemu.go
amshinde: Interesting, didnt know the "fallthrough" keyword didnt work for type switches.

container.go
egernst: thanks - fixed....

network.go
sameo: @egernst yes, `isVhostUserInterface()` is almost what I'm looking for, except that:

1. It's network specific
2. It's `clearcontainers/vpp` plugin specific
3. It returns both a boolean and the socket string

So my suggestion is to make it generic, have it embed the `clearcontainers/vpp` specific bits, and only returns a `(string, error)` where a nil string means no socket path could be found.
Once we understand how to handle the storage case, we can just add it to this routine.

sameo:
LGTM

sboeuf:
Yes looks good to me too. I'd like to wait for @amshinde approval before we merge.
@egernst the CI is not passing.

amshinde:
lgtm

egernst:
@sboeuf -- let me know if you have any more inputs on this patchset.  Thanks.

sboeuf:
@egernst nope, it's all good !

PR #537, opened 2018-01-03
ci: Handle complex revendoring cases

### LINKED ISSUE #513: ci: Improve handling of complex revendoring cases
We have run into an issue yesterday with #494, and there was no way to make the CI pass because of the vendoring complexity. This PR was about adding some code to virtcontainers, but at the same time it was revendoring several packages. The way our CI work in order to test our virtcontainers PRs with the Clear Containers runtime is pretty basic. It copies all the new files to the vendoring inside `runtime/vendor/..../virtcontainers`, but this is not enough in this case. Indeed, the runtime vendoring for the other packages that have been revendored from virtcontainers needs to be updated too.
Because I think that we don't want to handle such complexity with nested layers of vendoring, we should rely mostly on `dep` to do the job.

Therefore, here is the proposal:
- We should have our CI script creating a new branch and pushing to a temporary branch on virtcontainers repo.
- Go to the runtime repository and checkout the right branch/PR(in case we have `Depends-on:` flag)
- Modify the `Gopkg.toml` file to update the constraint revision of virtcontainers as the one corresponding to the virtcontainers branch we've just pushed/created from the CI.
- Run `dep ensure && dep ensure -update github.com/containers/virtcontainers && dep prune`
- Build the runtime from this point, and start our testing as it's usually done.

sboeuf:
ci: Handle complex revendoring cases
    
    This commit relies fully on dep tool so as to vendor the latest code
    coming from a pull request on virtcontainers repository.
    
    The information about the github repo URL and the commit revision are
    both passed from Jenkins (as environment variables), and the script
    relies on those values in order to override Gopkg.toml for both proxy
    and runtime repositories.
    
    The last step is to run dep tool, leaving the revendoring complexity
    (including nested vendoring sometimes) being handled by dep.
    
    Fixes #513
    
### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -37,6 +37,9 @@ cd "${GOPATH}/src/${vc_repo}"
 if [ "${ghprbPullId}" ] && [ "${ghprbTargetBranch}" ]
 then
 	git fetch origin "pull/${ghprbPullId}/head" && git checkout master && git reset --hard FETCH_HEAD && git rebase "origin/${ghprbTargetBranch}"
+
+	export AUTHOR_REPO_GIT_URL="${ghprbAuthorRepoGitUrl}"
+	export COMMIT_REVISION="${ghprbActualCommit}"
 else
 	git fetch origin && git checkout master && git reset --hard origin/master
 fi
--- a/.ci/update-vendoring.sh
+++ b/.ci/update-vendoring.sh
@@ -53,21 +53,30 @@ function apply_depends_on(){
 	popd
 }
 
-blacklist=".ci documentation hack hook pause shim test utils vendor"
-# Copy virtcontainers changes to the vendor directory of the repo
 function update_repo(){
-	repo="$1"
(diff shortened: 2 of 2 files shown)


.ci/jenkins_job_build.sh
```diff
@@ -37,6 +37,9 @@ cd "${GOPATH}/src/${vc_repo}"
 if [ "${ghprbPullId}" ] && [ "${ghprbTargetBranch}" ]
 then
 	git fetch origin "pull/${ghprbPullId}/head" && git checkout master && git reset --hard FETCH_HEAD && git rebase "origin/${ghprbTargetBranch}"
+
+	export AUTHOR_REPO_GIT_URL="${ghprbAuthorRepoGitUrl}"
```
grahamwhaley: I presume it was a conscious decision to export these as a different name, rather than just stick with exporting their original `ghprb` names? - just curious.

sboeuf: Yes, I didn't want to end up with some Jenkins plugin naming outside of the script `jenkins_job_build.sh`

jodh-intel:
lgtm

PR #539, opened 2018-01-04
mounts: Fix bug while checking if /dev was bind-mounted

amshinde:
There was a bug in the way we were checking if /dev was
bindmounted from the host.

### DIFF
--- a/container.go
+++ b/container.go
@@ -496,11 +496,11 @@ func (c *Container) fetchState(cmd string) (State, error) {
 
 func (c *Container) getSystemMountInfo() {
 	// check if /dev needs to be bind mounted from host /dev
+	c.systemMountsInfo.BindMountDev = false
+
 	for _, m := range c.mounts {
 		if m.Source == "/dev" && m.Destination == "/dev" && m.Type == "bind" {
 			c.systemMountsInfo.BindMountDev = true
-		} else {
-			c.systemMountsInfo.BindMountDev = false
 		}
 	}
 
--- a/container_test.go
+++ b/container_test.go
@@ -58,6 +58,11 @@ func TestContainerSystemMountsInfo(t *testing.T) {
 			Destination: "/dev",
 			Type:        "bind",
 		},
+		{
+			Source:      "procfs",
+			Destination: "/proc",
+			Type:        "procfs",
+		},
 	}
 
 	c := Container{


amshinde:
cc @egernst

container.go
```diff
@@ -496,11 +496,13 @@ func (c *Container) fetchState(cmd string) (State, error) {
...
 func (c *Container) getSystemMountInfo() {
 	// check if /dev needs to be bind mounted from host /dev
+	c.systemMountsInfo.BindMountDev = false
+
 	for _, m := range c.mounts {
-		if m.Source == "/dev" && m.Destination == "/dev" && m.Type == "bind" {
-			c.systemMountsInfo.BindMountDev = true
-		} else {
-			c.systemMountsInfo.BindMountDev = false
+		if m.Destination == "/dev" {
```
egernst: Nice catch - in the original case, it would only return true of the last of the mounts happened to be the bind mount from host /dev?  

Any rationale for dividing it into two if clauses?

sboeuf: @amshinde the logic of the patch works, but I have the same question than @egernst, why don't we keep only one if statement ?
```go
if m.Source == "/dev" && m.Destination == "/dev" && m.Type == "bind" {
```

jodh-intel: Also @amshinde, can we create a test to assert the expected behaviour (here, in `runtime` or `tests`)?

amshinde: @sboeuf I just add the split to make it clear we need to care of the destination first, thats all. But yeah we could very well have it as one statement.

sboeuf:
LGTM

container.go
```diff
@@ -496,11 +496,11 @@ func (c *Container) fetchState(cmd string) (State, error) {
 
 func (c *Container) getSystemMountInfo() {
 	// check if /dev needs to be bind mounted from host /dev
+	c.systemMountsInfo.BindMountDev = false
+
 	for _, m := range c.mounts {
-		if m.Source == "/dev" && m.Destination == "/dev" && m.Type == "bind" {
+		if m.Destination == "/dev" && m.Source == "/dev" && m.Type == "bind" {
```
jodh-intel: Huh? Isn't this change is a nop (and confusing)?

sameo: @amshinde Could you please revert that change so that we can merge that PR?

amshinde: @jodh-intel @sameo Seems like I missed repushing a change I made yesterday:( 
Changes are in.

jodh-intel:
lgtm

sameo:
LGTM

jodh-intel:
Coveralls is "stuck" again (https://github.com/clearcontainers/jenkins/issues/9), but has actually run if you look at the 16.04 CI build log: https://coveralls.io/jobs/32568132.

It claims coverage fell by 3.5% but clearly it's lying, so merging.

PR #542, opened 2018-01-05
Add support for linux capabilities

amshinde:


amshinde:
Change in the agent https://github.com/clearcontainers/agent/pull/188

pkg/oci/utils.go
```diff
@@ -255,6 +255,62 @@ func containerDeviceInfos(spec CompatOCISpec) ([]vc.DeviceInfo, error) {
...
+					c.Ambient = list
+				case "permitted":
+					c.Permitted = list
+				}
+
+			default:
+				return c, fmt.Errorf("Unexpected format for capabilities: %v", caps)
+			}
+		}
+	case []interface{}:
```
sboeuf: I would expect this to be:
```go
case []string:
```
This way, there is no need for another cast into a string later.

pkg/oci/utils.go
```diff
@@ -255,6 +255,62 @@ func containerDeviceInfos(spec CompatOCISpec) ([]vc.DeviceInfo, error) {
...
 
+func containerCapabilities(s CompatOCISpec) (vc.LinuxCapabilities, error) {
+	capabilities := s.Process.Capabilities
+	var c vc.LinuxCapabilities
+
+	// In spec v1.0.0-rc4, capabilities was a list of strings. This was changed
+	// to an object with v1.0.0-rc5.
+	// Check for the interface type to support both the versions.
+	switch caps := capabilities.(type) {
+	case map[string]interface{}:
```
sboeuf: I would expect this to be:
```go
import "github.com/opencontainers/runtime-spec/blob/master/specs-go"

...

case specs.LinuxCapabilities:
```
And this would simplify the whole conversion you're doing here since `specs.LinuxCapabilities` is the same structure than `vc.LinuxCapabilities`.

amshinde: @sboeuf Unfortunately it is not possible to do it that way. This is because of the interface{} introduced for Capabilities:
https://github.com/containers/virtcontainers/blob/470d2699f12d92ab3cac8fe1e161c36437820788/pkg/oci/utils.go#L65
This interface shadows the Capabilities field in the specs Process structure, causing the json to be unmarshalled in the interface. We need to do the necessary typecasting after that as seen here:
https://blog.golang.org/json-and-go (Topic: Decoding arbitrary data)

For 1.0.0rc5, this Capabilities is a  json object, while for rc4 it is a json array.
So in the first case, Capabilities gets unmarshalled into map[string]interface{}, while in the second case it is unmarshalled to []nterface{}

sboeuf: Oh yeah I get it now...

sboeuf:
LGTM

sameo:
LGTM

sameo:
Please rebase and we'll merge.

amshinde:
@sameo @sboeuf Changes have been rebased.

sameo:
@amshinde I guess those changes depend on the agent ones to pass CI?

pkg/oci/utils.go
```diff
@@ -246,6 +246,62 @@ func containerDeviceInfos(spec CompatOCISpec) ([]vc.DeviceInfo, error) {
 	return devices, nil
 }
 
+func containerCapabilities(s CompatOCISpec) (vc.LinuxCapabilities, error) {
```
jodh-intel: Could you create a unit test for this new function?

amshinde:
@sameo The tests in the tests repo were passing an empty structure for capabilities. I have added code to handle the nil case.

pkg/oci/utils.go
jodh-intel: Thanks! :smile:

PR #543, opened 2018-01-08
vendor: Force kata containers agent vendoring

sameo:
So that all runtimes based on virtcontainers will pull the same
kata agent commit and thus the same gRPC API.


sameo:
cc @sboeuf @jodh-intel

sboeuf:
LGTM

PR #546, opened 2018-01-09
debug: Check why CI does not work

sboeuf:
CI does not work as expected, make some testing


sboeuf:
Closing because of the fix here: https://github.com/containers/virtcontainers/pull/547

PR #549, opened 2018-01-10
Kata agent and proxy fixes

sameo:
Now that we can test all kata containers components with virtcontainers, here is my first round of fixes:

- Build and store a default gRPC socket.
- Move the gRPC client handling from kata_agent to kata_proxy.
- Misc fixes from Sebastien.

### DIFF
--- a/kata_proxy.go
+++ b/kata_proxy.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"os/exec"
+	"syscall"
 
 	kataclient "github.com/kata-containers/agent/protocols/client"
 	"github.com/kata-containers/agent/protocols/grpc"
@@ -91,7 +92,8 @@ func (p *kataProxy) register(pod Pod) ([]ProxyInfo, string, error) {
 
 // unregister is kataProxy unregister implementation for proxy interface.
 func (p *kataProxy) unregister(pod Pod) error {
-	return nil
+	// Kill the proxy. This should ideally be dealt with from a stop method.
+	return syscall.Kill(pod.state.ProxyPid, syscall.SIGKILL)
 }
 
 // connect is kataProxy connect implementation for proxy interface.
--- a/kata_proxy.go
+++ b/kata_proxy.go
@@ -17,7 +17,12 @@
 package virtcontainers
 
 import (
+	"context"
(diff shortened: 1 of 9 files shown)


jodh-intel:
Ping @sboeuf.

sboeuf:
@jodh-intel I am gonna review this PR today !

kata_proxy.go
```diff
@@ -84,15 +96,56 @@ func (p *kataProxy) unregister(pod Pod) error {
...
 }
 
 // sendCmd is kataProxy sendCmd implementation for proxy interface.
 func (p *kataProxy) sendCmd(cmd interface{}) (interface{}, error) {
-	return nil, nil
+	if p.client == nil {
+		return nil, fmt.Errorf("Client is nil, we can't interact with kata-proxy")
+	}
+
+	switch c := cmd.(type) {
```
sboeuf: I don't like this whole commit `kata_proxy: Implement all interface ops` because we are removing the support for `no_proxy` in that case... Indeed, the goal of no proxy is to bypass the proxy. I understand what we are trying to do here but I am wondering if we shouldn't avoid this kind of patch.

sameo:
@sboeuf I understand your concerns and it took me half a day to make the decision to go and implement the kata_proxy ops, for the following reasons:
* Make the current code consistent across proxy and agent implementations. Implementing everything inside kata_agent might be easier, but would make the current code base even more difficult to follow by having on one hand a cc_proxy+hyperstart_agent pattern that is different than the kata_proxy+kata_agent one.
* Implementing everything into kata_agent would not fix the overall problem we have with our agent+proxy implementation, it would only be an hybrid approach. I want to get a functional and consistent implementation working the broken proxy API that we have today, and then move to a reorganization of the agent+proxy architecture.
* No proxy could be supported with that approach and it could even be done through kata_proxy.go. If you use the kata_proxy.go code and a) do not start a proxy binary and b) Use the VSOCK URL as the proxy URL, you basically have a no_proxy implementation.

sboeuf:
LGTM

sboeuf:
PR merged since the Ubuntu 17.04 failure was not related to this PR but to the incapacity to reach some Ubuntu 17.04 repos.

Issue #554, opened 2018-01-12
resource allocation: clarify the comments on how we calculate this


grahamwhaley:
Clarify how we calculate how much RAM and how many vCPUs to allocate to a VM, as per:
https://github.com/containers/virtcontainers/blob/master/pkg/oci/utils.go#L402

and comment:
https://github.com/containers/virtcontainers/pull/541/files#r160336618

PR #556, opened 2018-01-12
pkg/oci: Clarify resource calculation comment

grahamwhaley:
We use a couple of relativley simple formulas to work out how much
RAM and how many vCPUs to allocate to the VM based on the
Resource request we are passed in.
Improve the comment on these calculations to make it clear which
way they round and if they impose a floor to the resulting value.

Fixes: #554

Signed-off-by: Graham Whaley <graham.whaley@intel.com>

### DIFF
--- a/pkg/oci/utils.go
+++ b/pkg/oci/utils.go
@@ -412,7 +412,8 @@ func vmConfig(ocispec CompatOCISpec, config RuntimeConfig) (vc.Resources, error)
 		if memBytes <= 0 {
 			return vc.Resources{}, fmt.Errorf("Invalid OCI memory limit %d", memBytes)
 		}
-		// round up memory to 1MB
+		// Use some math magic to round up to the nearest Mb.
+		// This has the side effect that we can never have <1Mb assigned.
 		resources.Memory = uint((memBytes + (1024*1024 - 1)) / (1024 * 1024))
 	}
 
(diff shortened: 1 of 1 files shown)


grahamwhaley:
Just fyi, even though this is a benign comment change, I wanted to build and test check it, but don't seem able to do that on my machine right now.

pkg/oci/utils.go
```diff
@@ -430,7 +431,12 @@ func vmConfig(ocispec CompatOCISpec, config RuntimeConfig) (vc.Resources, error)
 			return vc.Resources{}, fmt.Errorf("Invalid OCI cpu period %d", period)
 		}
 
-		// round up to 1 CPU
+		// Use some math magic to round up to the nearest whole vCPU
+		// (that is, a partial part of a quoata request ends up assigning
```
jodh-intel: Nit: "quota".

jodh-intel:
+1 for good comments!

lgtm

devimc:
lgtm

devimc:
thanks @grahamwhaley for improve the comments

pkg/oci/utils.go
grahamwhaley: fixed :-)

jodh-intel:
The CI builds all finished successfully and coveralls also ran successfully, so why it's complaining I'm not sure. Merging...

PR #568, opened 2018-01-17
kata_agent: Constraints the OCI spec

### LINKED ISSUE #559: kata_agent: Constraint the OCI spec
Passing the whole OCI spec provided through the config.json will not work because not everything will apply properly. Indeed, we are in a virtualized context, meaning that we have to limit what we provide to the agent.
Here is a list of things to do:
- [ ] `NEWNET` has to be removed from libcontainer `config.Namespaces`
- [ ] As a consequence of the previous point, we should not provide anything through `config.Networks`
- [ ] `config.Hooks` entry has to be `nil`
- [ ] `config.Seccomp` entry has to be `nil`
- [ ] `config.NoNewKeyring` has to be `true` since we don't want to force the kernel to support it
- [ ] `config.Mounts` has to be analyzed to remove unwanted mounts (TBD)

sboeuf:
This commit constraints the OCI spec provided to the agent on
several point such as seccomp, hooks, network namespace and
Intel RDT.

Fixes #559

### DIFF
--- a/kata_agent.go
+++ b/kata_agent.go
@@ -364,6 +364,38 @@ func (k *kataAgent) replaceOCIMountSource(spec *specs.Spec, guestMounts []*Mount
 	return nil
 }
 
+func constraintGRPCSpec(grpcSpec *grpc.Spec) {
+	// Disable Hooks since they have been handled on the host and there is
+	// no reason to send them to the agent. It would make no sense to try
+	// to apply them on the guest.
+	grpcSpec.Hooks = nil
+
+	// Disable Seccomp since they cannot be handled properly by the agent
+	// until we provide a guest image with libseccomp support. More details
+	// here: https://github.com/kata-containers/agent/issues/104
+	grpcSpec.Linux.Seccomp = nil
+
+	// Disable network namespace since it is already handled on the host by
+	// virtcontainers. The network is a complex part which cannot be simply
+	// passed to the agent.
+	for idx, ns := range grpcSpec.Linux.Namespaces {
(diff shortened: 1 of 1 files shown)


kata_agent.go
```diff
@@ -445,6 +445,22 @@ func (k *kataAgent) createContainer(pod *Pod, c *Container) error {
 	// We need to give the OCI spec our absolute rootfs path in the guest.
 	grpcSpec.Root.Path = rootPath
 
+	// Remove Hooks
+	grpcSpec.Hooks = nil
+
+	// Remove Seccomp
+	grpcSpec.Linux.Seccomp = nil
```
sameo: Any reason why we're removing SecComp? It should not be an issue to apply those inside the VM.

sameo:
Besides the fact that I think we could keep seccomp in:
LGTM

kata_agent.go
sboeuf: Unfortunately, we need several things to make the seccomp usable (as explained here https://github.com/kata-containers/agent/issues/104).
This is temporary until we can support seccomp with Kata Containers agent.

kata_agent.go
```diff
@@ -445,6 +445,22 @@ func (k *kataAgent) createContainer(pod *Pod, c *Container) error {
...
+
+	// Remove Intel RDT
```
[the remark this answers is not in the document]
sboeuf: All done !

sameo:
Now for real:
LGTM

PR #572, opened 2018-01-18
logging: Set source field for oci and hyperstart packages

### LINKED ISSUE #571: Make package logger specify correct source field
The `oci` and `hyperstart` packages have a `SetLogger()` API. But they don't correctly set the `source` field. This means that if a caller uses `SetLogger()`, all log messages coming from these packages will appear to come *from* the caller, not from virtcontainers.

See: https://github.com/clearcontainers/runtime/issues/934

jodh-intel:
Change the `SetLogger()` API calls in the `oci` and `hyperstart`
packages so that they set the `source` logger field. This allows
callers making use of these packages to see the true source of the log
messages (previously, it would appear that log calls made by these two
packages would be coming *from* the caller itself).

Fixes #571.


sboeuf:
LGTM

Issue #518, opened 2017-12-15
Rework of the proxy interface


sboeuf:
With both `cc-proxy`, `kata-proxy` and the `no-proxy` cases, I have been thinking that our current `proxy` interface does not meet all the needs for those different cases...
First I think we don't need to return a URL when calling into `connect()` or a list of URL when calling into `register()`, only one URL should be returned by `start()` now that we agreed on the model one proxy per VM. This opinion is based on the fact that a proxy is at a pod/VM level and I don't expect several URLs to be provided to the shim, the same way this shim will connect to only one socket on the VM in case we don't go through a proxy.
Then I think that most of the interface is fine, except `sendCmd()`. I have been turning the problem upside down, but no way to make all the cases work with this. We should agree on what should be a proxy in order to know what to expect from it. I think a proxy has to be accessed the same way we would access the VM directly, that is through a socket. This is already the case for all our different cases, but the problem is about the way the data are processed and are expected to be received by this proxy or directly by the VM. The most generic way I can think of is by sending an array of bytes through the socket opened through the `connect()`. This would suggest that `connect()` would actually return a handler to this socket connection. We should get rid of the `sendCmd()` because it assumes the agent to know about the kind of proxy that is going to be used, while a pure array of bytes sent through the provided socket would be processed either by the proxy or directly by the agent, transparently for virtcontainers. Indeed, we should use this generic socket handler from the agent implementation in order to call into some specific functions (specific to the agent protocol) but always requiring a socket to communicate to.
All what I am suggesting here would be pretty straight forward to implement for both `kata-proxy` and `no-proxy` since they follow this model of receiving the same data, the encapsulation being done by the proxy itself if needed.
But this would require some extra work for our current `cc-proxy` and `cc_proxy.go` implementation since the encapsulation is done by the client itself.
One first step to avoid too much changes would be to consider using the `sendCmd()` the way it is right now, only for the `cc-proxy` case, while we would not use it for other new implementations. Instead, we would only use the subset `start()`, `register()`, `unregister()`, `connect()`, `disconnect()`.
This is the first time we're really testing the flexibility of virtcontainers interfaces, and I would like to make sure we move in the right direction by refining them in case we see a need for that. And I think this is the case here.

@sameo @jodh-intel WDYT ?

sameo:
[...]

PR #575, opened 2018-01-19
Start shims after creating containers

sameo:
Right now the virtcontainers core code explictly starts shims as soon as possible, regardless of the shim implementation being able to handle non existing containers or not.

This PR reorders that sequence and guarantees that shims will not be started before their corresponding container process is not started.

Now we have:

- container.go `createContainer` basically only allocate a new container structure and asks the agent to create it on the guest.
- It is up to each agent implementation to decide if there's a need to associate a shim to a container's process or not. i.e. the agent is now responsible for launching the shims.
- pod.go `startShims` is now `createContainers()` and calls into `createContainer()` for each configured container.

### DIFF
--- a/container.go
+++ b/container.go
@@ -230,7 +230,7 @@ func (c *Container) startShim(token string, initProcess bool) (*Process, error) 
 		processToken = proxyInfo.Token
 	}
 
-	process, err := c.createShimProcess(processToken, url, c.config.Cmd)
+	process, err := c.startShimProcess(processToken, url, c.config.Cmd)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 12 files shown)


sameo:
@sboeuf This is an attempt at reorganizing the code for having the shims only started after the container is created on the guest, in order to avoid the SIGUSR1 synchronization.

sameo:
This should pass unit testing now, at least :-/

sameo:
Basic integration tests now work with kata and CC with that PR. I'd say it's ready for review @sboeuf @jodh-intel

sboeuf:
@sameo cool, I'll review this tomorrow.

container.go
```diff
@@ -216,24 +215,39 @@ func (c *Container) GetAnnotations() map[string]string {
...
-		return err
+		return nil, err
 	}
 
 	if err := c.pod.proxy.disconnect(); err != nil {
-		return err
+		return nil, err
+	}
+
+	processToken := token
```
sboeuf: No need to use `processToken` here. You can modify directly `token` instead.

sboeuf:
LGTM
Let's wait for the last patch fixing the `docker exec` command and we'll be good to merge this.
As a follow up, I'll raise a PR implementing what's described [here](https://github.com/containers/virtcontainers/issues/518#issuecomment-358100194)

sameo:
coverall says we're at 63% with that PR...merging.

PR #577, opened 2018-01-19
shim: Carry the info about terminal

sboeuf:
Some shim implementation (like Kata) might need to know if a
terminal has been setup or not. This patch relies on Interactive
field from the Cmd structure that was never really used but
already holding this information.


jodh-intel:
lgtm

sameo:
LGTM

PR #583, opened 2018-01-25
virtcontainers: Global re-organisation of the proxy/shim code

### LINKED ISSUE #518: Rework of the proxy interface

sboeuf:
This PR mainly reworks the proxy interface of virtcontainers, pushes shim and proxy down the stack, that is hidden behind the agent.

Fixes #518

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -157,7 +157,7 @@ type agent interface {
 	stopPod(pod Pod) error
 
 	// createContainer will tell the agent to create a container related to a Pod.
-	createContainer(pod *Pod, c *Container) error
+	createContainer(pod *Pod, c *Container) (*Process, error)
 
 	// startContainer will tell the agent to start a container related to a Pod.
 	startContainer(pod Pod, c Container) error
--- a/agent.go
+++ b/agent.go
@@ -18,7 +18,6 @@ package virtcontainers
 
 import (
 	"fmt"
-	"path/filepath"
 	"syscall"
 
 	"github.com/mitchellh/mapstructure"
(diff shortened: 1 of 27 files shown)


sboeuf:
@egernst yes that's the plan !

sboeuf:
@jodh-intel @sameo PR is ready, tested and CI is passing, please review it. I am gonna re-upload an updated version for the commit message later.

jodh-intel:
@sboeuf - Wowzers - a biggie. I'll take a look at this tomorrow morning as didn't get round to this today I'm afraid.

jodh-intel:
@sboeuf - I like it! A simpler proxy interface *and* a simpler agent interface :)

lgtm

sameo:
LGTM

PR #584, opened 2018-01-26
Add support for scsi for hotplugging block devices

amshinde:
With this PR, block devices are hotplugged using either virtio-scsi or virtio-block based on a configuration.

### DIFF
--- a/container.go
+++ b/container.go
@@ -679,23 +679,24 @@ func (c *Container) hotplugDrive() error {
 		"fs-type":     fsType,
 	}).Info("Block device detected")
 
+	driveIndex, err := c.pod.getAndSetPodBlockIndex()
+	if err != nil {
+		return err
+	}
+
 	// Add drive with id as container id
 	devID := makeNameID("drive", c.id)
 	drive := Drive{
 		File:   devicePath,
 		Format: "raw",
 		ID:     devID,
+		Index:  driveIndex,
 	}
 
 	if err := c.pod.hypervisor.hotplugAddDevice(drive, blockDev); err != nil {
 		return err
 	}
 	c.setStateHotpluggedDrive(true)
-
-	driveIndex, err := c.pod.getAndSetPodBlockIndex()
-	if err != nil {
-		return err
-	}
 
 	if err := c.setStateBlockIndex(driveIndex); err != nil {
 		return err
--- a/device.go
+++ b/device.go
@@ -340,7 +340,11 @@ type BlockDevice struct {
 	DeviceType string
 	DeviceInfo DeviceInfo
 
-	// Path at which the device appears inside the VM, outside of the container mount namespace.
+	// SCSI Address of the block device, in case the device is attached using SCSI driver
+	// SCSI address is in the format SCSI-Id:LUN
+	SCSIAddr string
+
+	// Path at which the device appears inside the VM, outside of the container mount namespace
 	VirtPath string
 }
 
@@ -359,12 +363,6 @@ func (device *BlockDevice) attach(h hypervisor, c *Container) (err error) {
 
 	device.DeviceInfo.ID = hex.EncodeToString(randBytes)
 
-	drive := Drive{
-		File:   device.DeviceInfo.HostPath,
-		Format: "raw",
-		ID:     makeNameID("drive", device.DeviceInfo.ID),
-	}
-
 	// Increment the block index for the pod. This is used to determine the name
 	// for the block device in the case where the block device is used as container
 	// rootfs and the predicted block device name needs to be provided to the agent.
@@ -380,6 +378,13 @@ func (device *BlockDevice) attach(h hypervisor, c *Container) (err error) {
 		return err
 	}
 
+	drive := Drive{
+		File:   device.DeviceInfo.HostPath,
+		Format: "raw",
+		ID:     makeNameID("drive", device.DeviceInfo.ID),
+		Index:  index,
+	}
+
 	driveName, err := getVirtDriveName(index)
 	if err != nil {
 		return err
@@ -393,7 +398,17 @@ func (device *BlockDevice) attach(h hypervisor, c *Container) (err error) {
 
 	device.DeviceInfo.Hotplugged = true
 
-	device.VirtPath = filepath.Join("/dev", driveName)
+	if c.pod.config.HypervisorConfig.BlockDeviceDriver == VirtioBlock {
+		device.VirtPath = filepath.Join("/dev", driveName)
+	} else {
+		scsiAddr, err := getSCSIAddress(index)
+		if err != nil {
+			return err
+		}
+
+		device.SCSIAddr = scsiAddr
+	}
+
 	return nil
 }
 
--- a/hyperstart_agent.go
+++ b/hyperstart_agent.go
@@ -459,6 +459,7 @@ func (h *hyper) startOneContainer(pod Pod, c *Container) error {
 				Path:         d.DeviceInfo.ContainerPath,
 				AbsolutePath: true,
 				DockerVolume: false,
+				SCSIAddr:     d.SCSIAddr,
 			}
 			fsmap = append(fsmap, fsmapDesc)
 		}
--- a/hyperstart_agent.go
+++ b/hyperstart_agent.go
@@ -414,13 +414,22 @@ func (h *hyper) startOneContainer(pod Pod, c *Container) error {
 	container.SystemMountsInfo.BindMountDev = c.systemMountsInfo.BindMountDev
 
 	if c.state.Fstype != "" {
-		driveName, err := getVirtDriveName(c.state.BlockIndex)
-		if err != nil {
-			return err
+		// Pass a drive name only in case of block driver
+		if pod.config.HypervisorConfig.BlockDeviceDriver == VirtioBlock {
+			driveName, err := getVirtDriveName(c.state.BlockIndex)
+			if err != nil {
+				return err
+			}
+			container.Image = driveName
+		} else {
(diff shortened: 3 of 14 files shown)


device.go
```diff
@@ -340,7 +340,10 @@ type BlockDevice struct {
 	DeviceType string
 	DeviceInfo DeviceInfo
 
-	// Path at which the device appears inside the VM, outside of the container mount namespace.
+	// SCSI Address of the block device, in case the device is attached using SCSI driver
```
jodh-intel: Could you add details of the format here as you did in https://github.com/clearcontainers/agent/pull/205?

mount.go
```diff
@@ -404,3 +404,28 @@ func bindUnmountAllRootfs(sharedDir string, pod Pod) {
 		}
 	}
 }
+
+// getSCSIIdLun gets the SCSI id and lun, based on the index of the drive being inserted.
+// qemu code suggests that scsi-id can take values from 0 to 255 inclusive, while lun can
+// take values from 0 to 16383 inclusive. But lun values over 255 do not seem to follow
+// consistent SCSI addressing. Hence we limit to 255.
+func getSCSIIdLun(index int) (int, int, error) {
```
jodh-intel: Could you rename to `scsiIDLun` or similar (https://golang.org/doc/effective_go.html#Getters)?

mount.go
```diff
@@ -404,3 +404,28 @@ func bindUnmountAllRootfs(sharedDir string, pod Pod) {
...
+	}
+
+	if index > 65535 {
+		return -1, -1, fmt.Errorf("Index cannot be greater than 65535, maximum of 65536 devices are supported")
+	}
+
+	return index / 256, index % 256, nil
+}
+
+func getSCSIAddress(index int) (string, error) {
```
jodh-intel: Same comment - we don't need the `get` prefix.

mount_test.go
```diff
@@ -282,3 +283,52 @@ func TestGetVirtDriveName(t *testing.T) {
...
+		{0, 0, 0},
+		{1, 0, 1},
+		{2, 0, 2},
+		{255, 0, 255},
+		{256, 1, 0},
+		{257, 1, 1},
+		{258, 1, 2},
+		{512, 2, 0},
+		{513, 2, 1},
+	}
```
jodh-intel: Nice :)

hypervisor.go
```diff
@@ -154,6 +154,10 @@ type HypervisorConfig struct {
 	// DisableBlockDeviceUse disallows a block device from being used.
 	DisableBlockDeviceUse bool
 
+	// BlockDeviceDriver specifies the driver to be used for block device
+	// either "virtio-scsi" or "virtio-blk" with the default driver being "virtio-scsi"
```
jodh-intel: For maintenance, it's going to be safer if you refer to `VirtioBlock` and `VirtioSCSI`. But It could also be useful to define a `DefaultVirtio` which could also be mentioned here.

mount.go
amshinde: @jodh-intel This is not technically a getter for a struct field, but a function that derives the SCSI id and LUN based on an index.

device.go
amshinde: I have added details here.

hypervisor.go
amshinde: Fixed.

amshinde:
@sboeuf Can you take a look at this as well. CI is failing, likely due to lack of SCSI support on the agent side that needs to be merged first.

sboeuf:
@amshinde yes I'll review both PR (virtcontainers and agent) tomorrow !

mount.go
```diff
@@ -321,3 +321,28 @@ func bindUnmountAllRootfs(sharedDir string, pod Pod) {
...
+// getSCSIIdLun gets the SCSI id and lun, based on the index of the drive being inserted.
+// qemu code suggests that scsi-id can take values from 0 to 255 inclusive, while lun can
+// take values from 0 to 16383 inclusive. But lun values over 255 do not seem to follow
+// consistent SCSI addressing. Hence we limit to 255.
+func getSCSIIdLun(index int) (int, int, error) {
+	if index < 0 {
+		return -1, -1, fmt.Errorf("Index cannot be negative")
+	}
+
+	if index > 65535 {
```
sboeuf: Please use a constant for this.

qemu.go
```diff
@@ -976,22 +1002,30 @@ func (q *qemu) hotplugBlockDevice(drive Drive, op operation) error {
...
-		if err := q.removeDeviceFromBridge(drive.ID); err != nil {
-			return err
-		}
+			// Get SCSI-id and LUN based on the order of attaching drives.
+			scsiID, lun, err := getSCSIIdLun(drive.Index)
+			if err != nil {
+				return err
+			}
 
+			if err = q.qmpMonitorCh.qmp.ExecuteSCSIDeviceAdd(q.qmpMonitorCh.ctx, drive.ID, devID, driver, bus, scsiID, lun); err != nil {
```
sboeuf: The code looks good but I have noticed that you're not handling the case where we want to remove a SCSI device. Am I missing something about the way SCSI works, or have you forgotten to implement this ?

hyperstart_agent.go
```diff
@@ -419,13 +419,22 @@ func (h *hyper) startOneContainer(pod Pod, c *Container) error {
...
-			return err
+		// Pass a drive name only in case of block driver
+		if pod.config.HypervisorConfig.BlockDeviceDriver == VirtioBlock {
+			driveName, err := getVirtDriveName(c.state.BlockIndex)
+			if err != nil {
+				return err
+			}
+			container.Image = driveName
+		} else {
+			scsiAddr, err := getSCSIAddress(c.state.BlockIndex)
```
sboeuf: Very nice that we don't need to predict here ;)

pkg/hyperstart/types.go
```diff
@@ -207,7 +207,7 @@ type Container struct {
 	Rootfs           string              `json:"rootfs"`
 	Fstype           string              `json:"fstype,omitempty"`
 	Image            string              `json:"image"`
-	Addr             string              `json:"addr,omitempty"`
+	SCSIAddr         string              `json:"scsiAddr,omitempty"`
```
sboeuf: Why do you want to change the name ? `Addr` is generic enough to include SCSI case , right ?

device.go
```diff
@@ -340,7 +340,11 @@ type BlockDevice struct {
...
+	// SCSI address is in the format SCSI-Id:LUN
+	SCSIAddr string
```
sboeuf: Don't you think this should be part of the `DeviceInfo` ?

container.go
```diff
@@ -420,12 +420,14 @@ func createContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 		return nil, err
 	}
 
-	agentCaps := c.pod.agent.capabilities()
-	hypervisorCaps := c.pod.hypervisor.capabilities()
+	if !c.pod.config.HypervisorConfig.DisableBlockDeviceUse {
```
sboeuf: Please rebase, as this has been merged in a separate PR.

device.go
amshinde: No the DeviceInfo includes fields that are common to all devices. This is specific to Block device.

qemu.go
amshinde: The way to remove SCSI device is the same as virtio-block device.

pkg/hyperstart/types.go
amshinde: I had this as Addr in the agent, but changed this to SCSIAddr following your review comment which I agree as SCSIAddr makes it clearer. Hence the change here.

mount.go
amshinde: Done

container.go
amshinde: Rebased. Moved some code from qemu.go to qemu_arch_base.go introduced with arm support.

qemu.go
sboeuf: Makes sense, thanks !

pkg/hyperstart/types.go
sboeuf: I have forgotten...

device.go
sboeuf: Makes sense

qemu_arch_base.go
```diff
@@ -277,6 +285,16 @@ func (q *qemuArchBase) appendImage(devices []govmmQemu.Device, path string) ([]g
 	return q.appendBlockDevice(devices, drive), nil
 }
 
+func (q *qemu) appendSCSIController(devices []govmmQemu.Device, podConfig PodConfig) []govmmQemu.Device {
```
sboeuf: need to be a function of `qemuArchBase`

amshinde: Fixed.

sboeuf:
LGTM

amshinde:
@jodh-intel Can you take another look at this?

sboeuf:
@amshinde please update the agent commit sha1 in the runtime, this way, after you get your PR on the runtime merged, this PR should be re-triggered and pass.

mount.go
jodh-intel: Ah - fair point.

jodh-intel:
lgtm

amshinde:
@sboeuf I reran the CI after agent SHA update. The CI has passed now. Can we merge this?
This does introduce SCSI as the default storage driver, which means we need to add support for SCSI for kata agent as well.  We can change this to virtio-block in a follow up PR in the interim for kata.

sboeuf:
@amshinde yes perfect !

PR #586, opened 2018-01-29
qemu: Log default kernel params

### LINKED ISSUE #585: Log default kernel parameters in debug mode
Log the set of kernel parameters passed to the guest kernel in debug mode.

See: https://github.com/clearcontainers/runtime/pull/964

jodh-intel:
Add a debug-only log call that logs the default kernel parameters. This is
useful for users to identify which parameters are being set by virtcontainers,
which are set by the runtime and which are user-specified.

A new `defaultKernelParams()` function was added to return these default
parameters. This allow the default kernel parameters to be logged once prior
to hypervisor startup and before they are combined with the defaults passed by
the runtime.

Fixes #585.


qemu.go
```diff
@@ -765,6 +775,18 @@ func (q *qemu) createPod(podConfig PodConfig) error {
 
 // startPod will start the Pod's VM.
 func (q *qemu) startPod() error {
+	if q.config.Debug {
+		params := q.defaultKernelParams()
```
devimc: wait, ```q.kernelParams``` are not logged?

devimc: should the final kernel cmdline be logged too ?

jodh-intel: That actually already happens in the `govmm` package:

- https://github.com/intel/govmm/blob/master/qemu/qemu.go#L1535

It might seem a bit odd to log these params "again" (well, before the hypervisor starts), but it will allow users to know which of the kernel params that are eventually passed to govmm came from vc, the runtime and the user (config file).

devimc:
lgtm

devimc:
```BenchmarkCreateStartStopDeletePodQemuHypervisorHyperstartAgentNetworkCNI``` failed  o_O ?

jodh-intel:
Weird.

/me kicks the CI again...

sameo:
LGTM

PR #587, opened 2018-01-29
kata_agent: Unmount properly the container

### LINKED ISSUE #562: kata_agent: Unmount container's mounts from stopContainer()
We have to unmount everything that have been previously mounted in order to properly stop the container. This has to be done from `stopContainer()` function.

sboeuf:
Only the unmounting of the container rootfs was performed, leaving
behind all the other mounts related to the container.

Fixes #562

### DIFF
--- a/kata_agent.go
+++ b/kata_agent.go
@@ -565,6 +565,10 @@ func (k *kataAgent) stopContainer(pod Pod, c Container) error {
 		return err
 	}
 
+	if err := bindUnmountContainerMounts(c.mounts); err != nil {
+		return err
+	}
+
 	if err := bindUnmountContainerRootfs(kataHostSharedDir, pod.id, c.id); err != nil {
 		return err
 	}


sboeuf:
cc @sameo @jodh-intel

jodh-intel:
Good catch. Clearly we need a test for this (in https://github.com/clearcontainers/tests maybe?)

The change itself...

lgtm

grahamwhaley:
I think @chavafg had a plan to add some checks similar to the ones that can be found here:
https://github.com/clearcontainers/tests/blob/master/integration/stability/soak_parallel_rm.sh#L182
to some, or maybe all?, of the tests. If we add (or have) an 'init()' and 'teardown()' pair of functions already, then the tests can be added there.

Once we have the current tests running stably with `kata`, then we should rework that soak_parallel test to understand kata, and then give it a spin...

sameo:
LGTM

PR #589, opened 2018-01-30
container: fix semantic error when container state is modified

### LINKED ISSUE #588: the container returned by createContainer must be the same one contained in pod.containers
in order to avoid semantic errors modifying the container state of two containers with the same container ID, the container returned by `createContainer` must be the same one contained in `pod.containers`

devimc:
the container returned by createContainer must be the same contained
in pod.containers, this patch is needed to avoid semantic errors
manipulating references to containers with the same container ID, but
different memory address

fixes #588

### DIFF
--- a/container.go
+++ b/container.go
@@ -415,6 +415,13 @@ func createContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 	state, err := c.pod.storage.fetchContainerState(c.podID, c.id)
 	if err == nil && state.State != "" {
 		c.state.State = state.State
+		// if we return at this point, the container that is returned
+		// must match with the container in the pod (pod.containers)
+		for i := range pod.containers {
+			if c.id == pod.containers[i].id {
+				pod.containers[i] = c
+			}
+		}
 		return c, nil
 	}
 


jodh-intel:
Hi @devimc - could you add a test for this please?

devimc:
@jodh-intel yes

container.go
```diff
@@ -415,6 +415,13 @@ func createContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 	state, err := c.pod.storage.fetchContainerState(c.podID, c.id)
 	if err == nil && state.State != "" {
 		c.state.State = state.State
+		// if we return at this point, the container that is returned
+		// must match with the container in the pod (pod.containers)
+		for i := range pod.containers {
+			if c.id == pod.containers[i].id {
+				pod.containers[i] = c
+			}
+		}
```
sameo: It's still not clear to me what semantic errors we have when manipulating 2 containers built from the same container ID and `ContainerConfig`...

So for a given `ContainerConfig` and 2 semantically identical `Pod`s, the following sequence:

```Go

	c, err := newContainer(pod, contConfig)
	if err != nil {
		return nil, err
	}

	if err := c.createContainersDirs(); err != nil {
		return nil, err
	}

	state, err := c.pod.storage.fetchContainerState(c.podID, c.id)
	if err == nil && state.State != "" {
		c.state.State = state.State
        }
```

should produce identical containers. In other words, `pod.containers[i]` and c should already be semantically identical event though they may be different pointers. Moreover I'd like to understand the sequence that leads us to basically create a container pointer twice for a given `Pod`.
So I think this PR hides potentially 2 issues that we need to fix.

devimc: api.go
```go
func StartContainer(podID, containerID string) (VCContainer, error) {
    ....
    c, err := fetchContainer(p, containerID)
    if err != nil {
        return nil, err
    }
```
At this point `c` and `pod.containers[i]` are equal but not the same, they both have the same container ID, but they are in different memory space, if you modify `pod.containers[i]` state, that change won't be applied in `c`.
For example in `pod.updateResources` (#512) I change the number of vCPUs in `pod.containers[i]` and then I save the state, but that change is not applied in `c` (different memory space), so when `c` is started and the state is saved, the number of vCPUs is not honoured

```go
...
    // Before starting the container, POD's resources MUST BE updated                                                                                                       
    if err := p.updateResources(); err != nil {
        return nil, err
    }

    // Start it.                                                                                                                                                            
    err = c.start()
    if err != nil {
        c.delete()
        return nil, err
    }
```

for that reason, `c` and `pod.containers[i]` should be pointing to the same memory address.
from my point of view, we shouldn't have containers with the same ID but in different memory space

devimc: CI finished, so I think we are ok

sboeuf: I am checking the code right now, please let me some time to analyze this, I'll try to give you some feedback later.

sboeuf: Ok (I think) here is the solution to our problem:
```diff
diff --git a/container.go b/container.go
index 244b3b5..e0a5ead 100644
--- a/container.go
+++ b/container.go
@@ -244,18 +244,13 @@ func fetchContainer(pod *Pod, containerID string) (*Container, error) {
 		return nil, errNeedContainerID
 	}
 
-	fs := filesystem{}
-	config, err := fs.fetchContainerConfig(pod.id, containerID)
-	if err != nil {
-		return nil, err
-	}
-
-	container, err := createContainer(pod, config)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create container with config %v in pod %v: %v", config, pod.id, err)
+	for _, c := range pod.containers {
+		if containerID == c.id {
+			return c, nil
+		}
 	}
 
-	return container, nil
+	return nil, fmt.Errorf("Could not find the container in the pod's containers list")
 }
 
 // storeContainer stores a container config.
@@ -356,7 +351,8 @@ func newContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 	return c, nil
 }
 
-// createContainer creates and start a container inside a Pod.
+// createContainer creates and start a container inside a Pod. It has to be
+// called only when a new container, not known by the pod, has to be created.
 func createContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 	if pod == nil {
 		return nil, errNeedPod
@@ -371,17 +367,12 @@ func createContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 		return nil, err
 	}
 
-	state, err := c.pod.storage.fetchContainerState(c.podID, c.id)
-	if err == nil && state.State != "" {
-		c.state.State = state.State
-		return c, nil
+	// Add the container to the containers list in the pod.
+	if err := pod.addContainer(c); err != nil {
+		return nil, err
 	}
 
-	// If we reached that point, this means that no state file has been
-	// found and that we are in the first creation of this container.
-	// We don't want the following code to be executed outside of this
-	// specific case.
-	process, err := c.pod.agent.createContainer(c.pod, c)
+	process, err := pod.agent.createContainer(c.pod, c)
 	if err != nil {
 		return nil, err
 	}
--
```

So basically we need 2 things here:
- First, we need to add a call to `pod.addContainer(c)` into `createContainer()` (in `container.go`), so that we actually add the newly created container to the containers list from the `pod` structure.
- Second, I have checked everywhere and I can confirm that `fetchContainer()` is always called from the high level API (in `api.go`) after a call to `fetchPod()`. Because `fetchPod()` rebuild everything from scratch/config (pod and then containers), there is no need for `fetchContainer()` to recreate a new container structure and fill it with info from the storage. Instead, this function is now super simple as it returns the container pointer held by the pod containers list.
@sameo @devimc Let me know if this makes sense to you !

devimc: @sboeuf closing this pr, please open a new one

sboeuf: Ok, will do in a bit !

PR #592, opened 2018-01-30
container: Fix container semantics

### LINKED ISSUE #588: the container returned by createContainer must be the same one contained in pod.containers

sboeuf:
This pull request ensure the container pointers held by the `pod` structure are not duplicated. This is needed to avoid a container pointer modification not to be taken into account because the wrong pointer is used.

Fixes #588

### DIFF
--- a/container.go
+++ b/container.go
@@ -332,7 +332,8 @@ func newContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 	return c, nil
 }
 
-// createContainer creates and start a container inside a Pod.
+// createContainer creates and start a container inside a Pod. It has to be
+// called only when a new container, not known by the pod, has to be created.
 func createContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 	if pod == nil {
 		return nil, errNeedPod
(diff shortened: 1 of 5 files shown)


sboeuf:
cc @devimc @sameo

jodh-intel:
Hi @sboeuf - please can you add some unit tests for this change.

devimc:
lgtm

sboeuf:
@jodh-intel I have added unit tests in order to cover `fetchContainer()` function since this is the one that matters here. It makes sure we retrieve the container pointer coming from the list of containers held by the pod structure.

container.go
```diff
@@ -244,18 +244,13 @@ func fetchContainer(pod *Pod, containerID string) (*Container, error) {
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to create container with config %v in pod %v: %v", config, pod.id, err)
+	for _, c := range pod.containers {
+		if containerID == c.id {
+			return c, nil
+		}
 	}
 
-	return container, nil
+	return nil, fmt.Errorf("Could not find the container in the pod's containers list")
```
jodh-intel: Could you add `containerID` to the returned error (and maybe `pod.ID()`)?

sboeuf: Oh yeah sorry I forgot that ! Done !

jodh-intel:
Thanks @sboeuf.

lgtm

egernst:
Nice.

LGTM

container.go
```diff
@@ -244,18 +244,14 @@ func fetchContainer(pod *Pod, containerID string) (*Container, error) {
...
-		return nil, err
-	}
-
-	container, err := createContainer(pod, config)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create container with config %v in pod %v: %v", config, pod.id, err)
+	for _, c := range pod.containers {
+		if containerID == c.id {
+			return c, nil
+		}
```
sameo: A couple of things on that one:

* This needs a comment.
* Having to do so means we're fetching a container from storage while having a fully built pod with all its container already fetched and added. When is that happening ?

sboeuf:
@sameo comments addressed, I have replaced the function `fetchContainer()` with `findContainer()`, which is more appropriate about the naming.

sameo:
LGTM

Issue #579, opened 2018-01-23
add support for ipv6


egernst:
Enhance virtcontainers as well as our agent protocol and agent to support discovery and configuration of IPv6.

egernst:
This issue was moved to kata-containers/runtime#147

PR #595, opened 2018-02-06
kata-agent: add network gRPC support

egernst:
Introduce support for setting up interface and routes when working with
Kata agent.

Signed-off-by: Eric Ernst <eric.ernst@intel.com>

### DIFF
--- a/kata_agent.go
+++ b/kata_agent.go
@@ -300,6 +300,83 @@ func (k *kataAgent) exec(pod *Pod, c Container, cmd Cmd) (*Process, error) {
 	return prepareAndStartShim(pod, k.shim, c.id, req.ExecId, k.state.URL, cmd)
 }
 
+func (k *kataAgent) generateInterfacesAndRoutes(networkNS NetworkNamespace) ([]*grpc.Interface, []*grpc.Route, error) {
+
+	if networkNS.NetNsPath == "" {
+		return nil, nil, nil
+	}
+
+	var routes []*grpc.Route
+	var ifaces []*grpc.Interface
+
+	for _, endpoint := range networkNS.Endpoints {
+
+		var ipAddresses []*grpc.IPAddress
+		for _, addr := range endpoint.Properties().Addrs {
+			// Skip IPv6 because not supported
+			if addr.IP.To4() == nil {
+				// Skip IPv6 because not supported
+				k.Logger().WithFields(logrus.Fields{
+					"unsupported-address-type": "ipv6",
+					"address":                  addr,
+				}).Warn("unsupported address")
+				continue
+			}
+			// Skip localhost interface
+			if addr.IP.IsLoopback() {
+				continue
+			}
+			netMask, _ := addr.Mask.Size()
+			ipAddress := grpc.IPAddress{
+				Family:  grpc.IPFamily_v4,
+				Address: addr.IP.String(),
+				Mask:    fmt.Sprintf("%d", netMask),
+			}
+			ipAddresses = append(ipAddresses, &ipAddress)
+		}
+		ifc := grpc.Interface{
+			IPAddresses: ipAddresses,
+			Device:      endpoint.Name(),
+			Name:        endpoint.Name(),
+			Mtu:         uint64(endpoint.Properties().Iface.MTU),
+			HwAddr:      endpoint.HardwareAddr(),
+		}
+
+		ifaces = append(ifaces, &ifc)
+
+		for _, route := range endpoint.Properties().Routes {
+			var r grpc.Route
+
+			if route.Dst != nil {
+				r.Dest = route.Dst.String()
+
+				if route.Dst.IP.To4() == nil {
+					// Skip IPv6 because not supported
+					k.Logger().WithFields(logrus.Fields{
+						"unsupported-route-type": "ipv6",
+						"destination":            r.Dest,
+					}).Warn("unsupported route")
+					continue
+				}
+			}
+
+			if route.Gw != nil {
+				r.Gateway = route.Gw.String()
+			}
+
+			if route.Src != nil {
+				r.Source = route.Src.String()
+			}
+
+			r.Device = endpoint.Name()
+			r.Scope = uint32(route.Scope)
+			routes = append(routes, &r)
+
+		}
+	}
+	return ifaces, routes, nil
+}
+
 func (k *kataAgent) startPod(pod Pod) error {
 	if k.proxy == nil {
 		return errorMissingProxy
@@ -333,6 +410,45 @@ func (k *kataAgent) startPod(pod Pod) error {
 	hostname := pod.config.Hostname
 	if len(hostname) > maxHostnameLen {
 		hostname = hostname[:maxHostnameLen]
+	}
+
+	//
+	// Setup network interfaces and routes
+	//
+	interfaces, routes, err := k.generateInterfacesAndRoutes(pod.networkNS)
+	if err != nil {
+		return err
+	}
+	for _, ifc := range interfaces {
+		// send update interface request
+		ifcReq := &grpc.UpdateInterfaceRequest{
+			Interface: ifc,
+		}
+		resultingInterface, err := k.sendReq(ifcReq)
+		if err != nil {
+			k.Logger().WithFields(logrus.Fields{
+				"interface-requested": fmt.Sprintf("%+v", ifc),
+				"resulting-interface": fmt.Sprintf("%+v", resultingInterface),
+			}).WithError(err).Error("update interface request failed")
+			return err
+		}
+	}
+
+	if routes != nil {
+		routesReq := &grpc.UpdateRoutesRequest{
+			Routes: &grpc.Routes{
+				Routes: routes,
+			},
+		}
+
+		resultingRoutes, err := k.sendReq(routesReq)
+		if err != nil {
+			k.Logger().WithFields(logrus.Fields{
+				"routes-requested": fmt.Sprintf("%+v", routes),
+				"resulting-routes": fmt.Sprintf("%+v", resultingRoutes),
+			}).WithError(err).Error("update routes request failed")
+			return err
+		}
 	}
 
 	// We mount the shared directory in a predefined location
@@ -659,6 +775,12 @@ func (k *kataAgent) sendReq(request interface{}) (interface{}, error) {
 	case *grpc.SignalProcessRequest:
 		_, err := k.client.SignalProcess(context.Background(), req)
 		return nil, err
+	case *grpc.UpdateRoutesRequest:
+		_, err := k.client.UpdateRoutes(context.Background(), req)
+		return nil, err
+	case *grpc.UpdateInterfaceRequest:
+		ifc, err := k.client.UpdateInterface(context.Background(), req)
+		return ifc, err
 	default:
 		return nil, fmt.Errorf("Unknown gRPC type %T", req)
 	}
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -77,7 +77,7 @@
     "protocols/client",
     "protocols/grpc"
   ]
-  revision = "3f6f6da6d469c31364aac417f336ec0d16772095"
+  revision = "11f8b0ad0a3bb5771b88cbccd0a9c7dca5fd8e56"
 
 [[projects]]
   name = "github.com/kubernetes-incubator/cri-o"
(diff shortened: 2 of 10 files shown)


egernst:
Functioning network gRPC calls dependent  on https://github.com/kata-containers/agent/pull/126

egernst:
This is dependent on the gRPC changes landing in kata-containers/agent and being re-vendorded.

kata_agent.go
```diff
@@ -295,6 +295,82 @@ func (k *kataAgent) exec(pod *Pod, c Container, cmd Cmd) (*Process, error) {
...
+	networkNS, err := pod.storage.fetchPodNetwork(pod.id)
+	if err != nil {
+		return err
+	}
+
+	if networkNS.NetNsPath == "" {
+		return nil
+	}
+
+	for _, endpoint := range networkNS.Endpoints {
```
jodh-intel: Random thought: this code is pretty similar to:

- https://github.com/containers/virtcontainers/blob/master/hyperstart_agent.go#L173

I wonder if it would be worth refactoring to make a generic function in `agent.go` that accepted a `Pod` and a function to operate on the `Endpoint`'s?

kata_agent.go
```diff
@@ -295,6 +295,82 @@ func (k *kataAgent) exec(pod *Pod, c Container, cmd Cmd) (*Process, error) {
...
+		var route grpc.Route
+		var ipAddresses []*grpc.IpAddress
+		for _, addr := range endpoint.Properties().Addrs {
+			// Skip IPv6 because not supported
```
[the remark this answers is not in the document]
egernst: Yes: https://github.com/containers/virtcontainers/issues/579

kata_agent.go
egernst: It is similar in that it loops over the endpoints and addresses, but with the protocols being different the implementation is different enough (operating on an interface and route at a time) that I think it may be a bit painful for a potentially short term code reuse benefit.  WDYT?

kata_agent.go
```diff
@@ -330,6 +406,10 @@ func (k *kataAgent) startPod(pod Pod) error {
 		hostname = hostname[:maxHostnameLen]
 	}
 
+	if err := k.sendNetworkInterfaceAndRouteRequests(pod); err != nil {
```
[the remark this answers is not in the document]
egernst: Hey @jodh-intel - I started to do this a couple of times and am having a hard time.  The current "mocking" in kata_agent_test.go is pretty limited (return nil, nil), so it is pretty hard to exercise without re-implementing a lot of logic from the existing kata agent in the test file.  I now recall all of the mock discussions that you were driving back in the CC-3.0 pre-release time frame.  

Let me think on how to handle this pragmatically.  Let me know if you have suggestion in the meantime.

egernst:
Tested with latest agent and this PR now includes revendoring for latest gRPC changes.  @sboeuf @jodh-intel  can you take a look?  Would be good to get working networking in...

sboeuf:
@egernst I'll review it today !

kata_agent.go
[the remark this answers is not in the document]
jodh-intel: I'm not sure I understand. We don't necessarily need 100% coverage but this PR adds a relatively big function with no tests. I don't think we should be attempting to land significant features without tests.

The approach I'd take is to break `sendNetworkInterfaceAndRouteRequests()` into smaller functions (for improved clarity, code complexity and testing) and you should then be able to unit-test most of those. But even as it is, can't you create test functions to atleast handle?:

- `fetchPodNetwork()` failing.
- `networkNS.NetNsPath == ""`.
- `addr.IP.IsLoopback() == true` (force it with `link, err := netlink.LinkByName("lo")` ?)

jodh-intel:
Yep - would be great to get networking in, once we have tests for it.

kata_agent.go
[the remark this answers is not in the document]
egernst: Yes, you are right. I will refactor and add tests today.

kata_agent.go
```diff
@@ -335,6 +410,45 @@ func (k *kataAgent) startPod(pod Pod) error {
 		hostname = hostname[:maxHostnameLen]
 	}
 
+	//
+	// Setup network interfaces and routes
+	//
+	networkNS, err := pod.storage.fetchPodNetwork(pod.id)
```
sboeuf: Note: This should be part of the `pod` and we should be able to access those infos from in-memory. Opened an issue here: https://github.com/containers/virtcontainers/issues/628
Will be fixed in another PR

kata_agent.go
```diff
@@ -335,6 +410,45 @@ func (k *kataAgent) startPod(pod Pod) error {
...
+	if err != nil {
+		return err
+	}
+	interfaces, routes, err := k.generateKataInterfacesAndRoutes(networkNS)
+	if err != nil {
+		return err
+	}
```
sboeuf: Please add a blank line to let the code breathe :smile:
And globally, this comment applies to every place in the code where there are not enough blank lines (for readability)

kata_agent.go
```diff
@@ -335,6 +410,45 @@ func (k *kataAgent) startPod(pod Pod) error {
...
+		ifcReq := &grpc.UpdateInterfaceRequest{
+			Interface: ifc,
+		}
+		resultingInterface, err := k.sendReq(ifcReq)
+		if err != nil {
+			k.Logger().WithFields(logrus.Fields{
+				"interface requested": fmt.Sprintf("%+v", ifc),
+				"resulting interface": fmt.Sprintf("%+v", resultingInterface),
+			}).WithError(err).Error("update interface request failed")
+			continue
```
sboeuf: Why don't we fail here ? Have we agreed on ignoring failures here ?

kata_agent.go
```diff
@@ -335,6 +410,45 @@ func (k *kataAgent) startPod(pod Pod) error {
...
+		//send add route request
+		routeReq := &grpc.AddRouteRequest{
+			Route: route,
+		}
+		_, err = k.sendReq(routeReq)
+		if err != nil {
+			k.Logger().WithFields(logrus.Fields{
+				"route requested": fmt.Sprintf("%+v", route),
+			}).WithError(err).Error("update route request failed")
+			continue
```
sboeuf: Why don't we fail here ? Have we agreed on ignoring failures here ?

kata_agent.go
```diff
@@ -300,6 +300,81 @@ func (k *kataAgent) exec(pod *Pod, c Container, cmd Cmd) (*Process, error) {
...
+			}
+			netMask, _ := addr.Mask.Size()
+			ipAddress := grpc.IPAddress{
+				Family:  grpc.IPFamily_v4,
+				Address: addr.IP.String(),
+				Mask:    fmt.Sprintf("%d", netMask),
+			}
+			ipAddresses = append(ipAddresses, &ipAddress)
+		}
+		ifc := grpc.Interface{
```
sboeuf: You can do
```go
ifc := &grpc.Interface{
```
so that later in the code you can do
```go
ifaces = append(ifaces, ifc)
```

kata_agent.go
```diff
@@ -300,6 +300,81 @@ func (k *kataAgent) exec(pod *Pod, c Container, cmd Cmd) (*Process, error) {
...
+			IPAddresses: ipAddresses,
+			Device:      endpoint.Name(),
+			Name:        endpoint.Name(),
```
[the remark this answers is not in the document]
egernst: Done: https://github.com/kata-containers/agent/issues/147

kata_agent.go
egernst: ack

kata_agent.go
egernst: Will add some air.

kata_agent.go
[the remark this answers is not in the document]
sboeuf: Thanks !

kata_agent.go
egernst: I agree on logging the error and continuing to apply remaining interfaces and routes.  I think that this is an error, but not necessarily catastrophic -- user can get into the workload and then assess what failed in setup.  WDYT?

kata_agent.go
egernst: yawn.  Sure. :)

kata_agent.go
sboeuf: Well, only logging without returning this error might be confusing since the user will expect everything to work as expected if he does not check the logs. And he might not understand why he does not get what he expects inside the container.
We could have a retry mechanism (maybe retry 5 times with `100ms` between each time), so that we really make sure there is a real issue here, that would worth throwing an error.

kata_agent.go
egernst: done.

egernst:
@sboeuf -- updated.

kata_agent.go
sboeuf: @sameo @jodh-intel WDYT ?

sameo: @egernst @sboeuf If an interface or a route failed to be set, it's likely that networking will be at least partly dysfunctional, leaving the end-user with a potentially non working pod. By only logging this and not failing we'd give a broken pod to the upper layers of the stack and pretend it's all fine, which is bad from a end user experience perspective. In other words, we should log **and** fail here and the pod must not be passed upward. It should be stopped at least.

jodh-intel: +1 for failing - if we continue, surely the system is in an unknown state (bad)? Turning it round, do we ever think this might fail? If so, why?

kata_agent.go
```diff
@@ -300,6 +300,81 @@ func (k *kataAgent) exec(pod *Pod, c Container, cmd Cmd) (*Process, error) {
...
+					k.Logger().WithFields(logrus.Fields{
+						"unsupported-route-type": "ipv6",
+						"destination":            route.Dest,
+					}).Warn("unsupported route")
+					continue
+				}
+			}
+
+			route.Gateway = r.Gw.String()
+			if route.Gateway == "<nil>" {
```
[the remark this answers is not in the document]
egernst: Agreed

kata_agent.go
egernst: Agreed. Thanks guys.

egernst: I should have stopped with error in the first place, mea culpa.  It won't work today as I didn't handle conflicting routes (for the initial interface route in the guest).    I have a new PR, https://github.com/kata-containers/agent/pull/148, R on the agent that I started (not ready just yet) which will act on a set of routes instead, acting fully declarative.  Without this I don't think we can really test networking in a container without bypassing conflict errors (because the initial interface route already exists).  This'll simplify our handling quite a bit --   I'll try to get this ready for merge today.

egernst:
see https://github.com/kata-containers/agent/pull/148

kata_agent.go
```diff
@@ -300,6 +300,81 @@ func (k *kataAgent) exec(pod *Pod, c Container, cmd Cmd) (*Process, error) {
 	return prepareAndStartShim(pod, k.shim, c.id, req.ExecId, k.state.URL, cmd)
 }
 
+func (k *kataAgent) generateKataInterfacesAndRoutes(networkNS NetworkNamespace) ([]*grpc.Interface, []*grpc.Route, error) {
```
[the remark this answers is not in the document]
egernst: apparently I really like to push the kata brand, in this case in error.  Will update...

egernst:
waiting for kata-containers/agent#148 to land so I can update the revendoring.

In the meantime, addressed feedback from @sameo and @jodh-intel and @sboeuf .

egernst:
@sboeuf  PTAL?

kata_agent.go
sboeuf: You can update this cause you don't need to fetch the pod network anymore since #628 has been solved by #629

egernst:
@jodh-intel @sboeuf  - updated based on updated pod structure.  PTAL

sboeuf:
LGTM

kata_agent.go
```diff
@@ -335,6 +412,45 @@ func (k *kataAgent) startPod(pod Pod) error {
...
+	}
+	for _, ifc := range interfaces {
+		// send update interface request
+		ifcReq := &grpc.UpdateInterfaceRequest{
+			Interface: ifc,
+		}
+		resultingInterface, err := k.sendReq(ifcReq)
+		if err != nil {
+			k.Logger().WithFields(logrus.Fields{
+				"interface requested": fmt.Sprintf("%+v", ifc),
```
jodh-intel: Nit: there is an informal understanding [*] that all log fields should be single-word (so I'd just replace spaces with hyphens here and below).

---

[*] - yep, we should make expectations formal by writing a doc! ;)

sameo:
LGTM

sameo:
@sboeuf @egernst Thanks for the PR. Let's merge it once CI passes.

PR #609, opened 2018-02-09
pod: Improve code performance and clarity

sboeuf:
This PR basically removes a bunch of not needed calls to fetch functions. This has two effects:
- It makes the code much clearer since we know that we maintain proper state in-memory and on-disk, with no need to fetch from the disk every time we want the state of a pod or a container
- It improves performance on the host by limiting the accesses to the filesystem.


jodh-intel:
lgtm

sameo:
LGTM

PR #616, opened 2018-02-12
kata_agent: Support block device passthrough

sboeuf:
This patch enables the block device passthrough for Kata Containers
agent. It gives the driver field "blk" so that the agent can actually
know what to do with the passed device.

### DIFF
--- a/kata_agent.go
+++ b/kata_agent.go
@@ -48,6 +48,7 @@ var (
 	type9pFs                    = "9p"
 	devPath                     = "/dev"
 	vsockSocketScheme           = "vsock"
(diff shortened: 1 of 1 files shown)


sboeuf:
Depends on https://github.com/kata-containers/agent/pull/133

sameo:
Shouldn't we export the driver string from kata-containers/agent ?
Should be part of a different PR, obviously.

For this one:
LGTM

sboeuf:
This patch is included into a more global PR about block devices and devicemapper [here](https://github.com/containers/virtcontainers/pull/626), I can close this PR now.

PR #617, opened 2018-02-12
container: Let some time before to force kill a container process

### LINKED ISSUE #606: Race condition with CRI-O
I have discovered that we have a bug about the way we work with CRI-O. Specifically this happens when CRI-O tries to stop a running container:
```bash
POD_ID=$(sudo crictl runp test/testdata/sandbox_config.json)
CONTAINER_ID=$(sudo crictl create $POD_ID test/testdata/container_redis.json test/testdata/sandbox_config.json)
sudo crictl start $CONTAINER_ID

sudo crictl stop $CONTAINER_ID
```
First you need to know that CRI-O stop command relates to a `kill` command from a runtime perspective, which translates into `cc-runtime kill` in our case.

Then, you also need to know that CRI-O expect to be able to unmount the storage (rootfs) provided at the `cc-runtime create` stage, after the `cc-runtime kill` returned. This is logical if we think about the `runc` case where we can expect to be able to unmount everything after the container process has been terminated. Here is the [call1](https://github.com/kubernetes-incubator/cri-o/blob/master/lib/stop.go#L10) which calls into [call2](https://github.com/kubernetes-incubator/cri-o/blob/master/oci/oci.go#L608). Here is the [call](https://github.com/kubernetes-incubator/cri-o/blob/master/lib/stop.go#L27) expecting to unmount.

Now you understand that we have an issue since we don't stop our container, which means in our case that we don't unmount what has been previously mounted. Take a look at [this](https://github.com/clearcontainers/runtime/blob/master/kill.go#L100-L120) to understand that we only send a signal in case we receive a `cc-runtime kill`, nothing else.

But why is it working then ? That's a good question, and here is the answer:

We have a mechanism that check a container status and the state of the container process (check shim process basically). This mechanism is triggered every time someone ask for the status of the container, and in case of our runtime, all our actions asks for this, but this can be basically achieved by calling into `cc-runtime state` too. And in case the container is marked as RUNNING, but the shim process is not here anymore, we actually update the container status by [calling](https://github.com/containers/virtcontainers/blob/master/api.go#L575-L595) into `container.stop` [here](https://github.com/containers/virtcontainers/blob/master/api.go#L590). This has 2 effects, it stops the container by calling into the agent, and by unmounting everything needed, and it updates the container status to STOPPED.

But you wonder who is calling into this so that we can get the container unmounted by luck. Well, CRI-O has a watcher running in a separate goroutine which monitors the container process. When this container process returns (after the kill has been sent by CRI-O), the watcher will trigger a call to `cc-runtime state` [here](https://github.com/kubernetes-incubator/cri-o/blob/master/server/server.go#L389), and this is how we eventually get the container STOPPED and properly unmounted almost at the same time.

And now you wonder why we don't get a more obvious race condition since the call to unmount from CRI-O happens before the update of the status. The reason is that the unmount is pretty flexible. It allows for the unmount to last up to 20 seconds. I don't know about the internals of devicemapper here, but I have confirmed this behavior by putting a 30 seconds sleep call between those [2 lines](https://github.com/kubernetes-incubator/cri-o/blob/master/server/server.go#L388-L389).
It issues an error after 20 seconds saying that the device unmount/remove could not be done because it is still busy:
```
Stopping the container failed: rpc error: code = Unknown desc = failed to unmount container bb325756e1c246f0c7b6639f17d84b841c46bc31ea638f0df716ae6e38425c13: Device is Busy
```

This is a behavior that we have seen sometimes on the CI and we always wondered why we couldn't reproduce locally. The answer is that the CI is much slower because of the nested virtualization and it sometimes hit those timeouts.

Now the question is how do we solve this ?

Well I think we have 2 options:
- Either we consider we should translate a `cc-runtime kill` into `vc.StopContainer` in case the signal is `SIGTERM` or `SIGKILL`.
- Or we ask CRI-O folks to fix this into CRI-O. It would honestly make more sense into CRI-O IMO. I think CRI-O should have the `UpdateStatus()` being called right after they stopped the container. This way we would get `cc-runtime state` being called  BEFORE they try to unmount the storage. But from their perspective, it would make sense too. They should not try to unmount/remove the storage after they sent a kill, they should actually make sure the container is properly stopped before to go further.

sboeuf:
Using force kill when we want to stop a container should only be
called after we left some time to make sure nothing has been killing
the container process. Indeed, signals can sometime take a few seconds
to terminate a process. In this case, we don't want to SIGKILL this
process, otherwise it would return with the exit code 137. Instead, we
want to leave it some time (10 seconds) so that it can properly end.

After those 10 seconds, we consider the previous signal didn't get the
expected effect and we force kill the container process. Indeed, we
still want to try to kill the container.

Fixes #606

Depends-on: github.com/clearcontainers/runtime#1003

### DIFF
--- a/shim.go
+++ b/shim.go
@@ -41,7 +41,7 @@ const (
 	KataShimType ShimType = "kataShim"
 )
 
-var waitForShimTimeout = 5.0
+var waitForShimTimeout = 10.0
 var consoleFileMode = os.FileMode(0660)
 
 // ShimParams is the structure providing specific parameters needed
--- a/container.go
+++ b/container.go
@@ -555,7 +555,7 @@ func (c *Container) stop() error {
 		// Make sure we stop the shim process
 		if running, _ := isShimRunning(c.process.Pid); running {
 			l := c.Logger()
-			l.Warn("Failed to stop container so stopping dangling shim")
+			l.Error("Failed to stop container so stopping dangling shim")
 			if err := stopShim(c.process.Pid); err != nil {
 				l.WithError(err).Warn("failed to stop shim")
 			}
@@ -563,17 +563,22 @@ func (c *Container) stop() error {
 
 	}()
 
-	ctrRunning, err := isShimRunning(c.process.Pid)
-	if err != nil {
-		return err
-	}
-
-	if ctrRunning {
+	// Here we expect that stop() has been called because the container
+	// process returned or because it received a signal. In case of a
+	// signal, we want to give it some time to end the container process.
+	// However, if the signal didn't reach its goal, the caller still
+	// expects this container to be stopped, that's why we should not
+	// return an error, but instead try to kill it forcefully.
+	if err := waitForShim(c.process.Pid); err != nil {
+		// Force the container to be killed.
 		if err := c.pod.agent.killContainer(*(c.pod), *c, syscall.SIGKILL, true); err != nil {
 			return err
 		}
 
-		// Wait for the end of container
+		// Wait for the end of container process. We expect this call
+		// to succeed. Indeed, we have already given a second chance
+		// to the container by trying to kill it with SIGKILL, there
+		// is no reason to try to go further if we got an error.
 		if err := waitForShim(c.process.Pid); err != nil {
 			return err
 		}


sameo:
LGTM

shim.go
```diff
@@ -41,7 +41,7 @@ const (
 	KataShimType ShimType = "kataShim"
 )
 
-var waitForShimTimeout = 5.0
+var waitForShimTimeout = 10.0
```
grahamwhaley: I was wondering if this is configurable via the API and therefore optionally through the runtime toml config for instance. Just a thought.

PR #620, opened 2018-02-14
kata_agent: Support CRI-O with overlay

sboeuf:
This pull request enables CRI-O to work with Kata Containers in case the storage driver used is overlay, meaning we rely on 9pfs to pass the rootfs inside the VM.

### DIFF
--- a/kata_agent.go
+++ b/kata_agent.go
@@ -414,6 +414,10 @@ func constraintGRPCSpec(grpcSpec *grpc.Spec) {
 	// here: https://github.com/kata-containers/agent/issues/104
 	grpcSpec.Linux.Seccomp = nil
 
+	// TODO: Remove this constraint as soon as the agent properly handles
+	// resources provided through the specification.
+	grpcSpec.Linux.Resources = nil
+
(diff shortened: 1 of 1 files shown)


sboeuf:
@chavafg could you check if you get CRI-O working for Kata with this PR ? Only with overlay.

sameo:
LGTM

kata_agent.go
```diff
@@ -414,6 +414,10 @@ func constraintGRPCSpec(grpcSpec *grpc.Spec) {
 	// here: https://github.com/kata-containers/agent/issues/104
 	grpcSpec.Linux.Seccomp = nil
 
+	// TODO: Remove this constraint as soon as the agent properly handles
+	// resources provided through the specification.
+	grpcSpec.Linux.Resources = nil
```
grahamwhaley: Do we have an appropriate Issue open so we don't forget this?
Or maybe a `FIXME` in the code might suffice?

sboeuf: Here it is https://github.com/kata-containers/agent/issues/144

jodh-intel:
lgtm

PR #629, opened 2018-02-21
pod: Fetch pod network only from createPod()

### LINKED ISSUE #628: network: Make the pod retain info about network
We don't want to fetch from the filesystem every time we need network info about the pod. This fetch should happen only once when the pod is created, and we should store on disk every time a change happens. The in-memory info about the pod should be consistent so that we can rely on them without fetching them from the filesystem.

sboeuf:
We don't want to fetch the pod network from the filesystem every
time we need information about it. Instead, the network is related
to the pod, that's why the pod should hold this information in-memory
through a new field into its structure.

Fixes #628

### DIFF
--- a/api.go
+++ b/api.go
@@ -63,13 +63,13 @@ func createPodFromConfig(podConfig PodConfig) (*Pod, error) {
 	}
 
 	// Add the network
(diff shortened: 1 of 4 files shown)


sameo:
LGTM

PR #630, opened 2018-02-21
documentation: Document the 1.0 API

sameo:
The virtcontainers godoc [1] does not highlight the 1.0 top level API
in a readable and clear way. This API documentation attempts to
provide a top to bottom description of the virtcontainers 1.0 API.

[1] https://godoc.org/github.com/containers/virtcontainers

### DIFF
--- a/documentation/api/1.0/api.md
+++ b/documentation/api/1.0/api.md
@@ -547,6 +547,39 @@ func ResumePod(podID string) (VCPod, error)
 
 ## Container API
 
+The virtcontainers 1.0 container API manages pod
+[container lifecycles](#container-functions).
+
+A virtcontainers container is process running inside a containerized
+environment, as part of a hardware virtualized context. In other words,
+a virtcontainers container is just a regular container running inside a
+virtual machine's guest OS.
+
+A virtcontainers container always belong to one and only one
+virtcontainers pod, again following the
+[Kubernetes](https://kubernetes.io/docs/concepts/workloads/pods/pod-overview/)
+logic and semantics.
+
+The container API allows callers to [create](#createcontainer),
+[delete](#deletecontainer), [start](#startcontainer), [stop](#stopcontainer),
+[kill](#killcontainer) and [observe](#statuscontainer) containers. It also
+allows for running [additional processes](#entercontainer) inside a
+specific container.
+
+As a virtcontainers container is always linked to a pod, the entire container
+API always takes a pod ID as its first argument.
+
+To create a container, the API caller must prepare a
+[`ContainerConfig`](#containerconfig) and pass it to
+[`CreateContainer`](#createcontainer) together with a pod ID. Upon successful
+container creation, the virtcontainers API will return a
+[`VCContainer`](#vccontainer) interface back to the caller.
+
+The `VCContainer` interface is a container abstraction hiding the internal
+and private virtcontainers container structure. It is a handle for API callers
+to manage the container lifecycle through the rest of the
+[container API](#container-functions).
+
 * [Structures](#container-structures)
 * [Functions](#container-functions)
 
--- a/documentation/api/1.0/api.md
+++ b/documentation/api/1.0/api.md
@@ -1,0 +1,810 @@
+# virtcontainers 1.0 API
+
+* [Pod API](#pod-api)
+* [Container API](#container-api)
+* [Examples](#examples)
+
+## Pod API
+
+* [Structures](#pod-structures)
+* [Functions](#pod-functions)
+
+### Pod Structures
+
+* [PodConfig](#podconfig)
+  * [Resources](#resources)
+  * [HypervisorType](#hypervisortype)
+  * [HypervisorConfig](#hypervisorconfig)
+  * [AgentType](#agenttype)
+  * [ProxyType](#proxytype)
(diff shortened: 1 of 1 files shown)


jodh-intel:
This looks really useful to me. My only question/reservation is how we can ensure that doc is updated as and when changes are made to the API. Was is entirely hand-written or is it atleast partly generated?

I'm also wondering if we could create (auto-generate) some state/sequence diagrams (maybe there is a tool like Doxygen for go that could do this for us)?

documentation/api/1.0/api.md
```diff
@@ -0,0 +1,864 @@
...
+```Go
+// CreatePod is the virtcontainers pod creation entry point.
+// CreatePod creates a pod and its containers. It does not start them.
+func CreatePod(podConfig PodConfig) (VCPod, error)
+```
+
+#### `DeletePod`
+```Go
+// DeletePod is the virtcontainers pod deletion entry point.
+// DeletePod will stop an already running container and then delete it.
```
grahamwhaley: Not directly related to this PR probably, but it would be nice if the func docs here noted when you should and should not call the funcs (like `not to be called on a running container` or similar, and what you can expect to get back in the cases of success and failure.

documentation/api/1.0/api.md
```diff
@@ -0,0 +1,864 @@
...
+```Go
+// ResumePod is the virtcontainers resuming entry point which resumes
+// (or unpauses) and already paused pod.
+func ResumePod(podID string) (VCPod, error)
+```
+
+## Container API
+
+The virtcontainers 1.0 container API manages specific pod containers
+[lifecycle](#container-functions).
```
grahamwhaley: I think this should be 'manages a specific', or 'pod container lifecycles' - feels slightly incomplete sentence as is ?

documentation/api/1.0/api.md
```diff
@@ -0,0 +1,864 @@
...
+
+A virtcontainers container is process running inside a containerized
+environment, as part of a hardware virtualized context. In other words,
+a virtcontainers container is just a regular container runing inside a
```
grahamwhaley: s/runing/running/

documentation/api/1.0/api.md
```diff
@@ -0,0 +1,864 @@
...
+```
+
+#### `ProcessListContainer`
+```Go
+// ProcessListContainer is the virtcontainers entry point to list
+// processes running inside a container
+func ProcessListContainer(podID, containerID string, options ProcessListOptions) (ProcessList, error)
+```
+
+## Examples
```
jodh-intel: Should there be some examples here?

documentation/api/1.0/api.md
```diff
@@ -0,0 +1,864 @@
...
+
+#### `PodConfig`
+```Go
+// PodConfig is a Pod configuration.
+type PodConfig struct {
+	ID string
+
+	Hostname string
+
+	// Field specific to OCI specs, needed to setup all the hooks
```
jodh-intel: This looks like it should start "Hooks is a field specific...".

documentation/api/1.0/api.md
```diff
@@ -0,0 +1,864 @@
...
+)
+```
+
+##### `ProxyType`
+```Go
+// ProxyType describes a proxy type.
+type ProxyType string
+
+const (
+	// NoopProxyType is the noopProxy.
```
jodh-intel: It might be worth explaining somewhere what all the "noop" bits are for.

documentation/api/1.0/api.md
```diff
@@ -0,0 +1,864 @@
...
+	NoopProxyType ProxyType = "noopProxy"
+
+	// NoProxyType is the noProxy.
```
jodh-intel: Also, explaining the difference between `NoopProxyType` and `NoProxyType` (which I think is documented at the top of each of those implementations at the moment (maybe we should move those comments to here).

documentation/api/1.0/api.md
```diff
@@ -0,0 +1,864 @@
...
+```
+
+##### `NetworkModel`
+```Go
+// NetworkModel describes the type of network specification.
+type NetworkModel string
+
+const (
+	// NoopNetworkModel is the No-Op network.
+	NoopNetworkModel NetworkModel = "noop"
```
jodh-intel: It would be good to clarify if this means "no network" or "default network" or "fake network for testing purposes".

jodh-intel:
A diagram showing how pods and containers differ would also be useful if that could be added somehow.

sameo:
@jodh-intel @grahamwhaley 
> My only question/reservation is how we can ensure that doc is updated as and when changes are made to the API. Was is entirely hand-written or is it atleast partly generated?

It was entirely hand written. The API that's described here is the 1.0 one, and at least we know we will not push backward incompatible changes without changing the version number. So those changes at least won't have to be propagated to the document.

The next step is to use godoc instead of manually generating this documentation and I think the proper way to do so would be to create a specific `api` package for virtcontainers and possibly split the code into more packages. Right now, our entire exported API is under the `virtcontainers` package and as a result the generated godoc is messy. I'll look into that next.

jodh-intel:
Hi @sameo - makes sense to me. Thanks for putting this together!

lgtm

PR #632, opened 2018-02-21
pod: Simplify and cleanup the pod

sboeuf:
This pull request globally cleans up a few things related to the pod and simplify a few functions in order to make the code easier and clearer.

### DIFF
--- a/pod.go
+++ b/pod.go
@@ -727,6 +727,28 @@ func (p *Pod) findContainer(containerID string) (*Container, error) {
 		containerID, p.id)
 }
 
+// removeContainer removes a container from the containers list held by the
+// pod structure, based on a container ID.
+func (p *Pod) removeContainer(containerID string) error {
+	if p == nil {
+		return errNeedPod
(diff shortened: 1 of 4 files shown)


pod.go
```diff
@@ -727,6 +727,28 @@ func (p *Pod) findContainer(containerID string) (*Container, error) {
 		containerID, p.id)
 }
 
+// removeContainer removes a container from the containers list held by the
+// pod structure, based on a container ID.
+func (p *Pod) removeContainer(containerID string) error {
+	if p == nil {
```
jodh-intel: Nit: Great to see lots of error checking, but I don't think this scenario is possible?

sboeuf: Yeah you're right, let me remove this from this function and from the other one `findContainer()`

sboeuf: Actually, this is possible... The tests are testing this possibility and there is no segfault. So yes basically it could happen. Let's keep the extra check then !

PR #633, opened 2018-02-22
block: Fix disabling block device use

amshinde:
This configuration was added to disable block devices from being
used and using 9p instead. This got dropped somewhere, add it back.


sboeuf:
LGTM

Issue #638, opened 2018-02-26
/var/lib/virtcontainers (and subdirs) as chmod 01000 (d---------)


grahamwhaley:
I'd removed my `/var/lib/virtcontainers/pods` dir (it was dirty), and tried to run the 'make functional' tests from the cc [tests repo](https://github.com/clearcontainers/tests), and got some errors like:

```
...
Stderr: open /var/lib/virtcontainers/pods: permission denied
...
```

Running up a single cc-runtime container and having a look down in `/var/lib/virtcontainers` shows us:

```bash
$ ls -laR virtcontainers
virtcontainers:
total 12
d---------  3 root root 4096 Feb 26 16:58 .
drwxr-xr-x 50 root root 4096 Feb 26 16:58 ..
d---------  3 root root 4096 Feb 26 16:58 pods

virtcontainers/pods:
total 12
d--------- 3 root root 4096 Feb 26 16:58 .
d--------- 3 root root 4096 Feb 26 16:58 ..
d--------- 3 root root 4096 Feb 26 16:58 b9634f3cb5b46b4dc8500176c9f8c27a132b0f1e9f3f0be7ccc3c116541394dd

virtcontainers/pods/b9634f3cb5b46b4dc8500176c9f8c27a132b0f1e9f3f0be7ccc3c116541394dd:
total 64
d--------- 3 root root  4096 Feb 26 16:58 .
d--------- 3 root root  4096 Feb 26 16:58 ..
drwxr-x--- 2 root root  4096 Feb 26 16:58 b9634f3cb5b46b4dc8500176c9f8c27a132b0f1e9f3f0be7ccc3c116541394dd
-rw-r--r-- 1 root root 52021 Feb 26 16:58 config.json

virtcontainers/pods/b9634f3cb5b46b4dc8500176c9f8c27a132b0f1e9f3f0be7ccc3c116541394dd/b9634f3cb5b46b4dc8500176c9f8c27a132b0f1e9f3f0be7ccc3c116541394dd:
total 36
drwxr-x--- 2 root root  4096 Feb 26 16:58 .
d--------- 3 root root  4096 Feb 26 16:58 ..
-rw-r--r-- 1 root root 26806 Feb 26 16:58 config.json
```

I think the root cause is coming from:
https://github.com/containers/virtcontainers/blob/master/filesystem.go#L157
where we have code like:

```
err = os.MkdirAll(path, os.ModeDir)
```

[...]

PR #647, opened 2018-02-28
filesystem: set correct access mode for pod dir tree

grahamwhaley:
Some of the filesystem pod data tree (/var/lib/virtcontainers/*)
is currently created with the permission 01000 (d---------). This
looks like we have forgotten to or in the dirMode bits into
the MkdirAll() calls.

Or in the appropriate bits.

Fixes: #638

Signed-off-by: Graham whaley <graham.whaley@intel.com>

### DIFF
--- a/filesystem.go
+++ b/filesystem.go
@@ -88,7 +88,7 @@ const mountsFile = "mounts.json"
 const devicesFile = "devices.json"
 
 // dirMode is the permission bits used for creating a directory
-const dirMode = os.FileMode(0750)
+const dirMode = os.FileMode(0750) | os.ModeDir
 
 // storagePathSuffix is the suffix used for all storage paths
 const storagePathSuffix = "/virtcontainers/pods"
@@ -154,7 +154,7 @@ func (fs *filesystem) Logger() *logrus.Entry {
 func (fs *filesystem) createAllResources(pod Pod) (err error) {
 	for _, resource := range []podResource{stateFileType, configFileType} {
 		_, path, _ := fs.podURI(pod.id, resource)
-		err = os.MkdirAll(path, os.ModeDir)
+		err = os.MkdirAll(path, dirMode)
 		if err != nil {
 			return err
 		}
@@ -163,7 +163,7 @@ func (fs *filesystem) createAllResources(pod Pod) (err error) {
 	for _, container := range pod.containers {
 		for _, resource := range []podResource{stateFileType, configFileType} {
 			_, path, _ := fs.containerURI(pod.id, container.id, resource)
-			err = os.MkdirAll(path, os.ModeDir)
+			err = os.MkdirAll(path, dirMode)
 			if err != nil {
 				fs.deletePodResources(pod.id, nil)
 				return err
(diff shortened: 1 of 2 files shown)


sboeuf:
LGTM

grahamwhaley:
repushed with gofmt fixes to appease gods of CI

jodh-intel:
Good catch! Could you add a test or two to `filesystem_test.go`?

grahamwhaley:
heh, OK @jodh-intel , I did consider post-push that maybe I can add a test for it - let me add it to my list...
I'll mark this as DNM in the mean time.

filesystem.go
```diff
@@ -154,7 +154,7 @@ func (fs *filesystem) Logger() *logrus.Entry {
 func (fs *filesystem) createAllResources(pod Pod) (err error) {
 	for _, resource := range []podResource{stateFileType, configFileType} {
 		_, path, _ := fs.podURI(pod.id, resource)
-		err = os.MkdirAll(path, os.ModeDir)
+		err = os.MkdirAll(path, os.ModeDir|dirMode)
```
amshinde: Do we even need os.ModeDir here with the mkdir system call?

grahamwhaley: Yep, as per the Issue and commit. If we don't do that then we get dirs created as `d---------` (01000). `ModeDir` **only** gives you the `d` bit - which is why the go docs say you must have it when creating dirs - but they omit to say you still need to supply the rest of the mode bits you want... afaict. Feel free to double check and verify. When I tested this it did what I expected (changed the dirs from 01000 to 01750)

grahamwhaley: Ah, or maybe I see what you mean now ... :-) We might only need `dirMode`, like we use for other calls to `os.Mkdir[All]`.

From the golang docs at:
https://golang.org/pkg/os/#FileMode
we have:

> The only required bit is ModeDir for directories.

But, at https://github.com/containers/virtcontainers/blob/master/filesystem.go#L91
we have
```
const dirMode = os.FileMode(0750)
```
Maybe the better solution then is to add `os.ModeDir` to our `dirMode`, and just use that everywhere (as we pretty much do already).

thoughts @amshinde @sboeuf ?

grahamwhaley:
sure, np. I was going to ask you if the tests were OK once the CI ran.
I'll wait to see if @sboeuf @amshinde have any more thoughts on the merging of the mode bits into the global var, and then rework as necessary...

filesystem.go
sboeuf: @grahamwhaley the point from @amshinde is that we might not need `os.ModeDir` since this should be added by default by `Mkdir()`.
I have checked this with a tiny Go program and I can confirm this is not needed at all since this bit is set by default.
So let's remove it completely, and keep only the `dirMode`

grahamwhaley: Hmm. I'll quote again, more, from the golang [docs](https://golang.org/pkg/os/#FileMode):

> A FileMode represents a file's mode and permission bits. The bits have the same definition on all systems, so that information about files can be moved from one system to another portably. Not all bits apply to all systems. The only required bit is ModeDir for directories.

which implies to me, for future system portability, we should have the bit set. It is doing no harm, so I vote to leave it to be 'correct' according to the golang docs - no ?

sboeuf: Oh yeah I've missed that sorry... I agree it looks safer to keep this in. Then please move this inside the var `dirMode` !

grahamwhaley: np. I'll rework and merge in the tests to the commit as well...

grahamwhaley:
updated, re-pushed

PR #659, opened 2018-03-02
vendor: Constraint containerd vendoring

### LINKED ISSUE #658: vendor: containerd package version should be constrained since it is imported
`pkg/oci/utils.go` does import `""github.com/containerd/cri-containerd/pkg/annotations""`, meaning this is a direct import, and that's why we should constraint this package in our vendoring.

sboeuf:
Since we are directly importing containerd package as a dependency,
this commit sets a constraint on the revision.

Fixes #658


jcvenegas:
lgtm

jodh-intel:
lgtm

PR #660, opened 2018-03-05
test: move required test support files locally

grahamwhaley:
Rather than have some of our required test support files live in `/tmp`,
let's install them in a local dir instead.

### DIFF
--- a/utils/virtcontainers-setup.sh
+++ b/utils/virtcontainers-setup.sh
@@ -17,10 +17,7 @@
 
 set -e
 
-if [[ $EUID -ne 0 ]]; then
-	echo "This script must be run as root"
-	exit 1
-fi
+SCRIPT_PATH=$(dirname $(readlink -f $0))
 
 if [[ -z "$GOPATH" ]]; then
 	echo "This script requires GOPATH to be set. You may need to invoke via 'sudo -E PATH=$PATH ./virtcontainers-setup.sh'"
@@ -42,7 +39,7 @@ virtcontainers_build_dir="virtcontainers/build"
 echo -e "Create temporary build directory ${tmpdir}/${virtcontainers_build_dir}"
 mkdir -p ${tmpdir}/${virtcontainers_build_dir}
 
-TMPDIR="/tmp"
+TMPDIR="${SCRIPT_PATH}/supportfiles"
 OPTDIR="/opt"
 ETCDIR="/etc"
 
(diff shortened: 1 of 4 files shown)


grahamwhaley:
Marked as DNM so we can discuss (so really an RFC).

- the `make check` works as before
- the `go test -bench=.` gives the same results as before (see below)

I didn't realise we had some benchmarks ;-) I managed to run them with:

```bash
$ sudo -E env PATH="$PATH" go test -bench=.
```

It looks like we may have bitrotted though (as I suspect nobody runs these). The output I have from both the original and the modified-in-this-patch I get is:

```
Some output on stderrERRO[0004] hook error                                    arch=amd64 error="exit status 1: stdout: , stderr: " hook-type=pre-start source=virtcontainers subsystem=hooks
Some output on stderrBenchmarkCreateStartStopDeletePodQemuHypervisorHyperstartAgentNetworkCNI-4              2         699418793 ns/op
BenchmarkCreateStartStopDeletePodQemuHypervisorNoopAgentNetworkCNI-4                   2      552637127 ns/op
BenchmarkCreateStartStopDeletePodQemuHypervisorHyperstartAgentNetworkNoop-4            3      532006579 ns/op
BenchmarkCreateStartStopDeletePodQemuHypervisorNoopAgentNetworkNoop-4                 10      251059396 ns/op
BenchmarkCreateStartStopDeletePodMockHypervisorNoopAgentNetworkNoop-4                500        4059463 ns/op
BenchmarkStartStop1ContainerQemuHypervisorHyperstartAgentNetworkNoop-4                 2     1117480785 ns/op
BenchmarkStartStop10ContainerQemuHypervisorHyperstartAgentNetworkNoop-4                1     6304859774 ns/op
PASS
ok      github.com/containers/virtcontainers    27.195s
```

@sboeuf - something you think we should fix? Any idea how?

utils/virtcontainers-setup.sh
```diff
@@ -42,16 +39,16 @@ virtcontainers_build_dir="virtcontainers/build"
 echo -e "Create temporary build directory ${tmpdir}/${virtcontainers_build_dir}"
 mkdir -p ${tmpdir}/${virtcontainers_build_dir}
 
-TMPDIR="/tmp"
+TMPDIR="${SCRIPT_PATH}/supportfiles"
```
sboeuf: Could you add a commit adding this whole directory to `.gitignore` ? This way, we won't have git tracking those files that we could add to a patch by mistake.

grahamwhaley: (oops, I meant to but forgot...) done and pushed

sboeuf:
LGTM

PR #666, opened 2018-03-08
kata_agent: Add virtio-scsi support

sboeuf:
This patch allows to plug block devices to Kata Containers agent,
using virtio-scsi driver. Those devices can be both rootfs or
simple block device passthrough.


sboeuf:
Related to https://github.com/kata-containers/agent/pull/163

